PR #233, opened 2026-07-26
feat(api): read the committed catalog.json, falling back to the GraphQL fan-out

gtanczyk:
## Status: rebased onto #234, now composes with it

This PR originally carried two changes. The first (resilient catalog cache — request coalescing + stale-on-error) is **already on master** as 6d019fdb and survived #234's rewrite intact. What remains here is the single-read fast path, rebased so it *builds on* #234 rather than competing with it.

## Why this is still worth landing after #234

#234 cut catalog rebuilds from ~167 Contents reads to `1 + ceil(N/30)` requests. That fixed the outage. Two things it structurally cannot do:

1. **`touch` never reaches visitors.** It is derived from each game's *code* by the games repo's `inferTouchSupport`. No amount of SPEC.md frontmatter reading can reproduce it. That field exists specifically so the site can tell a phone visitor what they can actually play — today it is computed, committed, and silently dropped at the API boundary.
2. **It grows with the catalog.** `1 + ceil(N/30)` is 4 requests at 83 games, 11 at 300. The games repo went 73 → 83 in a single day.

The games repo now commits a website-ready `catalog.json` (gamedevpl/www.gamedev.pl-games#56, **merged**), held fresh by validate Check 14. Reading it answers the whole catalog in **one request, constant in N**, and carries `touch`. Right now that artifact is maintained but unconsumed — this PR is what makes it pay for itself.

## How

Fast path reads `catalog.json`; **#234's chunked GraphQL fan-out stays as the fallback** for refs predating the artifact and for PR branches mid-build. Nothing is removed.

The artifact is re-validated on the way in — slug and media-filename regexes, player bounds, orientation and touch enumerations — because the filenames it vouches for become servable URLs, and a repo file earns no more trust than any other remote input. Malformed entries are skipped individually; a payload that is not a catalog at all falls through to the fan-out. Status coercion unchanged: only `archived`/`disabled` hide a merged game.

## Verified

- Against the **real 83-game artifact**: every entry parses, all carry `touch` (68 gamekit, 12 controllers, 3 native), in a single request.
- github-client 17/17 (4 new: fast path, touch preservation, hostile-artifact re-validation, malformed-JSON fallback). Existing GraphQL tests retained and still exercising the fallback.
- Full API suite 437/437; lint and type-check clean.

---

> [!NOTE]
> **Medium Risk**
> Catalog loading behavior changes at the API boundary (new primary data source) with security-sensitive media filename validation; fallback limits blast radius if the artifact is wrong or missing.
> 
> **Overview**
> **`getCatalog`** now tries a **one-request fast path** by reading root **`catalog.json`** from the games repo when present, instead of always listing `games/` and batching GraphQL reads.
> 
> The committed file is parsed with **`parseCommittedCatalog`** and strict re-validation (slugs, media paths, orientation/touch enums, multiplayer bounds) because those filenames become servable URLs. Malformed rows are skipped; invalid JSON, non-arrays, or arrays where every row fails validation trigger **fallback** to the existing SPEC/GraphQL path so a bad artifact does not empty the site. A valid **`[]`** still means an empty catalog.
> 
> **`CatalogGameEntry`** gains optional **`touch`** (`gamekit` | `native` | `controllers` | `none`), which only the committed artifact can supply—the fallback path leaves it unset.
> 
> Tests mock a **`catalog.json`** probe on the fallback path, bump expected fetch counts, and add coverage for the fast path, **`touch`** preservation, hostile payloads, schema-mismatch fallback, empty artifact, and non-JSON fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d21f615add8de5a30b51e67612c5f6e0c0c628bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/github-client.test.ts
+++ b/apps/api/src/github-client.test.ts
@@ -15,6 +15,10 @@ function specMd(frontmatter: Record<string, string>): string {
 function catalogFetchImpl(dirs: string[], files: Record<string, string | null>): typeof fetch {
   return vi.fn(async (input: RequestInfo | URL, init?: RequestInit) => {
     const url = String(input);
+    if (url.includes('/contents/catalog.json')) {
+      // No committed artifact on this ref — exercises the GraphQL fallback below.
+      return new Response('not found', { status: 404 });
+    }
     if (url.includes('/contents/games?')) {
       return new Response(
         JSON.stringify([
@@ -89,9 +93,10 @@ describe('getCatalog', () => {
         orientation: 'any',
       },
     ]);
-    // One directory listing + one GraphQL chunk — not one Contents read per file.
-    expect(fetchImpl).toHaveBeenCalledTimes(2);
-    expect(String(vi.mocked(fetchImpl).mock.calls[1]?.[0])).toContain('/graphql');
+    // Artifact probe + one directory listing + one GraphQL chunk — not one
+    // Contents read per file.
+    expect(fetchImpl).toHaveBeenCalledTimes(3);
+    expect(String(vi.mocked(fetchImpl).mock.calls[2]?.[0])).toContain('/graphql');
   });
 
   it('strips quotes from frontmatter values and skips games without a title', async () => {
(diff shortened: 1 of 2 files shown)


gtanczyk:
Good catch — fixed in d21f615a.

You're right that this was a real hole: `parseCommittedCatalog` returned `null` only when the payload failed to parse or wasn't an array. A **valid** JSON array whose every row failed validation — exactly what a field rename in the games repo would produce — returned `[]`, which the caller read as success. The fan-out never ran and every visitor would have seen a catalog with no games.

Rows present but none usable now returns `null` so the GraphQL fan-out takes over. An artifact that is genuinely `[]` still reads as an empty catalog, since that's the one case where empty is the honest answer.

Two tests added for the two sides of that line: an array of unrecognised rows falls through to the fan-out, and `[]` returns empty without the fan-out running.

PR #250, opened 2026-07-26
Add written player feedback capture (IL-1, signal source #1)

gtanczyk:
## Summary

- Adds `POST /api/games/:slug/feedback` — free-text player feedback for a **published** game, keyed by games-repo slug (`player-feedback.ts`). This is the sibling of the creator-feedback route in `submissions.ts`, with three deliberate differences documented in the code and in `docs/improvement-loop-plan.md`:
  1. Gated by `PublishedSlugGate` (catalog membership) and stored under `games/{slug}/playerFeedback/{id}` — like votes and telemetry, not the creator route's per-submission token. Most published games have no `submissions/{n}` document, so addressing by submission would silently drop most real feedback (the same bug telemetry and votes already hit and fixed).
  2. Requires a signed-in session, unlike votes' public-read/session-write split — free text is a materially bigger abuse surface than a thumb. Documented tradeoff: this excludes the anonymous players the funnel exists to measure.
  3. Does **not** post to GitHub, unlike creator feedback — most published games have no open PR/issue to comment on, and this is comparatively high-volume. It's stored only, for a future IL-2 scorecard/theme-extraction pass.
- Moderated via `contentChecker.checkFields`, sanitized via `sanitizeCreatorText`, same 422 contract as the creator route. Separate daily quota (`playerFeedback` `UsageCounters` kind) from the creator `feedback` counter.
- Ships the minimal post-play prompt: `PlayerFeedbackWidget` in the theater bar next to the vote widget (i18n in `en.json`/`pl.json`).
- Corrects `docs/improvement-loop-plan.md`'s Data model section, which originally specified `submissions/{issueNumber}/playerFeedback/{id}` — the same mistake votes made and fixed, now fixed here too, in the same commit as the code.
- Marks IL-1 (Capture) complete in the plan doc, and corrects a stale claim in `product-instrumentation`'s Known Gaps ("no game calls `GameKit.progress`" — already false before this PR; see companion PR gamedevpl/www.gamedev.pl-games#70).

## Review fixes

Three findings, all real and all addressed:

1. **Privacy notice did not disclose this collection.** The notice's data table states "*The table below is the complete list. If a category is not here, we do not collect it*", and UID-attributed player feedback was not in it. Added a row in **both** languages. It also covers **votes** — same uid-keyed shape, never disclosed when they shipped in #244 either, and collecting undisclosed today — so this closes a pre-existing gap in the same change. Section 8 (account deletion) now promises to remove both, making the stated "until you delete your account" retention a promise that is actually kept.
2. **Success copy over-promised.** It said "the creator will see this", but nothing reads this collection yet (the creator scorecard is IL-2). Now: "your feedback was received."
3. **Minimum length was checked before destructive sanitization.** Ten asterisks cleared the schema's raw-text minimum, survived moderation, and sanitized to the empty string — spending a day's quota to store a blank row. Now sanitizes first and validates the text that would actually be stored, before any quota is spent. Regression test added asserting both the 400 and that quota is untouched.

Investigating (3) surfaced a more serious variant of the same root cause, also fixed: **moderation ran only on the raw text**, and `sanitizeCreatorText` strips markdown emphasis characters to nothing — so `sh*it` cleared the term check and was then stored in its matching form. Both forms are now moderated when they differ; identical strings are checked once, so plain prose costs exactly one call and only markup-bearing input pays for two. Raw is still checked because it is the only form retaining markdown link targets for the URL-count rule. HTML tags are *not* a vector (they sanitize to a space, which does not rejoin a word) — verified empirically; the first draft of that test used `<b>` and correctly failed.

## Rebased onto current master

Merged `origin/master` (through `71f8d5c`) to clear a conflict introduced by the personal-access-token work. Resolutions:

- **`store.ts`** — purely additive on both sides; the `playerFeedback` and `accessTokens` methods coexist in the interface, `InMemoryStore`, and `FirestoreStore`.
- **`telemetry.test.ts`** — master fixed the *same* midnight-UTC backdating flake I had, and did it better: a shared `listBackdated` helper covering both backdating tests, plus an explicit `toBeDefined()` with a diagnostic message, where mine was inline in one test. **Took master's version and dropped mine.**

Post-merge the suite is 623 API tests (mine plus the new PAT suites), all passing.

## Instrumentation definition-of-done

Per the `product-instrumentation` skill: this affects **question 6 (per-game health)** — "does the game load, run, hold attention, and get finished?" Free-text feedback is signal source #1 in the improvement-loop plan, the last one untouched. It does not yet have a read/aggregate path (that's IL-2's scorecard); today the only query answering anything is a raw Firestore read of `games/{slug}/playerFeedback`, enough to confirm capture works but not yet a creator-facing answer. Named explicitly rather than left silent.

## Test plan

- [x] `npm run lint`
- [x] `npm run type-check`
- [x] `npm test` — `apps/api` 623 tests, `apps/web` 254 tests
- [x] `npm run build`
- [ ] Manual browser check of the post-play prompt — not done in this sandbox: with no GitHub token the published-slug gate answers false for everything, so the route 404s locally. Flagging for reviewer verification.


gtanczyk:
All three findings were real and are fixed. Two of them turned out to be bigger than reported, so noting the deltas:

**1. Privacy notice** — added a disclosure row in both language versions covering the account identifier, the improvement/moderation purpose, the Art. 6(1)(f) basis, and retention. Two things worth flagging beyond the report:

- The gap was wider than player feedback. **Votes** (`games/{slug}/votes/{uid}`) have the identical uid-attributed shape and were never disclosed when they shipped either. Since the notice states "*the table below is the complete list*", leaving votes out would have kept it inaccurate. One row now covers both reactions.
- Retention is only a real promise if deletion honours it, so §8 (account deletion) now commits to removing votes and feedback too.
- ⚠️ **Left for a human:** the notice's own change clause promises 14 days' notice before changes take effect. `LEGAL_EFFECTIVE_DATE` was already today so nothing was bumped, but whether this new collection needs that notice period before the endpoint goes live is your call, not mine.

**2. Success copy** — now "Thanks — your feedback was received." (and the Polish equivalent). Agreed it was over-promising; I chose the honest capture-only wording rather than shipping a creator-facing reader, which belongs with IL-2's scorecard.

**3. Sanitize-then-validate** — fixed, with a regression test asserting both the 400 and that quota is *not* spent. Investigating it surfaced a second, more serious variant of the same root cause, now also fixed:

> Moderation also ran on the raw text only. `sanitizeCreatorText` strips markdown emphasis characters to nothing, so **`sh*it` cleared the term check and was then stored in its matching form** — a moderation bypass, not just a quality-gate miss. Both forms are now moderated when they differ; identical strings are checked once, so plain prose still costs exactly one call and only markup-bearing input pays for two. Raw is still checked because it is the only form that retains markdown link targets for the URL-count rule.
>
> HTML tags are *not* a vector here — those sanitize to a space, which does not rejoin a word. I verified that empirically rather than assuming it; my first draft of the test used `<b>` and correctly failed.

**Also fixed, unrelated to this branch:** `telemetry.test.ts` asserted that a six-hour-backdated event lands in *today's* partition. Events are filed under the date their own timestamp names, so between 00:00 and 06:00 UTC it lands in yesterday's and the test failed nightly with an opaque `cannot read properties of undefined`. It reproduces with this branch's changes stashed, so it is pre-existing — but it would have turned this PR red on any re-run in that window, so it is fixed here.

On the CI question in your review: the hosted checks were green on the previous commit (`Lint, Test & Build`, CodeQL ×2, Secret Scanning, Games-repo contract) and are re-running now on `3c58c44`. Locally: lint, type-check, build, and 543 API + 254 web tests all pass.

---

gtanczyk:
Withdrawing the 14-day-notice flag from my previous comment — I over-flagged it.

The clause is scoped to **material** changes: *"We will announce material changes in the Service at least 14 days in advance."* This isn't one. No existing data changes hands, purpose, or retention; the feedback row describes collection that only occurs if someone opts to use a feature that does not exist yet, and the notice ships in the same commit as the endpoint — so it is accurate before the first record is written. That is what Art. 13 asks for (inform at the time data is obtained); a waiting period adds nothing to it.

On the **votes** half of that row the delay would be actively counterproductive. Votes shipped in #244 and are collecting uid-attributed data today with no disclosure. Holding the corrected notice for 14 days would extend the period during which the notice is inaccurate — that half is an argument for merging sooner, not later.

No code change; this comment supersedes the ⚠️ item in the PR description, which I have removed.

---

PR #255, opened 2026-07-27
Block deploys that break playable games

gtanczyk:
Playable games are the product. Every check we have stops at HTTP, so today a deploy can break play completely and still go green — proving `/api/games/<slug>` returns HTML says nothing about whether that HTML **runs**. A CSP header, a sandbox change, a bundle regression or a broken assembler each return a healthy 200 and a black screen.

This drives a real browser against the candidate revision before traffic is promoted.

## Asserted on the deployed, rendered page

- **The sandbox invariant** — `allow-scripts`, never `allow-same-origin`. Previously only checked against the component in jsdom; this is the first time it's asserted on what a visitor actually receives.
- **Games actually run** — the game document executes, its canvas draws, and the drawing *changes between two samples*. A game that painted one frame and froze is not "working", and a single sample can't tell the difference.
- **A `bot:ci` token exchanges for a session cookie the SPA honours** — proving auth works in the rendered app, not just via curl.

## One broken game must not stop the website shipping

Games live in a separate, agent-maintained repo and change independently. So the suite samples several published games and decides by **breadth**, not by assertion type:

| Sample result | Meaning | Outcome |
| --- | --- | --- |
| most games fail | the deploy broke play | **blocks promotion** |
| exactly one fails | that game is broken | `::warning::`, ships |
| only one game exists | nothing to compare against | **blocks** — it *is* the site |

## Verification

Ran end-to-end against a local instance serving the three fixture games (`odd-one-out`, `pixel-dodge`, `range-squad`) — the play-route/sandbox test and the games-actually-run test both pass against a real browser and real game documents.

Before writing any of it, I probed the core assumption against a real sandboxed `srcdoc` frame: Playwright **can** read inside it (`origin: "null"`, canvas pixels readable) because it works through CDP, beneath the same-origin policy. The rAF-throttling trap the `verify-agent-work` playbook documents doesn't apply — a headless Playwright page counts as visible.

`npm run type-check && npm run lint && npm run test && npm run build` all exit 0, each run unpiped with its status checked. Lockfile verified stable under regeneration, so `npm ci` won't break.

**Not verified:** the suite has not run against production from here — this container's TLS-intercepting proxy resets Chromium's connections (curl works, the browser doesn't). CI has no such proxy. The `HTTPS_PROXY` and `SMOKE_CHROMIUM_PATH` support exists so the same suite runs from an agent VM.

## Measurement hygiene

Play telemetry is aborted client-side, so CI plays never enter the product numbers. Those events carry no uid by design, so the `bot:` exclusion that keeps agents out of creator metrics cannot reach them — without this, several fake plays per deploy would corrupt exactly the per-game health signal the data exists to provide.

## Costs and one correction

Adds a pinned `@playwright/test` devDependency, `e2e/` specs, and ~60–90s per deploy (browser install). It runs only in the deploy workflow, never in `ci-gate`. Traces and screenshots upload as an artifact on failure, so a red gate is diagnosable without reproducing the deploy locally.

This **contradicts guidance I added earlier** in `docs/agent-access-tokens.md` — "do not add Playwright to the repo's `package.json`; this app has no e2e harness by design." That was aimed at agents making unversioned dependency edits for ad-hoc checks. Now that a versioned harness exists, the doc is corrected to say: reuse `npm run e2e:deployed`, and still never add or bump the dependency for a one-off.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HT8ueQTiejuiksCUd2oxQx)_

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -306,6 +306,38 @@ jobs:
 
           echo "Authenticated smoke passed as ${UID_VALUE}: bearer auth, cookie exchange, and the session wall all live."
 
+      # Playable games are the product, and every check above this point can pass while
+      # play is completely broken: `/api/games/<slug>` returning HTML says nothing about
+      # whether that HTML *runs*. A CSP header, a sandbox change, a bundle regression or a
+      # broken assembler all return a healthy 200 and a black screen. This drives a real
+      # browser against the candidate and asserts games actually draw
+      # (e2e/deploy-smoke.spec.ts, docs/deployment.md).
+      #
+      # Blocking, deliberately. The suite itself distinguishes "one game is broken" (warns,
+      # since games live in a separate agent-maintained repo) from "most games are broken"
+      # (fails, because that is the deploy having broken play).
+      - name: Install Chromium for the e2e smoke
+        run: npx playwright install --with-deps chromium
+
+      - name: End-to-end smoke (real browser, real games)
+        env:
+          SMOKE_BASE_URL: ${{ env.CANDIDATE_URL }}
+          GAMEDEV_ACCESS_TOKEN: ${{ secrets.GAMEDEV_ACCESS_TOKEN }}
+        run: npm run e2e:deployed
+
+      # Traces and screenshots for the failing run only — a red deploy gate is useless if
+      # diagnosing it means reproducing the whole deploy locally.
+      - name: Upload e2e artifacts on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-failure-${{ github.run_id }}
+          path: |
+            test-results/
+            playwright-report/
+          retention-days: 7
+          if-no-files-found: ignore
+
       - name: Promote candidate revision to 100% traffic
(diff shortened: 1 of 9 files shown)


gtanczyk:
Superseded by #257, which landed a browser e2e suite with a better foundation than this branch — a proper `apps/e2e` workspace, `findChromium()` that globs by build number, one token exchange in `globalSetup`, and the TLS-through-proxy fix (`--ssl-version-max=tls1.2`) that explains the `ERR_CONNECTION_RESET` I couldn't get past here.

This branch also now conflicts with master on `package.json` / `package-lock.json`, since both add a Playwright dependency and an e2e script.

The two things worth keeping are carried forward in #261, rebuilt on `apps/e2e` instead of alongside it:

- **the deploy gate** — `npm run e2e` in `deploy.yml` against `CANDIDATE_URL` before promotion, which #257 deliberately left open
- **the breadth rule** — sample several games, block when most fail, warn when exactly one does, so a single broken game in the games repo can't hold the website's deploys hostage

Closing in favour of #261.

---

PR #260, opened 2026-07-27
Serve published games from a Cloud Storage snapshot instead of rebuilding per request

gtanczyk:
Paired with gamedevpl/www.gamedev.pl-games#74 — that PR adds the dispatch that triggers the bake. Neither changes behaviour without the other plus the setup steps below.

## The problem

Every play of a published game rebuilt it on the request path:

1. `getGameSources` fans out into authenticated GitHub reads — the game's own files, plus each GameKit module its `GAME.json` selects, plus the shared shell and audio.
2. `bundleGameTypeScript` runs esbuild over the result.
3. `assembleGameHtml` produces the self-contained document.

Cached five minutes per slug, per instance, with `--min-instances 0`.

Three consequences, all avoidable:

- **GitHub API availability and quota became a dependency of the play route** — the one route that must never be down. The per-file Contents fan-out for the *catalog* had already caused a rate-limit outage once; that is why `getCatalog` batches through GraphQL (`github-client.ts:919`). Game assembly kept the same shape.
- **The work is identical every time.** A published game changes only on a merge to the games repo's `main`. Rebuilding per cache miss recomputes a constant.
- **The cost lands on the visitor**, in a service pinned to `--max-instances 1`.

Baking moves the same work to merge time. The work itself doesn't change — only how often it happens.

## What changed

- **`apps/api/src/game-snapshot.ts`** — the store. Immutable `snapshots/<id>/` prefixes plus a small mutable `current.json` pointer. Objects are written first and the pointer last, so publishing is atomic, rollback is a pointer rewrite, and every snapshot object is cacheable forever. Talks to the Storage JSON API over `fetch` with an ADC bearer token rather than adding `@google-cloud/storage`: the surface needed is three verbs, `google-auth-library` was already a dependency, and keeping transport at the `fetch` boundary lets tests fake it the way `local-games-repo.ts` fakes GitHub.
- **`apps/api/src/game-snapshot-publish.ts`** — the bake.
- **`apps/api/scripts/publish-snapshot.ts`** — CLI with `--dry-run`, `--ref`, `--commit-sha`.
- **`apps/api/src/submissions.ts`** — catalog, play and media routes prefer the snapshot.
- **`.github/workflows/publish-games.yml`** — runs on `repository_dispatch` from the games repo, plus `workflow_dispatch` for manual re-bakes.
- **Infra** — bucket, split IAM and lifecycle rule in `setup-gcp.sh` (now step 7/7); `GAMES_SNAPSHOT_BUCKET` wired through `deploy.yml` and `deploy-api.sh`.
- **`docs/games-snapshot.md`** — design, publish path, configuration, and what this deliberately does not solve.

## Two decisions worth reviewing

**Assembly stayed on this side.** The games repo has its own `tools/build.ts` that emits standalone HTML, and uploading that instead would have removed GitHub from the loop entirely. It was the wrong trade: the restrictive CSP, the AI Act art. 50(2) provenance meta and the credential scan are serve-time policy owned by *this* repo. Baking on the games-repo side would move that policy into a repo that does not own it, invert the drift problem `games-repo-contract.ts` exists to prevent, and require giving the games repo GCP credentials.

**The snapshot is a fast path, never a requirement.** An unset `GAMES_SNAPSHOT_BUCKET`, a game that failed to bake, an unbaked media file, and a Cloud Storage outage all degrade to exactly the behaviour the site has today. That fallback is why this is safe to ship in one step rather than behind a flag.

A game that fails to bake stays in the snapshot catalog and gets no game object, so its play route falls back to GitHub. Dropping it would silently unpublish a game over a transient build failure, and catalog membership is the games repo's decision. The job still exits non-zero, since the games repo's own gate should have caught it.

## Invariants that survive, and are tested

- **Publication is still decided by the catalog**, not by what sits in the bucket — a stale object cannot resurrect an unpublished game.
- **Media is still gated by the catalog allowlist** *before* the snapshot is consulted, so a stray object cannot widen what the API serves.

One deliberate exception: `isSlugPublished` skips the snapshot when it forces a refresh. That only happens during the publishing→published transition, which is precisely the window where the snapshot is stale and GitHub is fresh.

## Verification

Lint, type-check and build clean. **677 API tests pass, 43 of them new** — 18 covering the store (object layout, pointer caching/coalescing/stale-on-error, path safety, cache headers), 12 the bake (policy reapplied, pointer written last, failure handling), 13 the serve path (snapshot hit avoids GitHub; miss, error and unset bucket all fall back; both allowlist invariants). Web (254) and generator (12) suites unaffected. Publish-script guard rails exercised directly; shell and YAML syntax checked.

## Note on the diff

The pre-commit prettier hook reformats `submissions.ts` wholesale on any commit that touches it, because master's copy was never prettier-clean. That buried the ~90-line change in an 806-line diff, so the untouched lines were restored to master's formatting and only the real edits kept — `submissions.ts` is now 88 insertions / 6 deletions. The file is left exactly as formatted as it is on master, and CI runs eslint rather than `prettier --check`, so nothing regresses. Worth fixing repo-wide separately, not here.

Both findings from the Copilot review are addressed in the second commit: `deploy-api.sh` now honours an explicitly empty `GAMES_SNAPSHOT_BUCKET` as the documented opt-out (`-` rather than `:-`, which also makes the `-n` guard below it meaningful), and `setup-gcp.sh` enables the Cloud Storage API that step 7 and the Cloud Run runtime both depend on.

## Before this does anything in production

Owner-run, none of which I could do from here:

1. `./infra/setup-gcp.sh` — provisions the bucket, split IAM (deploy SA writes, Cloud Run SA reads), and the 90-day lifecycle rule.
2. `SITE_DISPATCH_TOKEN` on the **games** repo — fine-grained PAT that may dispatch into this repo. Until it exists that step warns and skips.
3. Confirm the bucket name/region — I defaulted to `gamedevpl-games-snapshots` in `europe-west1` (co-located with Cloud Run); change in `deploy.yml`, `publish-games.yml` and `setup-gcp.sh` if you'd rather.

Worth a `--dry-run` via `workflow_dispatch` first to see the bake report against the real 83-game catalog before anything is written.

## Not addressed

Bytes still flow through Cloud Run — the API stays the games origin, which is what keeps `PRIVATE_BETA` and the rate limits meaningful. Serving the bucket to browsers directly would be faster but makes published games publicly reachable URLs. The object layout is already CDN-shaped for when that trade is worth making. Preview and draft routes still read GitHub live, and should. `--max-instances 1` remains the scaling ceiling.

https://claude.ai/code/session_01858DHfkaq7psPoJbveLkAk

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,6 +39,10 @@ jobs:
       SERVICE: 'gamedev-app'
       REPO: 'gamedev'
       GAMES_REPO: 'gamedevpl/www.gamedev.pl-games'
+      # Pre-assembled published games (apps/api/src/game-snapshot.ts), written by
+      # publish-games.yml. Unset would simply mean every play rebuilds from GitHub
+      # again — serving treats the snapshot as a fast path, never a requirement.
+      GAMES_SNAPSHOT_BUCKET: 'gamedevpl-games-snapshots'
       WIF_PROVIDER: 'projects/334141807880/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
       WIF_SERVICE_ACCOUNT: 'github-actions-deployer@gamedevpl.iam.gserviceaccount.com'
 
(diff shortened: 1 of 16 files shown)


gtanczyk:
Dry-ran the bake against the full live catalog before merge (isolated worktree, nothing written): **84 games, 356 media files, 0 failures**. Assembly holds across every published game, not just the fixtures.

That run surfaced one bug, fixed in 2b61075a: the root `snapshot:publish` wrapper lacked a trailing `--`, so npm ate the flags and forwarded only their values as positionals — `--dry-run` never reached the script. The workflow's `dry_run` input would have attempted a real publish (fails closed on the runner, since that path skips GCP auth; not so locally against ADC), and `--commit-sha` was dropped, leaving every pointer's `commitSha` null.

Bucket is provisioned: `gs://gamedevpl-games-snapshots`, europe-west1, UBLA, 90-day lifecycle on `snapshots/`, deployer `objectAdmin` / Cloud Run SA `objectViewer`, no public binding.

PR #323, opened 2026-07-28
P3: authoritative real-time zones — infra, protocol, host, and a game on it

gtanczyk:
All four P3 phases, one commit per phase. **Paired with gamedevpl/www.gamedev.pl-games#148** (`GAME_KIT_MODULES` moves in both; `contract:games-repo` enforces it).

Phase 1 was the blocking gate and is approved. Phases 3–4 landed against it.

---

## 1. Where the zone host runs — `docs/p3-zone-host-infra.md`

A separate **`gamedev-world` Cloud Run service**: own image (the `isolated-vm` native build lives there, contained), `--min-instances 0 --max-instances 1`, request-based billing, 60-minute timeout. The main API stays the front door and mints short-lived HMAC tickets; the world service verifies them and never sees a cookie.

It works because **a zone only needs CPU while somebody is connected, which is exactly what request-based billing sells**. Hibernation isn't an optimization — it's what makes scale-to-zero honest.

| | zero traffic | low traffic (beta) |
| --- | --- | --- |
| Compute | $0 — no instances | ~$0.09/active hour; first ~50 h/month free |
| Firestore | storage only | one snapshot per zone per 30 s while active |

Rejected with named triggers for revisiting: zones inside `gamedev-app` (every website deploy would mass-hibernate every live world), an always-on VM (right past ~150 active hours/month), Durable Objects (posture), session affinity (pins a client, not a room — the zone directory replaces it, designed and deferred).

## 2. Protocol + shell transport — `docs/p3-zone-protocol.md`

**A tick delta is the events the host applied, not a state diff.** That only works because every client runs the same deterministic `sim.ts` the host arbitrates with, so the event list reproduces the tick exactly. Smaller, needs no diff format the contract doesn't define, and keeps the shell from having opinions about game state — snapshots pass through as opaque strings, like saves. A periodic checksum on a delta is how a drifted client finds out within a second.

**The ticket carries a per-zone pseudonym, never a uid.** The host runs untrusted game code; handing it durable identity would make the sim isolate a privacy boundary as well as a safety one. Guests are admitted with a per-ticket tag — a sign-in wall in front of first play is the one thing this platform won't build.

A bug the tests found before the host existed: the input value field had to move from `v` to `d`, because every frame carries `v` as the protocol version and an enum input would have overwritten it. Party mode learned this on a key-release of `0`; a zone would have learned it on the first step north.

## 3. The host — `packages/zone-core` + `apps/world`

**The cage is `isolated-vm` in production and `node:vm` nowhere near it.** Both implement one interface and pass one suite, which is the claim the spike actually made: the scrubbing is portable, so swapping cages is a safety and cost decision rather than a behaviour one. `assertProductionCage` refuses to start with `node:vm` in production — it would satisfy every test here and none of the promise. The isolate test pins the line `node:vm` can't hold: a runaway tick is interrupted rather than taking the event loop with it.

**Waking is a rewind.** A snapshot records how many numbers the generator has drawn — that position isn't in the state, since the state is the sim's and the stream is the platform's — and a restore re-seeds and replays exactly that many. One test runs a world straight through, another sleeps in the middle, and they land on the same world.

**Budgets are enforced at runtime.** Check 23 proves a sim *can* tick in 8 ms on a build machine; this is what happens when it doesn't on a real one with fifteen other zones on the core. Twenty overruns in a hundred ticks and the zone is hibernated with prejudice — players told, last good snapshot standing, core returned. One slow tick is forgiven.

Two more bugs the tests found:

- **The tick budget was measuring the wrong clock.** The zone's `now` schedules ticks and stamps snapshots, and a test drives it in fixed jumps — so metering against it measured the *schedule*, and passed for a sim spending 40 ms a tick.
- **Storability has to be decided before `JSON.stringify`.** Stringify doesn't refuse the values it can't carry, it rewrites them: `Infinity`/`NaN` → `null`, `Map` → `{}`, `-0` → `0`. By the time the host had a string to inspect the evidence was gone, and the state would have kept looking storable until the zone slept and woke as a different world.

Plus an ordering bug: the opening snapshot was sent from inside `join()` before the host had registered the seat, so every arrival got a welcome and then silence.

## 4. A game on it — [games repo #148](https://github.com/gamedevpl/www.gamedev.pl-games/pull/148)

Ember Watch's stand-in is deleted and it runs against the real host, staying **identical offline** — same `sim.ts`, alone or with strangers.

---

## Verification

1,627 tests green (98 new here), lint and type-check clean, and in the games repo 91/91 games pass `validate.ts` with the **determinism gate unchanged**.

**Still inert.** `ZONE_HOST_URL` is unset, so admission 404s and all 88 published games behave exactly as before. Turning it on is `infra/deploy-world.sh` followed by one `--update-env-vars` on the app service.

Note on branching: the session is pinned to one designated branch, so phases land as separate commits rather than separate PRs. Happy to split.


gtanczyk:
CI is green here except `Games-repo contract`, and that one needs a merge, not a fix.

**Now green:** Lint/Test & Build, Secret Scanning, CodeQL, both Analyze jobs. Three real failures were fixed since the last review:

- `isolated-vm` is an optional native addon that a plain `npm ci` skips on a runner with no C++ toolchain, so `tsc` couldn't resolve it. The dynamic import now takes its specifier from a variable, and the isolate test suites skip — loudly — when the addon is absent. The test that a missing isolate is a *refusal to start* rather than a silent downgrade to `node:vm` runs either way.
- Master had moved 28 commits ahead and added `apps/api/fixtures/games-repo/shared/assemble-contract.json`, a third place the GameKit module list lives. Merged; `zone` added there too.
- The Ember Watch capture conflicted in the games PR because both sides had regenerated it. Re-ran it against the merged tree rather than picking a side, since neither recorded the game as it now is.

**Worth flagging:** while merging I found the serve cap was 24,000 bytes behind games-repo #123 (`gfx3d` 56,000 → 80,000) and fixed it — then master merged #308 doing exactly the same thing, so the two collided. I resolved in favour of master's version verbatim, since that's where the change belongs. Nothing of mine remains in that constant.

**The one red check is the lockstep working as intended.** It compares this repo's `GAME_KIT_MODULES` against games-repo `main`, and the only remaining difference is `zone`:

```
games-repo: … save, commons, mascot
website:    … save, commons, mascot, zone
```

The byte totals now agree, so this is the sole mismatch. It cannot go green until gamedevpl/www.gamedev.pl-games#148 merges — that PR is `clean` and fully green (91/91 games pass `validate.ts`, determinism gate unchanged).

**Merge order: #148 first, then re-run this check here.** Reverting `zone` from the list to make the check pass would break the pairing it exists to enforce, so I've left it.

---

PR #356, opened 2026-07-30
Add a how-to-play panel to the game player

jactan77:
Players arrive at a game with no idea which keys it takes. The catalog has carried a per-game `controls` string all along — required in every SPEC, non-empty for all 92 published games, already typed as `CatalogEntry.controls` in the browser — but the player never showed it; only two marketing surfaces did.

This adds a **How to play** control to the theater bar (and to the More menu on a phone) that opens a card of key → action rows, built from that string.

## What it looks like

Apex Sprint, entered from the arcade. Every row is that game's own `controls` string from the
catalog, split into keys and action; the Touch row is stated from `entry.touch`, which the
games-repo build derives from the game's source rather than from a spec claim.

How to play, desktop

The control sits on the theater bar next to Sound and Fullscreen:

Theater bar

On a phone the bar has no room for it, so it moves into the More menu, and the card fills the
width above the game's on-screen pad:

| More menu | Card |
| --- | --- |
| More menu on a phone | How to play on a phone |

## Why this is host chrome and not something inside the game

The player injects `#game-title,#game-desc,.game-controls,.hint{display:none!important}` into every game document (`gamePlayer.ts`), because the theater shows the title, description and sound instead. A controls affordance built into a game's own shell is therefore invisible here by construction. `HIDE_CHROME` is unchanged and `gamePlayer.test.ts` is untouched — that is deliberate, not an oversight.

Nothing is read across the iframe boundary: the card is built from the catalog entry the app already holds, so the sandbox invariant (`allow-scripts allow-pointer-lock`, no `allow-same-origin`) is untouched. No games-repo change and no snapshot re-bake is needed; this ships with a normal web deploy.

## Parsing `controls`

The field is free text an agent authored in a SPEC, and two delimiters are live in the catalog: semicolon-separated clauses that contain their own commas, and plain comma-separated lists. Each clause is then split once, at its first `" to "` / `" for "`, into a key column and an action column.

The split is deliberately conservative, and the threshold was measured rather than guessed: across all 92 entries (326 clauses) 293 split cleanly, and every one of the 33 left whole genuinely has no key/action shape — either prose (`Press A, S, D, F, or Space in sync with...`) or the key-space-action form (`W/S pitch`), whose first space is not a reliable boundary because `Shift boost / Ctrl cut` would split wrong. A clause that cannot be split keeps its own full-width row.

`controls` is prompt-influenced text, so it renders as JSX text nodes, never through an HTML sink, and a test pins that intent.

## Accessibility and player behaviour

- Escape closes the card before it closes the game, in **both** paths — the window listener and the Escape the sandboxed frame relays over the player bridge. Without that, reaching for the controls and pressing Escape throws you out of the game.
- Closing hands focus back to the game frame, not to the trigger: in a player the next key press is meant for the game, and focus left on a button turns the next Space into "reopen the card" instead of "fire".
- The card traps Tab. `aria-modal` is an assistive-tech hint and does not constrain focus on its own, so without this, tabbing walks into the chrome behind a running game.
- Entering fullscreen closes the card, because the bar that holds both triggers unmounts there.
- The card is portalled to the body: the theater bar carries a `backdrop-filter`, which would otherwise become the containing block for its fixed position — the bug `AuthModal` already documents.
- A long press does not raise the iOS selection loupe over a running game, matching the player everywhere else.

## Telemetry

`how_to_play_opened` carries no slug, keeping the visit stream unjoinable with the play stream exactly as `play_started` does. It has a route-level test: the mapper ends in a `default` branch that relabels anything unmatched as `play_started`, so a missing case is mislabelled data rather than an error — which no schema failure would ever reveal.

## Known gap

On a direct `/play/<slug>` link the control does not appear. The catalog is fetched on the home route only and `App.catalog.test.ts` pins that ("a direct game link costs the game and nothing else"), so the player holds a placeholder entry whose `controls` is empty — the same reason the byline and orientation are placeholders there today. I tried closing it and reverted: the fix broke that test, and honouring the repo's decision seemed right. Closing it properly needs a per-game catalog read the API does not currently offer.

Generated and draft games have no catalog entry and so no controls; the control does not render for them.

## Verification

`npm run type-check && npm run lint && npm run test && npm run build` — all green. New tests: `howToPlay.test.ts` (parser, against verbatim catalog strings), `HowToPlayPanel.test.tsx` (rows, dialog semantics, focus, Tab trap, duplicate clauses, escaping), additions to `GameTheater.test.tsx` (control visibility, both Escape paths, focus on open from the More menu, fullscreen) and to `visit-telemetry.test.ts`.

Also driven in a real browser against a local build serving the live catalog of 92 games — which is how the pointer-lock/deep-link behaviour above was found rather than assumed.

### DIFF
--- a/apps/api/src/store.ts
+++ b/apps/api/src/store.ts
@@ -477,7 +477,7 @@ export interface TelemetryEvent {
 export interface VisitEvent {
   /** Per-tab uuid from `sessionStorage`. Dies with the tab; never a uid. */
   visitId: string;
-  type: 'visit_started' | 'route_viewed' | 'play_started' | 'create_step' | 'waitlist_step';
+  type: 'visit_started' | 'route_viewed' | 'play_started' | 'how_to_play_opened' | 'create_step' | 'waitlist_step';
   /** Server-anchored instant, derived like `TelemetryEvent.at`. */
   at: string;
   /** Milliseconds from visit start — the trustworthy measure of within-visit timing. */
--- a/apps/api/src/visit-telemetry.test.ts
+++ b/apps/api/src/visit-telemetry.test.ts
@@ -71,6 +71,25 @@ describe('POST /api/telemetry/visit', () => {
     expect(events.filter((event) => event.type === 'play_started')).toHaveLength(2);
   });
 
+  it('records how_to_play_opened as itself, and without a game identity', async () => {
+    // The mapper ends in a `default` branch that relabels anything unmatched as
+    // `play_started`, so a member with no case of its own is silently mislabelled
+    // rather than rejected — which no schema error would ever reveal.
+    await post(app, {
+      visitId,
+      flushMsSinceStart: 400,
+      events: [
+        { type: 'play_started', msSinceStart: 100 },
+        { type: 'how_to_play_opened', slug: 'space-hop', msSinceStart: 400 },
+      ],
+    });
+
+    const events = await store.listVisitEvents(today(), { visitId });
+    expect(events.filter((event) => event.type === 'how_to_play_opened')).toHaveLength(1);
+    expect(events.filter((event) => event.type === 'play_started')).toHaveLength(1);
+    expect(JSON.stringify(events)).not.toContain('space-hop');
+  });
+
   it('never stores a game identity, even if a client sends one', async () => {
     await post(app, {
       visitId,
--- a/apps/api/src/visit-telemetry.ts
+++ b/apps/api/src/visit-telemetry.ts
@@ -88,6 +88,7 @@ const EventSchema = z.discriminatedUnion('type', [
   }),
   z.object({ type: z.literal('route_viewed'), route: RouteKindSchema, ...offsetField }),
   z.object({ type: z.literal('play_started'), ...offsetField }),
+  z.object({ type: z.literal('how_to_play_opened'), ...offsetField }),
   z.object({ type: z.literal('create_step'), step: CreateStepSchema, ...offsetField }),
   z.object({ type: z.literal('waitlist_step'), step: WaitlistStepSchema, ...offsetField }),
 ]);
@@ -182,6 +183,8 @@ export async function registerVisitTelemetryRoutes(
         case 'create_step':
         case 'waitlist_step':
           return { ...base, type: event.type, step: event.step };
+        case 'how_to_play_opened':
+          return { ...base, type: event.type };
         default:
           return { ...base, type: 'play_started' };
       }
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -930,6 +930,8 @@ export function App() {
                 orientation={stageContent.game.orientation}
                 reportSlug={stageContent.game.slug}
                 submittedBy={stageContent.game.submittedBy}
+                controls={stageContent.game.controls}
+                touch={stageContent.game.touch}
               />
             )}
 
--- a/apps/web/src/GameTheater.test.tsx
+++ b/apps/web/src/GameTheater.test.tsx
@@ -171,11 +171,51 @@ describe('GameTheater how-to-play', () => {
 
     // The panel portals to the body, not into the theater's own subtree.
     expect(document.querySelector('.howto-card')).not.toBeNull();
-    expect([...document.querySelectorAll('.howto-list li')].map((li) => li.textContent)).toEqual([
-      'Left/Right to move',
-      'Space to fire',
-      'M to mute',
+    expect(
+      [...document.querySelectorAll('.howto-row')].map((row) => [
+        row.querySelector('dt')?.textContent,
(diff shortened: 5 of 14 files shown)


gtanczyk:
@codex

gtanczyk (changes requested):
Thanks for this — it's careful work. The parser threshold measured against the real catalog, the Escape precedence handled on both the window listener and the sandbox bridge, the backdrop-filter/containing-block portal reasoning, and the slug-free telemetry all check out against the code. The host-chrome placement is right: `HIDE_CHROME` hides the game's own `.game-controls`/`.hint`, and this card is the replacement on the correct side of the sandbox boundary.

Two things to fix before merge:

**1. The Tab trap leaks forward when focus is outside the card** (`HowToPlayPanel.tsx`)

The `!cardRef.current.contains(active)` escape hatch exists only on the Shift+Tab branch. Click on the card's non-focusable text (a `dt`, the title) and `document.activeElement` becomes `body`; a forward Tab then matches neither branch, isn't prevented, and focus walks into the chrome behind the running game — the exact failure the trap exists to stop. Mirror the condition:

```js
} else if (!event.shiftKey && (active === last || !cardRef.current.contains(active))) {
```

The existing trap test passes only because focus is still on the close button; worth a case that blurs to `body` first.

**2. Rebase onto master** — the branch is ~56 commits behind and there's one merge conflict, in `apps/api/src/store.ts` (master added more `VisitEvent` union members). Everything else auto-merges. CI can't be trusted until this lands.

Two small things worth folding in while you're there (not gates):

- A wide (unsplit) row renders a `dl > div` containing only a `dd`. HTML wants each group to be dt(s)-then-dd(s), and a screen reader announces an orphan definition with no term. Either move wide clauses to a `<p>` outside the `<dl>` or give them a visually-hidden `dt`.
- The new CSS block is inserted between the `/* AUTH MODAL */` header comment and `.modal-backdrop`, so that header now labels the wrong section. Move the block above it.

Non-blocking observations, noted so they don't read as missed:

- `touch: 'native'` and `'controllers'` produce no output — `arena-tag` (controllers) does support touch and the card, which reads as the complete answer, stays silent about it. Follow-up-sized.
- Escape handling lives in the parent rather than the panel, so a second call site would silently lose it; and `howToPlayDismiss` doesn't mention Escape, which is the fastest way out on desktop.
- `id="howto-title"` is fine as a singleton; `useId()` would remove the assumption for free.
- Worth an eyeball of the bar at ~800px wide — labels only hide below 768px, and the bar there now carries How to play + a 112px-min Sound + vote + fullscreen + More + Exit.
- The `/play/` deep-link gap: agreed with deferring rather than breaking `App.catalog.test.ts`; we'll track it as a follow-up issue since shared links are exactly where a player has the least context.

---

jactan77:
Thanks — both gates are addressed, and the ~800px "eyeball" turned out to be a real defect rather than a look-over.

**1. Tab trap leaked forward.** Confirmed and fixed exactly as suggested: the `!contains(active)` hatch now guards both branches. You were also right that the old test proved nothing — it passed only because focus was still on the close button. There is now a case that blurs to `body` first and asserts both directions pull focus back into the card.

**2. Rebased onto master.** The one conflict was `store.ts`'s `VisitEvent` union, where master had added `waitlist_step`; kept both members. Everything else auto-merged, including the two telemetry unions. Full gate green on the rebased branch: type-check, lint, 89+91 test files, build.

**Both folded in:**

- Unsplittable clauses now carry a term (`Control`), visually hidden and read aloud, so no `dl` group is a `dd` without a `dt`. I went with the hidden-`dt` option rather than moving those rows to a `<p>`, because the split and unsplit clauses interleave within a game (`biplane-skirmish` splits "Click to lock mouse look" but not "W/S pitch"), and pulling some out would reorder the controls.
- The CSS block moved above `/* AUTH MODAL */`, which it had been silently relabelling.

**On the ~800px bar — you were right to ask, and it was worse than cramped.** Measured in a browser at 800px: the title's right edge is 184px and the vote widget's left edge is 179px, so they overlap. Hiding only `.howto-btn` moves the widget back to 333px, which makes it this feature's doing rather than pre-existing — the meta column does not shrink, so anything added to the actions row eats the title. The bar copy now sheds to the More menu at 900px instead of the shared 768px; verified reachable at 1280 / 880 / 500px, with a test asserting both copies render so neither breakpoint can strand it.

Also took the free one: `howToPlayDismiss` now names Escape.

**Left as follow-ups, as you framed them:** `touch: 'native'` / `'controllers'` staying silent (agreed it reads as a complete answer when it isn't — `arena-tag` is the case in point), Escape living in the parent rather than the panel, and `useId()`. Happy to take any of them here if you would rather not carry them.

Screenshots in the description are refreshed from the fixed build.

gtanczyk (approved):

PR #381, opened 2026-07-31
feat(party): voice on phone controller + voice module contract

gtanczyk:
## Summary

Productionizes **voice-on-phones Layer 1** (and the serve-side half of Layer 0):

- Optional **Voice** toggle on the party phone controller maps short en/pl phrases onto the existing `up/down/left/right/a` keys (pad remains primary).
- Assemble-contract bump: add GameKit `voice` module + **8 KiB** platform reserve (`maxProjectBytes` 441687 → **449687**). **Merge this before** the games-repo PR (`www.gamedev.pl-games` voice loudness).
- `GameFrame` gains `allow="microphone"` so opt-in loudness games can request the mic inside the sandboxed iframe (player still taps Enable mic).
- CI: `contract:games-repo` accepts website-ahead (budget / trailing modules) so this PR can go green before games `main` catches up; still fails when the website is behind.

Paired games work: shout-dash pilot + `createVoiceMeter` (branch `cursor/voice-loudness-production-9296`, PR gamedevpl/www.gamedev.pl-games#272).

## Test plan

- [x] `vitest` contract / GameFrame sandbox / voice phrase map / locale parity
- [x] `npm run type-check` + `npm run lint`
- [x] `assemble.test.ts` + `games-repo-contract-check` website-ahead cases
- [ ] Manual: join a party lobby on Chrome Android / Safari iOS, toggle Voice, say “left” / “lewo” / “go”
- [ ] Confirm Firefox shows unsupported notice and pad still works

<div><a href="https://cursor.com/agents/bc-bf5adf4d-35a4-414f-a92c-5e3022509296?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bf5adf4d-35a4-414f-a92c-5e3022509296&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #390, opened 2026-07-31
Raise serve cap to 473,687 for the games-repo world + gameplay reserves

gtanczyk:
Website half of a cross-repo contract move, paired with
**gamedevpl/www.gamedev.pl-games#278**.

## ⚠️ Merge order: this side first

`games-repo-contract.ts` documents why, and it is not order-neutral. Merging the games
repo first opens a window where the build ceiling sits above the serve ceiling, and a
game that grows into that gap bakes as a failure *quietly* — a partial bake leaves the
pointer on the previous snapshot, so the site keeps serving and the merged game simply
never appears. Website-first cannot do that: a serve cap above the build cap means no
assemblable game is refused here. `contract:games-repo` now explicitly tolerates
website-ahead, which is exactly the state this leaves the repos in.

## What changed

`MAX_PROJECT_BYTES` 449,687 → 473,687, via `GAMEKIT_PLATFORM_BYTES` 244,887 → 268,887.

- `apps/api/src/games-repo-contract.ts` — the constant plus a history entry
- `apps/api/fixtures/games-repo/shared/assemble-contract.json` — vendored snapshot gains
  `world: 15000` and `gameplay: 9000`; `maxProjectBytes` follows
- `apps/api/src/games-repo-contract.test.ts` — expected total, and the extractor fixture
  that mirrors games-repo `validate.ts` allowance for allowance
- `apps/api/src/assemble.test.ts` — the size-cap assertion (its own comment says to move
  it to match the games repo rather than relax it, which is what this does)

**Rebased onto the `voice` reserve.** #381 landed the +8,000 `voice` reserve
(441,687 → 449,687) and the `voice` module while this branch was open. Both are carried
forward untouched and the Pass #10 raise stacks on top of them, rather than either side
clobbering the other.

## Why the games repo needs it

Games-repo convergence Pass #10 promotes `carjack-city`'s generic systems into GameKit —
arcade vehicle physics, a 2D chase camera, loop paths, spiral search, a combo chain, a
message ticker, tile/hull collision resolvers, and the sim→presentation frame-events bus.
Those live in the `world` and `gameplay` modules, which shipped tiny and were charged to
the 200 KiB author budget.

`carjack-city` cannot afford them there. It is the tightest bundle in the catalog: #276
("make the cars the point") had to re-encode the engine sample at 11 kHz to claw back
2,646 bytes just to fit its own new car code. Both modules therefore graduate to named
opt-in reserves, the same argument that gave `zone` one in games-repo #163 and `voice`
one in #381: a module nobody pays for unless they ask for it gets a reserve rather than a
slice of another game's author budget.

Measured transpiled at the Pass #10 tip: `world` 13,263 → 15,000 reserve; `gameplay`
7,371 → 9,000 reserve. With both selected, carjack-city assembles at 459,041 — 6,646
under the new ceiling. `GAME_BUDGET_BYTES` is unchanged at 200 KiB, so this does not let
games grow.

## Verification

`npx vitest run` in `apps/api`: **91 files, 1,548 tests, all passing.**

### DIFF
--- a/apps/api/fixtures/games-repo/shared/assemble-contract.json
+++ b/apps/api/fixtures/games-repo/shared/assemble-contract.json
@@ -37,9 +37,11 @@
     "gfx3d": 96000,
     "lookControls": 4683,
     "spatialAudio": 2977,
-    "zone": 12000
+    "zone": 12000,
+    "world": 15000,
+    "gameplay": 9000
   },
-  "maxProjectBytes": 441687,
+  "maxProjectBytes": 465687,
   "audio": {
     "musicField": "string",
     "musicCatalogRequiredKeys": ["version", "tracks"],
--- a/apps/api/src/assemble.test.ts
+++ b/apps/api/src/assemble.test.ts
@@ -19,7 +19,7 @@ describe('the size cap', () => {
   it('matches the budget the games repo enforces in Check 4', () => {
     // Sourced from games-repo-contract.ts; CI re-checks the live games-repo
     // MAX_BUNDLE_BYTES when GAMES_REPO_TOKEN is set (issue #247).
-    expect(MAX_PROJECT_BYTES).toBe(441_687);
+    expect(MAX_PROJECT_BYTES).toBe(465_687);
   });
 
   it('accepts a game that fills the budget', () => {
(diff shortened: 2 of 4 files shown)


gtanczyk:
@codex

PR #397, opened 2026-07-31
Add a 2D/3D switch to the prompt card, and route it to the agent

jactan77:
The creator can now say which dimension they want before pressing create. The control sits on its own toolbar row at the top of the prompt card rather than in the action bar: the wanted dimension is a setting on what gets built, and beside "Build My Game" it read as part of the submit control instead.

Shaped as a switch (`role="switch"`), not a segmented radiogroup, because the two options are not peers. 2D is the default and the light path; 3D is an opt-in onto the heavier one, where the agent has to select the `gfx3d` GameKit module, which draws on its own byte reserve rather than the author budget. The knob only takes the accent colour on 3D for the same reason — colour marks the opt-in, not the state. Side labels are pointer-only shortcuts, kept out of the tab order so one control does not become three tab stops.

The choice is only real if it survives to the agent, so it is threaded the whole way: hero -> App -> the confirm step (where the game is named) -> POST /api/submissions -> the brief. It rides inside the pending spec rather than as a loose argument, which is what makes it survive both the naming round and the localStorage restore after a reload.

In the brief it lands in its own "Build requirement" section outside the fenced creator block, since everything in that block is untrusted text the agent is told to treat as data, and this line is the platform telling it what to build.

The refiner is told the dimension too, so it stops spending one of its four clarifying questions re-asking something answered by a switch. That value joins the refine cache key: the same concept asked in 2D and in 3D can need different questions, and sharing one cached answer would serve the wrong set to whoever arrived second.

Instrumented as `dimension_3d_selected` — one extra funnel step rather than a new event shape. It is a branch off `prompt_started`, not a rung, and is labelled and ordered so it does not read as a drop-off between two stages. 2D is deliberately not recorded: it is where every visit starts, so a row for it would restate the starter count.

Absent `dimension` means 2D, so an older web build still submits and lands on the behaviour every game published before this already has.

Also writes down the convention that visual changes ship with screenshots and a GIF captured from the running app, along with the three traps that cost time getting them: measure the crop in the session that records, clear the sticky header when scrolling, and Playwright's bundled ffmpeg has no GIF muxer.


gtanczyk:
@jactan77 This loads too many toggles to the main form, lets move the decision about 2d vs 3d, to creator QA, sometimes it won't be even needed
Agent will ask for 2d vs 3d if needed

gtanczyk (changes requested):
Lets move the decision to Creator QA, which means slightly modifying the QA prompt to ask the 2d vs 3d question more often

PR #401, opened 2026-07-31
Make the studio-shell browser gate assert without a shelf (#392)

### LINKED ISSUE #392: The studio-shell browser gate passes without asserting anything
`apps/e2e/src/studio-shell.test.ts` guards the shape of the Creator Studio thread against the deployed candidate: the page must not scroll, the transcript must keep its own scroller, and nothing anchored to the bottom of the window may cover the composer. It runs at four viewport widths.

All four skip. From the browser gate on the #391 merge (run [30646772652](https://github.com/gamedevpl/www.gamedev.pl/actions/runs/30646772652)):

```
[e2e] studio shell: the e2e account has no games open in the studio; layout check skipped   (phone)
[e2e] studio shell: the e2e account has no games open in the studio; layout check skipped   (narrow tablet)
[e2e] studio shell: the e2e account has no games open in the studio; layout check skipped   (wide tablet)
[e2e] studio shell: the e2e account has no games open in the studio; layout check skipped   (desktop)
✓ src/studio-shell.test.ts (4 tests) 17203ms
```

The early return is deliberate — a guard that silently stops guarding is worse than no guard, so it warns rather than passing quietly — but the effect is that the gate reports four green ticks while checking nothing. The bugs it exists to catch have shipped twice (#386, then again in #391 at every width above 800), so this is not a hypothetical guard.

## What it needs

A game on the `bot:e2e` identity's shelf, so `/studio` opens a thread with a composer in it. Any published or in-progress game will do; the test does not care which. That costs one real agent build.

## Worth deciding at the same time

Whether the skip should become a failure once the account is seeded. Right now an account that loses its games silently returns the gate to a no-op. Options:

- Fail the test when the account has no games — strongest, but couples the gate to shelf state that other runs can change.
- Fail only in CI (`process.env.CI`), keep the warning locally.
- Leave it as a warning and rely on someone reading gate output — which is what has been happening.

## Where

- `apps/e2e/src/studio-shell.test.ts`
- The shell contract it guards: the unconditioned `.app:has(.studio-layout.is-game-open)` block in `apps/web/src/styles.css`
- Background on the failure mode: the "Layout changes" section of `AGENTS.md`

gtanczyk:
Fixes #392.

## Problem

`apps/e2e/src/studio-shell.test.ts` is meant to guard the Creator Studio shell layout (page must not scroll, transcript needs its own scroller, bottom bars must not cover the composer) at four viewport widths. When `bot:e2e` had no games on its shelf, all four checks **skipped with a warning** while the deploy browser gate still went green — so the bugs this gate exists for shipped twice (#386, then again in #391).

## Approach

Instead of seeding a real game onto the e2e identity (and coupling the gate to mutable shelf state), the suite now **stubs** the JSON those views fetch:

- `GET /api/me/studio` → one fixture game
- `GET /api/me/studio/health` and `/scorecards` → empty
- `GET /api/submissions/<fixture-token>` → published status (composer still mounts for any non-abandoned state)

The real React tree (`CreatorStudioView` + embedded `SubmissionStatusView`) and the real stylesheet still run. If the fixture fails to put a composer on screen, the test **fails** instead of skipping.

This stays read-only against production (no `POST /api/submissions`, no agent build) and answers the issue’s “should skip become a failure?” question by making an empty assertion impossible.

## Docs

- Short note in `apps/e2e/README.md`
- Pointer in `AGENTS.md` layout section

## Verification

- `npm run type-check && npm run lint && npm run test && npm run build` — green
- `npm run type-check -w @gamedevpl/e2e` — green
- Full `npm run e2e` against a candidate needs `GAMEDEV_ACCESS_TOKEN` (not available in this agent env); expected to exercise the four viewport cases for real on the next deploy gate run

<div><a href="https://cursor.com/agents/bc-7f6e3ab7-ee4d-4f2c-9240-e3e63ddcdfa6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-7f6e3ab7-ee4d-4f2c-9240-e3e63ddcdfa6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,7 +85,8 @@ When one shape holds across widths, write it **once, outside every media query**
 media blocks carry only what genuinely differs. A rule that must never be forgotten cannot
 live in a place where it has to be remembered twice. The same applies to the test: a guard
 pinned to a single viewport stops guarding the moment the layout gains a second one, so cover
-each band the CSS assembles differently — see `apps/e2e/src/studio-shell.test.ts`.
+each band the CSS assembles differently — see `apps/e2e/src/studio-shell.test.ts`
+(that gate stubs the shelf/status JSON so an empty `bot:e2e` shelf cannot skip the check).
 
 ## Current architecture
 
--- a/apps/e2e/README.md
+++ b/apps/e2e/README.md
@@ -88,6 +88,16 @@ Never work around this by disabling TLS verification or unsetting `HTTPS_PROXY`.
 [`.claude/skills/browse-live-site/SKILL.md`](../../.claude/skills/browse-live-site/SKILL.md)
 for driving the same setup by hand.
 
+## Studio shell layout gate
+
+`src/studio-shell.test.ts` checks that an open Creator Studio thread owns the window
+(no page scroll, transcript scroller of its own, bottom bars never cover the composer)
+at every CSS band. It **stubs** `/api/me/studio` and the fixture game's status response
+rather than borrowing whatever happens to sit on `bot:e2e`'s shelf: an empty shelf used
+to make all four widths skip while the deploy still went green (#392). The real React
+tree and stylesheet still run; only the JSON is replaced, so the suite stays read-only
+against production.
+
 ## Adding a test
 
 Two rules, both learned the hard way:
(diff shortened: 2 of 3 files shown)


gtanczyk:
@codex

PR #430, opened 2026-08-01
byoca(BY-05): remote MCP endpoint

gtanczyk:
## Why

BYOCA M1 needs a streamable-HTTP MCP server on the API so Claude Code / Codex / Cursor / Kimi can drive a self-build round without a published npm package.

## What

- `POST|GET|DELETE /api/mcp` — MCP streamable HTTP wrapping the existing agent build channel
- Auth (security core):
  - `Mcp-Session-Id` is transport-only — **never** authority
  - `start({ key })` validates the round key and returns a short-lived HMAC `sessionKey` (session id + jobId + roundGeneration + exp)
  - Every other tool requires `sessionKey` **or** `Authorization: Bearer <round key>`, verified per call
  - When a transport session id is present, it must match the `sessionId` bound into the sessionKey
  - Terminal receipt: `get_gate_verdict` readable one generation behind for that delivery only; writes/other reads reject
- `/api/mcp` exempt from the private-beta session wall (agents have no site session)
- Channel inject propagates caller IP so rate limits are not shared on `127.0.0.1`
- Shot ceiling is **700 KiB decoded** (Firestore 1 MiB doc + base64) with matching Fastify `bodyLimit`; true 2 MB needs object storage
- `MAX_UPLOAD_FILES` / MCP `submit_sources` aligned at 200

## Merge hold

**Do not merge until owner security review / verdict.** This is the single largest security surface in the project.

## Review follow-ups addressed

Copilot + Codex + supervisor required changes (sessionId binding, shot bodyLimit, beta wall, IP inject, Firestore-safe shot cap, file-limit alignment, schema Bearer mode, map pruning).

## Test plan

- [x] Green gate locally for touched tests
- [x] MCP auth + tool + receipt + beta-wall + sessionId-mismatch tests
- [ ] Owner security review before merge

<div><a href="https://cursor.com/agents/bc-fc0fdc2f-be8a-48de-81cb-d2b23d54845d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fc0fdc2f-be8a-48de-81cb-d2b23d54845d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/api/src/agent-channel.ts
+++ b/apps/api/src/agent-channel.ts
@@ -7,7 +7,15 @@ import {
   DEFAULT_BUILD_ORIENTATION,
 } from './agent-build-brief.js';
 import { getAgentBuildExample, listAgentBuildExamples } from './agent-build-examples.js';
-import { assertAgentTokenActive, InvalidAgentTokenError, readBearerToken, verifyAgentToken } from './agent-token.js';
+import {
+  assertAgentTokenActive,
+  classifyAgentTokenAccess,
+  InvalidAgentTokenError,
+  readBearerToken,
+  STALE_AGENT_TOKEN_REASON,
+  verifyAgentToken,
+  type AgentTokenAccess,
+} from './agent-token.js';
 import { selfBuildDeliveryCap } from './builder.js';
 import { DEFAULT_SIGNED_URL_TTL_SECONDS, type GcsObjectStore } from './gcs-sign.js';
 import { InvalidUploadError, type GamesStore } from './games-store.js';
@@ -70,8 +78,11 @@ const AckRequestSchema = z.object({
 });
 
 const MAX_SHOT_LABEL = 120;
-/** ~300 KB decoded, which a pixel-art frame sits well inside. */
-const maxShotBytes = 300 * 1024;
+/**
+ * 2 MB decoded — the MCP `send_screenshot` contract (BY-05). Pixel-art frames sit
+ * well inside; a full-resolution capture from a kit check may not.
+ */
+const maxShotBytes = 2 * 1024 * 1024;
 const PNG_SIGNATURE = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]);
 
 const BuildShotInputSchema = z.object({
@@ -134,7 +145,8 @@ const BuildSourcesInputSchema = z.object({
       }),
     )
     .min(1, 'files is required')
-    .max(64, 'too many files'),
+    // MCP submit_sources allows ≤200; the filename allowlist + byte budget still bite.
+    .max(200, 'too many files'),
   /**
    * Creator Kit engineRef the sources were built against. Required for self-build
    * deliveries (BY-06); the gate compares it to `kits/current.json`'s N/N−1 window.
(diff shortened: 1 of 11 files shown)


gtanczyk (commented):
**Supervisor security review at `e4a521b`: approve after two small pre-merge changes.** This got the slow read it was promised — the auth core is spec-compliant and correctly built, and the gate is green in my worktree (**1685/1685 API tests, type-check, lint**).

**What holds up under adversarial reading:**
- `Mcp-Session-Id` is never authority anywhere: `resolveAuth` requires a `sessionKey` argument or a Bearer round key on every call, with no session-id fallback path. The explicit error for a session-id-only call says so.
- The `sessionKey` is a distinct-scope HMAC (`mcp-session-v1`) over `(sessionId, jobId, roundGeneration, exp)`, timing-safe compared, expiry-checked, and — critically — the **generation is re-classified against the job document on every call**, so a round closing revokes mid-session. `start` refuses receipt-class keys (`access !== 'active'`), so a one-behind key cannot open a fresh session.
- The terminal receipt matches the transport-independent spec from the design review: `classifyAgentTokenAccess` grants `terminal_receipt` only for unexpired capabilities exactly one generation behind, only the gate-verdict read accepts it (`allowTerminalReceipt`), and receipt mode is pinned to the job's current `deliveredVersion` — asking for any other version 401s. Tested on all three transports, plus the forged/absent-sessionKey spec-compliance case, expired keys, invalid-`start` IP limiting, oversized payloads, and the scripted start→brief→seed→submit→verdict flow.
- The internal ephemeral channel token minted for `inject` never leaves the server.

**Change 1 (required): enforce the sessionId binding you already signed.** `resolveAuth` verifies the sessionKey's signature but never compares `sessionClaims.sessionId` to the request's `Mcp-Session-Id`, so the binding the capability encodes is currently decorative — a sessionKey lifted from logs works from any session. Rejecting a mismatch (when a transport session id is present) is a few lines, makes a leaked sessionKey useless without also replaying the session id, and matches the execution plan's stated design. If there's a client-compat reason not to (session ids regenerating across reconnects), document it at the verification site instead — but decide explicitly, don't leave it silent.

**Change 2 (required, one line): the raised 2 MB screenshot cap is unreachable on the plain channel.** `maxShotBytes` went to 2 MB for all callers, and the MCP route got `bodyLimit: 3 MB` — but `/api/agent/build/shot` keeps fastify's 1 MB default, so a direct-channel caller sending what the schema now promises gets a transport-level 413 before validation. Either raise that route's `bodyLimit` to match, or scope the 2 MB claim to MCP.

**Minor, document-only:** a sessionKey (24 h TTL) can outlive the round key's own `exp` — generation revocation still bounds it, but the comment should say the creator-visible expiry is not the sessionKey's ceiling.

On the channel-wide limit raises (also asked by the owner): **unify — with the bodyLimit fix.** The 200-file cap is safely bounded by the existing byte budget and filename allowlist regardless of caller; the 2 MB shot cap is legitimate for full-resolution kit captures, applies only to authenticated callers, and stays bounded by the per-build shot rate caps. There's no reason Copilot sessions should live under a stricter media cap than MCP agents — but only if the transport actually accepts the bytes (change 2).

---

PR #434, opened 2026-08-01
Mount a catalog card's video only when somebody asks to watch it

jactan77:
Scrolling the arcade stutters, and gets worse with every game published.

A card mounted a `<video>` as soon as it came near the viewport, and `useInView` is sticky, so a full-page scroll of sixty games left sixty media players in the document — none of which anybody had asked to watch. Chromium allocates a real player per element, and that is what the stutter was.

Measured over a scripted full-page scroll of a sixty-card catalog, three runs each, same machine and same session:

                       before        after
  dropped frames       33 / 335      4 / 332
  long tasks           5             0
  time in long tasks   380 ms        0 ms
  worst frame          83 ms         33 ms
  95th pct frame       33 ms         17 ms

The poster image already renders in the same box, so the card carries itself until the player is wanted. Nothing about the card changes visually: the screenshot in docs/media/catalog-scroll is byte-identical before and after, taken from the running app — one has a `<video>` in the document and the other does not.

Two things fall out of deferring the mount.

`preload` moves from `metadata` to `auto`. The element now only exists because somebody asked for it, so fetching the clip is the point rather than waste, and starting the fetch in the same gesture is what buys back the round trip the deferral costs.

The toggle had to stop following the element and start following the viewer. `isPreviewPlaying` cannot be true before the element exists, so a press now sits in a gap where the button still reads "Watch preview" — which looks like a missed press, and on a slow line invites a second press that cancels the first. A `previewWanted` state carries the intent, leads the element's own state, and is what the label, the icon and `aria-pressed` read. jsdom never reaches `canplay`, which makes it an exact stand-in for that slow line in the test.

Not done here, with the numbers to decide on: the four moment thumbnails per card are the whole of what is left. Removing them takes the four dropped frames to roughly one, but deferring them the same way would hide the strip on touch, where there is no hover to reveal it — the honest fix there is serving thumbnails at thumbnail size rather than decoding four full screenshots per card.

`content-visibility: auto` on the cards was measured and dropped: it halves the long-task time on its own, but adds nothing once the players are gone.


gtanczyk:
What do we show before someone clicks on play? Should it work differently on mobile than on desktop? So that you dont need to tap all the games, and we usually have 1-2 players in the viewport

gtanczyk:
Fixed in https://github.com/gamedevpl/www.gamedev.pl/pull/497

PR #447, opened 2026-08-02
Convert CreatorQA from inline panel to full-screen wizard

gtanczyk:
Redesigns the game creation confirmation flow from an inline panel below the hero into a full-screen modal wizard with one decision per screen.

## Summary
The CreatorQA component has been restructured from a single-page form into a multi-stage wizard interface. This change improves UX on mobile devices by reducing cognitive load (one question per screen) and eliminates UI competition between the hero's "Build my game" button and the confirmation panel's submit button.

## Key Changes

- **Wizard Architecture**: Converted to a stage-based system with distinct screens for: name input → questions → builder choice → review summary
- **Full-Screen Modal**: Component now renders as a portal to `document.body` with fixed positioning, preventing scroll conflicts and ensuring consistent presentation across devices
- **One Decision Per Screen**: Each question/choice occupies its own dedicated screen; selections no longer auto-advance (explicit Next/Skip buttons required)
- **Review Stage**: Added a final review screen showing all answers with edit buttons to jump back to specific stages
- **Navigation**: 
  - Back button to return to previous stage (disabled on name stage)
  - Next/Skip button that changes label based on whether current question is answered
  - Shortcut button to jump directly to review (skipping remaining questions)
  - Progress indicator showing current stage and total stages
- **Accessibility**: 
  - Stage headings receive focus on navigation for screen reader announcement
  - Progress bar with visual and semantic indicators
  - Proper ARIA labels for modal dialog
  - Legend hiding in BuilderChoice when context already provides the question
- **Styling**: Complete redesign of `.qa-*` classes to support full-screen layout, modal header/footer, stage animations, and improved touch targets (44px minimum)
- **Test Updates**: Refactored test helpers to work with document-wide queries (wizard portals outside component container) and updated assertions to match new wizard structure

## Implementation Details

- Body overflow is locked while wizard is open to prevent background scrolling
- Scroller resets to top on each stage change
- Name field gates progression to subsequent stages
- Multi-choice questions accumulate selections; single-choice replace them
- Custom text input combines with selected options rather than replacing them
- Submission state persists on review stage rather than unmounting the component

https://claude.ai/code/session_015wgLLexRUKneRr6cCYAD8R

### DIFF
--- a/apps/web/src/App.qa.test.ts
+++ b/apps/web/src/App.qa.test.ts
@@ -130,9 +130,31 @@ async function submitIdea(container: HTMLElement) {
   });
 }
 
+/**
+ * The confirm wizard portals itself to document.body, so it is never under the
+ * container the app was mounted into — everything about it is queried document-wide.
+ */
+const inWizard = <T extends Element>(selector: string): T | null => document.querySelector<T>(selector);
+const wizardOptions = () => document.querySelectorAll<HTMLButtonElement>('.qa-option');
+
+/** One progress cell per stage: the name, one per question, the builder, the review. */
+const stageCount = () => document.querySelectorAll('.qa-wizard-progress span').length;
+
+/** Advance one stage with the footer's primary button. */
+async function advance() {
+  await act(async () => {
+    inWizard('.qa-next')?.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+    await flushEffects();
+  });
+}
+
+/** The wizard opens on the name stage; the questions are one step in. */
+const gotoFirstQuestion = advance;
+
 describe('the QA gate in App', () => {
   afterEach(() => {
     document.body.innerHTML = '';
+    document.body.style.overflow = '';
     localStorage.clear();
     window.history.pushState(null, '', '/');
     vi.restoreAllMocks();
@@ -147,13 +169,14 @@ describe('the QA gate in App', () => {
     const first = await renderApp();
     await submitIdea(first.container);
 
-    expect(first.container.querySelector('.qa-container')).not.toBeNull();
-    expect(first.container.querySelectorAll('.qa-card')).toHaveLength(2);
+    expect(inWizard('.qa-wizard')).not.toBeNull();
+    // Name, the two questions, the builder, the review.
+    expect(stageCount()).toBe(5);
 
     // Answer one question, then throw the whole app away — the reload.
-    const chips = first.container.querySelectorAll<HTMLButtonElement>('.qa-chip');
+    await gotoFirstQuestion();
     await act(async () => {
-      chips[0].dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      wizardOptions()[0].dispatchEvent(new MouseEvent('click', { bubbles: true }));
       await flushEffects();
     });
     await act(async () => first.root.unmount());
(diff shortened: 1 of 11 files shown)


gtanczyk:
All four review findings were real and are fixed in 010ecb0.

**Wizard sat below the app chrome** (P1, raised by both reviewers) — confirmed: `.qa-wizard` was `z-index: 90`, under the sticky `.app-header` (100) and the `.install-prompt` / `.app-update` banners (900). The header painted over the wizard's own header and the banners covered its footer navigation. Now `1200` — above both, deliberately still below `.modal-backdrop` (10000) so a sign-in prompt lands over the wizard rather than under it.

**Focus escaped the overlay** — correct that `aria-modal` doesn't contain focus, and the shell behind a covering element stays tabbable. Tabbing past the last control landed on the hero prompt box and build button that this screen exists to take over from. Tab and Shift+Tab now cycle within the wizard, and focus returns to whatever opened it on close.

**Icon-only exit was unnamed on mobile** — confirmed: below 560px the CSS hides the label and `PixelIcon` renders the glyph `aria-hidden`, so the only route back to editing was an unnamed button. It now carries an explicit `aria-label`.

**Misleading footer comment** — the claim that the shortcut "hugs the right edge" isn't what happens (with auto margins on both children flexbox splits the free space equally, so it lands midway), but the comment did overstate it as centering. Reworded to describe the actual behaviour.

Three regression tests added for the focus trap, focus restoration, and the exit button's accessible name. 863 tests pass; type-check and lint clean.

---

gtanczyk:
Second review pass addressed in c386818 — and it flushed out a bug in my own first-round fix.

**Top safe-area inset** (P1) — correct. `index.html` sets `viewport-fit=cover`, and this overlay covers `.app-header`, so it inherits that header's obligation. Installed on a notched iPhone the step label and the only exit control sat under the status bar with taps intercepted. Now `max(12px, env(safe-area-inset-top))`, matching the nine other rules already doing this.

**Correcting myself on the z-index.** My previous fix used `1200` on the claim that `.modal-backdrop` was `10000` — it isn't. `10000` is `.sketch-modal-backdrop`; the auth modal's backdrop is **1000**, as are `.howto-backdrop` and `.stage.is-playing-full-viewport`. So `1200` fixed the reported bug and introduced a new one: it sealed a sign-in prompt behind the wizard. Now **950** — above `.app-header` (100) and the install/update banners (900), below the 1000 tier that has to be able to appear over it.

Added `styles.qaWizard.test.ts` to stop this being re-litigated by hand: it reads the stylesheet and pins the wizard between those two groups by comparing against its neighbours' *real* values rather than hard-coded numbers, and asserts both safe-area insets are paid. That test is what caught the 1200 overshoot — it failed the moment I wrote it.

866 tests pass; type-check and lint clean.

---

gtanczyk:
Third pass addressed in 44b54c6 — a real hole in the focus trap I added last round.

The trap collected the overlay's focusable controls and returned early when it found none. `submitting` disables every button and input in the wizard, so that branch was reached in exactly the state the overlay is least meant to be escapable — and returning handed Tab back to the browser, which walked straight into the app shell behind. Good catch, and specifically the kind that only shows up once the first fix exists.

The empty case now swallows the keypress and parks focus on the overlay root (`tabIndex={-1}`) until a control comes back. Covered by a test that renders with `submitting: true`, asserts nothing enabled remains, and checks Tab leaves focus inside.

Also, for the record on the earlier "Production image" failure: that was a Docker Hub registry timeout pulling `node:20-slim` (`dial tcp 3.210.130.203:443: i/o timeout`), unrelated to this diff, which touches only CSS and tests. I re-ran the job and it passed — all 7 checks green.

867 tests pass; type-check and lint clean.

---

PR #452, opened 2026-08-02
byoca(BY-24): agent-opened improvement rounds via open_round

gtanczyk:
## Summary
BY-24: MCP `open_round` for agent-opened post-publish improvement rounds (opt-in, durable game key, same daily improvement quota as Studio).

### Review follow-ups (`1cc3cb0f`)
- **Atomic admission:** `gameAgentKeys.agentOpenRoundPending` + `beginAgentOpenRound` / `finishAgentOpenRound` so concurrent `open_round` cannot double-create jobs or double-spend quota
- **Transactional opt-in:** merged BY-23 tip — `setGameAgentOpenRounds` is one Firestore transaction
- **Studio rotate:** show/copy fresh kickoff after rotate; `connect_copied` only on actual copy
- **Telemetry:** `round_opened` on first status snapshot when `openedBy` is already set

Stacked on #450 (`byoca/per-game-agent-key`).

## Test plan
- [x] `npm test --workspace=apps/api -- mcp-open-round`
- [x] `npm test --workspace=apps/web -- StudioAgentKeyPanel SubmissionStatusView`
- [x] Concurrent `open_round` admits one job / one quota spend

<div><a href="https://cursor.com/agents/bc-fc0fdc2f-be8a-48de-81cb-d2b23d54845d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fc0fdc2f-be8a-48de-81cb-d2b23d54845d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #456, opened 2026-08-02
Roll up editor_step into the visit funnel

gtanczyk:
Follow-up to #448. The editor funnel has been recording `opened` / `draft_saved` / `previewed` / `published` since the Studio editor shipped, and nothing could read it. This is the read half.

## The change

An **Editing** block on the operator page, beside Creating and Waitlist, with the same posture: every step, in step order, zeroes included — a rung nobody reached shows as zero rather than vanishing, because the interesting case is exactly the step where everyone stopped.

## Two decisions worth reviewing

**Shares are against the first rung, not against all visits.** Editing is meant to be rare (see the games-repo side of this work — editors stay opt-in). Against total traffic every rung rounds to 0–1%, which makes a working loop look dead. Against openers the question is the one worth asking: of the people who opened an editor, how many saved a draft, played it, and published.

**Editor steps get their own `Set` per visit.** `create_step`, `waitlist_step` and `editor_step` all carry a `step` field on the wire; separate Sets are what stop an editor save from ever reading as a create or waitlist rung. There is a test for exactly that.

The events carry no slug, so this stays a per-visit funnel and the visit stream stays unjoinable to the game stream.

## Checks

- `npx vitest run` on the affected suites — 48 tests pass
- `npm run lint` — clean

Note: `@mediapipe/tasks-vision` needed an `npm install` after #451 merged; that was a stale local `node_modules`, not a change here.

---
_Generated by [Claude Code](https://claude.ai/code/session_017kX5dgxdkYGLc8ZGAAvQMw)_

### DIFF
--- a/apps/api/src/visit-funnel.test.ts
+++ b/apps/api/src/visit-funnel.test.ts
@@ -168,6 +168,47 @@ describe('summarizeVisitFunnel', () => {
     expect(funnel.waitlist.find((row) => row.step === 'joined')?.visits).toBe(0);
   });
 
+  it('reports the editing funnel in step order, zeroes included', () => {
+    const step = (visitId: string, step: string): VisitEvent =>
+      ({ visitId, type: 'editor_step', at: '2026-07-26T10:00:00.000Z', msSinceStart: 0, step }) as VisitEvent;
+
+    const funnel = summarizeVisitFunnel([
+      started('a'),
+      step('a', 'opened'),
+      step('a', 'draft_saved'),
+      step('a', 'previewed'),
+      step('a', 'published'),
+      started('b'),
+      step('b', 'opened'),
+      // A repeated save in one visit is one visit at that rung, not two.
+      step('b', 'draft_saved'),
+      step('b', 'draft_saved'),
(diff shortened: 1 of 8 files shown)


gtanczyk:
Flagging one commit here that has nothing to do with the funnel, so it isn't a surprise in review.

`master` is red. #452 landed `const { agentOpenRoundPending: _pending, ...rest }` in two places in `store.ts`, and `Lint, Test & Build` rejects both — so every PR in the repo inherits the failure. I confirmed it reproduces on a clean `origin/master`, with no changes of mine in the tree.

The idiom is correct; the rule config is what's wrong. ESLint's base `no-unused-vars` ignores rest siblings by default and the TypeScript plugin's does not, so `ignoreRestSiblings: true` now says it explicitly. I went with that rather than a `varsIgnorePattern: '^_'` because it permits exactly the omit-a-key pattern and still rejects a genuinely dead local that happens to be underscore-prefixed.

Happy to drop it from this branch if you'd rather land it straight on `master` — say the word and I'll rebase it out.

---

PR #486, opened 2026-08-03
byoca: an improvement round must tell the agent what game it is improving

gtanczyk:
Four fixes from CP-2, the live-client checkpoint. Two are severe and compound; two are wording.

## An improvement round arrived empty, and the loop said "scaffold"

**The change request was never persisted.** `startImprovementRound` passed the creator's feedback to `dispatchBuild` and stopped. The only non-test caller of `setSubmissionBrief` was the new-submission path. For a platform build that is invisible — the backend reads the dispatch prompt. For a self build it is total: there is no backend, the creator's own agent calls `get_brief`, and `get_brief` serves the stored brief. It got `spec: ""`.

**Nothing routed the agent to `get_sources`.** Step one of `SESSION_WORKFLOW` offered `get_seed`, which is a new-game facility and absent on an improvement round; step three said to build from the kit. `get_sources` appeared exactly once in `mcp-server.ts` — its own tool definition. The tool described itself correctly and no ordered instruction ever mentioned it.

Together an improvement round read, end to end, as *scaffold a new game*. An agent following the numbered loop it was handed would replace the published game it was asked to improve — and the gate would pass the replacement, because it is a valid game. It just isn't the creator's.

`get_sources` is now step two, unconditional. It answers `available:false` on a new game, and the round type it would otherwise be gated on is not something the agent can see.

## Two refusals misdescribed what arrived

- `MCP_MISSING_CREDENTIAL_HINT` still said `<round key>` — written before creator keys and OAuth existed.
- A sessionKey offered where an opener belongs fell through to `key is required`, which tells an agent nothing was sent when something was. The mirror case already names the credential (*"this creator key only opens a session via start()"*); this direction now does too, via a shape-only `looksLikeMcpSessionKey` that checks format and never signature.

## Verification

Each new test was run against the unfixed source first and fails there — brief `expected undefined`, workflow `expected … to match /get_sources/`, sessionKey naming — with the surrounding tests passing on both sides as a control.

Full gate green: `type-check`, `lint`, api 1983 passed (129 files), web 932 passed (106 files).

## Note

Neither severe defect was reachable by a unit test, because both live in what the platform *tells* an agent rather than in what it computes. They surfaced only by driving a real MCP client through a live round.

---

### DIFF
--- a/.claude/skills/verify-agent-work/SKILL.md
+++ b/.claude/skills/verify-agent-work/SKILL.md
@@ -140,6 +140,36 @@ Two concrete instances of that (observed 2026-07-23):
   class of bug fails closed (it rejects access), so it is a contract/UX defect rather
   than a security hole — say so explicitly when reporting, so severity is not
   overstated.
+- **A fix that narrows a predicate can reintroduce the same defect through a smaller
+  door — and its regression test will not notice.** Observed (BY-25, 2026-08-02): a
+  correct fix replaced "does the creator have ANY non-abandoned job with this slug"
+  (a scan) with "is the NEWEST job for this slug owned by them and non-abandoned" (a
+  point read). Both agree in the seeded regression scenario, so a 250-job probe
+  encoded straight from the bug report passed. They disagree when a *newer* job on the
+  same slug is abandoned — a canceled improvement round, or the `no_connect` sweep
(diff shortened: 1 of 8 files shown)


gtanczyk:
@codex

PR #488, opened 2026-08-03
byoca: three promises the system did not keep

gtanczyk:
From CP-2 run against production by a real MCP client. None of the three was reachable from a unit test written against the code as it stood.

## OAuth was dead for every returning creator

The service worker answers every same-origin GET navigation with the precached shell, exempting only `/api/`. `/oauth/authorize` is server-rendered, so the SPA router — which has no route for it — rendered its NotFound page behind an HTTP 200 and the request never reached the server.

It looked healthy to `curl` and in a fresh incognito window, because neither runs a service worker. It failed for everyone else — and since minting a key requires visiting Studio, everyone else is everyone. The worker's own comment anticipated losing the 200-vs-404 distinction for *typo'd* URLs; it did not anticipate a real server route living outside `/api/`.

Server-rendered routes (`/oauth/*`, `/.well-known/*`) now bypass the worker.

## Rotation was not a kill switch

Studio promises, verbatim: *"rotating stops every agent still using the old key."* It did not. A `sessionKey` authenticates on the **round's** generation and never consults the creator key, so a session opened before a rotation kept full write access for the rest of its 24-hour life. Confirmed against production: `report_progress` succeeded after a rotation and the write landed in the creator's thread.

Rotate and revoke now advance `roundGeneration` on the creator's open rounds — the same revocation the round-close path already performs, and one `classifyAgentTokenAccess` already checks on every tool call, so nothing new has to be trusted.

Deliberately broader than the sentence promises: it also ends per-game-key sessions on those rounds. Rotation is a security action and should fail safe, and the cost is recoverable — the agent calls `start` again.

## The key panel invited the destructive button

`GET` re-minted at the wall clock, so the displayed `Ends in …` tail changed on every visit and never matched the header the agent was holding. A creator checking whether their agent was current always saw a mismatch — and the remedy the panel offers for a mismatch is **Rotate**.

A generation now mints to exactly one key, and the stated expiry is a real deadline rather than one that slid forward on every page load.

## Verification

Each new test was run against the unfixed source first.

The stable-key test **passed there on its first draft** — two mints in the same second are byte-identical whatever the code does, so the probe could never have produced the failure it was meant to catch. Rewritten to backdate the generation, which is the case that actually discriminates; it now fails pre-fix with `expected 1793521998 to be 1790640000`.

The service-worker test executes the shipped predicate via `new Function` rather than matching source text, so it cannot pass on a version that never reaches the fetch handler, and asserts the exemption precedes the navigation branch — the ordering *is* the bug.

Full gate green: `type-check`, `lint`, api 1987, web 934, world 19, rules 23.

---

### DIFF
--- a/apps/api/src/creator-agent-key-routes.test.ts
+++ b/apps/api/src/creator-agent-key-routes.test.ts
@@ -1,6 +1,11 @@
 import type { FastifyInstance } from 'fastify';
 import { afterEach, describe, expect, it } from 'vitest';
-import { assertCreatorAgentKeyActive, looksLikeCreatorAgentKey, verifyCreatorAgentKey } from './agent-creator-key.js';
+import {
+  assertCreatorAgentKeyActive,
+  DEFAULT_CREATOR_AGENT_KEY_TTL_DAYS,
+  looksLikeCreatorAgentKey,
+  verifyCreatorAgentKey,
+} from './agent-creator-key.js';
 import { NO_OPEN_ROUND_REASON, PLATFORM_ROUND_REASON, SLUG_NOT_ON_ACCOUNT_REASON } from './agent-game-key.js';
 import { mintGameAgentKey } from './agent-game-key.js';
 import { mintAgentToken } from './agent-token.js';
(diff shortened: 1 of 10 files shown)


gtanczyk:
@codex

PR #491, opened 2026-08-03
byoca: an OAuth agent could join a round but never open one

gtanczyk:
From CP-2's PASS B re-run, which got OAuth working end to end for the first time and then found where the path stops.

## The gap

`start` accepts an OAuth Bearer + slug. `open_round` never learned the same identity — it has branches for a creator key and a per-game key, and an OAuth token falls through to the final `else`:

```json
{"error":"pass a durable per-game key, or Authorization Bearer with a creator key + slug"}
```

An agent holding a valid OAuth token *is* presenting an `Authorization: Bearer`, so the refusal reads as though its credential were malformed rather than out of scope.

The effect is the promise inverted. A creator who connects by signing in gets one round; when it closes, the agent goes idle until a human opens the next one in Studio. *"After one-time configuration, only a slug goes in the prompt"* held only while somebody else kept opening rounds.

## The fix

`open_round` accepts OAuth alongside the two key shapes.

The identity-agnostic half of the creator-key resolver — slug ownership, publish state, active round — is extracted as `resolveOwnedSlugForOpenRound` and shared. Both credentials answer the same question, *which creator is this*, and duplicating the lookups downstream is how two paths drift into refusing different things for the same situation. The refusals and both tool descriptions now name OAuth as accepted.

## Verification

The parity test asserts the unowned-slug refusal is byte-identical to the creator-key one — slug-centred, never blaming the credential — so a mistype cannot push someone toward a destructive rotate on either path.

Both tests fail against the unfixed source:

```
× opens an improvement round over OAuth, the same as a creator key does
  → expected true to be false
× refuses an unowned slug over OAuth with the same reason the creator key gives
  → expected 'pass a durable per-game key, or Autho…' to be 'no game with that slug on your accoun…'
```

Full gate green: `type-check`, `lint`, api 2007, web 949, world 19, rules 23.

## Not in this PR

The addendum's other finding — the consent screen never names the client, and the connected-agents list labels every entry `localhost` from the redirect host rather than the client's own name — is a UI change with a mock already drawn, and it belongs with the consent-CSRF hardening rather than here.

---

### DIFF
--- a/apps/api/src/agent-creator-key-resolve.ts
+++ b/apps/api/src/agent-creator-key-resolve.ts
@@ -128,7 +128,34 @@ export async function resolveCreatorAgentKeyForOpenRound(
   const verified = await verifyDurableCreatorAgentKey(store, key, secret, nowMs);
   if (!verified.ok) return verified;
 
-  if (!(await creatorOwnsSlug(store, slug, verified.claims.creatorUid))) {
+  const owned = await resolveOwnedSlugForOpenRound(store, slug, verified.claims.creatorUid);
+  if (!owned.ok) return owned;
+  return {
+    ok: true,
+    claims: verified.claims,
+    publishedRecord: owned.publishedRecord,
+    activeRound: owned.activeRound,
+    slug: owned.slug,
+  };
+}
+
+/**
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #492, opened 2026-08-03
byoca: tell the agent which field, and that the session id matters

gtanczyk:
Two CP-2 leftovers, both cases of the platform answering an agent with less than it knows.

## `report_progress` rejected without naming anything

The tool declares `text` required and then forwarded whatever arrived. An agent that guessed `phase`/`message` — plausible names, and the ones CP-2 actually tried — got the channel's bare:

```json
{"error":"Required"}
```

That names neither the field that was missing nor the fields that exist. It now names both, and an out-of-enum `step` says which values are legal instead of failing the same anonymous way.

## The `sessionKey` ↔ `Mcp-Session-Id` binding was documented nowhere

A `sessionKey` is refused under a different transport session, which is correct. But `start` said only *"Does not treat Mcp-Session-Id as authority"* — true, and read as *"the header does not matter."* It is **necessary, just not sufficient**, and nothing said so.

The description now says to send the same one, and that a lost transport session is recovered by calling `start()` again. CP-2 discovered that by having it happen overnight and guessing correctly.

## Verification

Both tests fail against the unfixed source. The first fails with the literal string `Required`, which is the entire complaint:

```
× names the fields when report_progress is called with the wrong ones
  → expected 'Required' to match /report_progress needs text/i
× documents that a sessionKey is bound to the session that minted it
  → expected 'Short-lived session capability from s…' to match /same Mcp-Session-Id/i
```

Full gate green: `type-check`, `lint`, api 2007, web 949, world 19, rules 23.

## The third item is not a defect

CP-2 flagged that unauthenticated `initialize` returns 200, so `claude mcp list` reports **Connected** with no credential and never volunteers OAuth. Returning a 401 challenge there would be wrong: **game keys and legacy round keys arrive as the `key` tool argument, not as a header**, so those clients legitimately send no `Authorization` on `initialize`. Challenging it would break both paths to fix a cosmetic signal.

`shouldIssueMcpOAuthChallenge` already encodes this — it exempts the handshake and fires only on `tools/call`. The current behaviour is deliberate and correct; the false "Connected" is a client-side inference from a successful handshake, and there is no protocol-legal way to say "you will need auth later" while still supporting argument-carried credentials.

---

### DIFF
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -1074,4 +1074,68 @@ describe('POST /api/mcp (BY-05)', () => {
     expect(instructions).toMatch(/get_gate_verdict/);
     expect(instructions).toMatch(/honour stop/i);
   });
+
+  // CP-2: an agent that guessed `phase`/`message` got the channel's bare
+  // {"error":"Required"} — which names neither the missing field nor the real ones.
+  // The tool declared `text` required and then forwarded whatever arrived.
+  it('names the fields when report_progress is called with the wrong ones', async () => {
+    const store = new InMemoryStore();
+    await seedActiveSelfJob(store);
+    app = await createApp(store);
+    const sessionId = await initialize(app);
+    const started = await callTool(app, 'start', { key: roundKey() }, { 'mcp-session-id': sessionId });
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #489, opened 2026-08-03
Speed up catalog load and scroll (poster-first videos)

gtanczyk:
## Problem
The home arcade felt slow in two ways:
1. **Catalog paint blocked on sort signals** — and painting unsorted then reshuffling was worse.
2. **Sluggish scrolling** — every near-viewport card attached a `<video src>` and never unloaded it.
3. **`/api/recommendations` itself was slow** — every home load hit Firestore for scorecards + recent publishes with no server cache.

## Changes
- **No provisional re-sort:** wait for sort signals (or viewer-scoped session cache) before first paint; apply network payload only when it differs (or after play refresh).
- **Signals fetch resilience:** transport failure marks signals ready so the grid never sticks on loading.
- **Viewer-scoped cache:** sessionStorage keyed to uid, cleared on logout.
- **Parallel client fetch** of recommendations alongside catalog.
- **Server cache** for shared community signals (~5 min, coalesce, stale-on-failure).
- **Poster-first video** with unload off-screen; prefer non-`opening` still.
- Gallery media: `Accept-Ranges` / `206`, and **If-Range** mismatch → full `200`.

## Test plan
- [x] Unit: recommendations shared cache + coalesce
- [x] Unit: signals reject fallback / viewer-scoped cache / If-Range
- [x] Unit: poster-first + unload
- [x] Green gate
- [ ] Manual: cold load, account switch, hover preview, scroll

<div><a href="https://cursor.com/agents/bc-8fb05ef3-d984-43f7-acdf-5cbf729c728e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8fb05ef3-d984-43f7-acdf-5cbf729c728e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #493, opened 2026-08-03
Take the moment strip and full-size decodes off the scroll path

gtanczyk:
Rebase of #434 onto current master. The original branch lives on a fork this session cannot force-push to, so the rebased result is re-submitted from an in-repo branch; the commit keeps @jactan77's authorship.

What survived the rebase, and what did not:

- **Dropped**: the first commit ("Mount a catalog card's video only when somebody asks to watch it") — master already does this since #489 (`videoArmed` + arm-on-intent), so replaying it had nothing left to add.
- **Kept**: the media size variants. Screenshots are baked at 96px (moment thumbnail) and 640px (card poster) at snapshot publish time, requested with `?w=` from an allowlist of widths, falling back to the original for snapshots baked before variants existed.
- **Kept, re-mapped**: the `engaged` gate on the moment strip, rebuilt on top of master's `armPreview` mechanics instead of the PR's own (superseded) preview state. Hover, focus or the preview toggle reveals the strip; leaving the viewport gives it back. Cards with screenshots but no video are keyboard-reachable and get their strip on focus — the silent-feature-loss case the original PR called out, with its test carried over.

Verified: type-check clean across all workspaces; ArcadeCatalog/App.catalog/snapshot-publish/snapshot-serve/image-variants suites (53 tests) and submissions (144 tests) all green.

Supersedes #434.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -40,11 +40,13 @@
     "genaicode": "^2.1.0",
     "google-auth-library": "^10.9.0",
     "jose": "^6.2.4",
+    "pngjs": "^7.0.0",
     "web-push": "^3.6.7",
     "zod": "^3.23.8"
   },
   "devDependencies": {
     "@types/node": "^20.14.15",
+    "@types/pngjs": "^6.0.5",
     "@types/web-push": "^3.6.4",
     "@types/ws": "^8.18.1",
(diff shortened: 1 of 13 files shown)


gtanczyk:
@codex

PR #495, opened 2026-08-03
byoca: let an agent create a game, not only join one

gtanczyk:
A connected client could join a round on a game that already existed and had no way to make one. A creator opening ChatGPT and saying *"let's build a game"* dead-ends — the agent asks for a slug, there is no slug, and nothing it can call produces one. Studio was the only door.

`create_game` opens that door: creator key or OAuth access in `Authorization: Bearer`, a title and a concept in the creator's words, and it returns the slug that `start` takes.

## Not a second implementation

Studio's `POST /api/submissions` ran nine ordered steps — configured check, beta/blocked, payload validation, per-IP limit, moderation, the global creation circuit-breaker, per-user quota, slug mint and claim, brief persistence, dispatch — each with a documented reason for sitting where it does. Rate limiting is ahead of moderation because moderation is a *paid* call; the circuit-breaker is ahead of the per-user quota so a refusal there costs the creator nothing.

That body is lifted into `createGame` and injected into the MCP server exactly the way `startImprovementRound` already is. Both surfaces run the identical steps in the identical order. A creator reaching creation through their agent spends the same quota against the same limits as one reaching it through Studio — and there is no second copy to drift.

## Who may create

Only creator-wide credentials. A **per-game key** is scoped to a game that already exists and cannot widen into making another; a **sessionKey** is an in-round capability. Both are refused by name rather than by a generic message. With no credential at all the HTTP layer challenges first with a 401 and the OAuth discovery header — BY-18a working as intended, and asserted as such.

`builder` is forced to `self`: an MCP-created game is by definition being built by the caller's agent.

## Two things the refactor had to get right

Both were caught by tests that already existed, which is the system working:

- **503 before auth.** An unconfigured server answered 503 before authenticating; my first cut authenticated first and returned 401. A test pins the order.
- **The moderation category.** `moderation-metrics.test.ts` scans every moderating route for `category: … ?? 'other'` at the send site, because a verdict's category is optional and JSON drops undefined — producing a 422 the client cannot look up. Moving the normalization inside `createGame` put it where that scan couldn't see it. It's back at the send site, with a comment saying why it lives there.

## Verification

All four new tests fail against the unfixed source. The happy-path test doesn't stop at "a game was created" — it feeds the returned slug straight into `start` and asserts the round binds, because a slug `start` won't take is not a fix for this bug.

Gate verified **by exit code** this time — `type-check`, `lint`, `test`, `build` all 0. api 2022, web 949, world 19, rules 23.

---

### DIFF
--- a/apps/api/src/creator-agent-key-routes.test.ts
+++ b/apps/api/src/creator-agent-key-routes.test.ts
@@ -793,6 +793,9 @@ describe('creator agent key routes + MCP start (BY-27a)', () => {
     expect(job?.ownerUid).toBe(OWNER);
     // The caller's own agent builds it — that is what this tool is for.
     expect(job?.builder).toBe('self');
+    // The queued transition carries where it came from, mirroring agent_open_round, so
+    // adoption of the chat-client flow is measurable instead of looking like Studio.
+    expect(job?.transitions?.[0]).toMatchObject({ to: 'queued', by: 'agent', reason: 'agent_create_game' });
     // The concept is persisted as the brief, so get_brief has something to serve.
     expect(job?.spec).toContain('television station');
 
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex

PR #496, opened 2026-08-03
byoca: a tool that only reads should not be advertised as destructive

gtanczyk:
ChatGPT badges `list_examples` **DESTRUCTIVE**, and it is right to.

MCP's annotation defaults are not "unknown" — an un-annotated tool reads as `readOnlyHint: false` and `destructiveHint: true`. So every tool on this surface was claiming it might destroy something. `get_brief`, `get_kit`, `get_sources` and `list_examples` read and nothing else, and even the writers deliver, report or open. **Nothing here deletes anything.**

That badge is worse than cosmetic: it is shown to a creator at the moment they decide whether to let an agent run, and a warning that is wrong on the safe tools teaches people to ignore it on the dangerous ones.

All fourteen tools now carry annotations with a human-readable title, and the readers say so.

## outputSchema

Declared for the tools whose payload **this file constructs** — `start`, `open_round`, `get_sources`, `list_examples`, and the `report_progress` / `send_screenshot` / `ack_inbox` replies that share the `{ ok, stop, pendingMessages }` shape.

Left undeclared, deliberately, for `get_brief`, `get_seed`, `get_kit`, `get_example`, `get_gate_verdict` and `read_inbox`: those return the channel's body verbatim, and declaring a shape this file does not build would assert a contract it cannot keep. A wrong `outputSchema` is worse than a missing one — it makes a model plan against a payload that never arrives. Doing those properly means pinning the channel's own response types first, which is a separate change.

## Verification

Both tests fail against the unfixed source, the first with `start has no annotations: expected undefined to be truthy`. The annotation test asserts across **every** tool rather than a sample, so a tool added later without annotations fails it.

Gate by exit code: `type-check`, `lint`, `test`, `build` all 0. api 2020, web 955.

## Note

`create_game` (#495) is not annotated here — it doesn't exist on master yet. I'll add its annotations to that branch so neither PR blocks the other, and whichever lands second carries the pair.

---

### DIFF
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -1163,7 +1163,18 @@ describe('POST /api/mcp (BY-05)', () => {
     for (const tool of tools) {
       expect(tool.annotations, `${tool.name} has no annotations`).toBeTruthy();
       expect(tool.annotations?.title, `${tool.name} has no title`).toBeTruthy();
-      expect(tool.annotations?.destructiveHint, `${tool.name} claims to be destructive`).toBe(false);
+      expect(typeof tool.annotations?.destructiveHint, `${tool.name} has no destructiveHint`).toBe('boolean');
+    }
+
+    // `destructiveHint: false` is a claim that the tool is purely *additive*, and a
+    // client may skip its approval prompt on that basis — so a tool that consumes a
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #499, opened 2026-08-03
Remix code lane: show the edit call the real GameKit, and type-check what comes back

gtanczyk:
Packages up the code-lane work that was sitting uncommitted in the shared checkout (bench-verified earlier today).

Two halves of the same failure class — properties invented on `GameKit` that esbuild transpiles happily and that break the game at runtime, sometimes silently:

- **The prompt now includes `shared/game-kit.d.ts`** (fetched once, cached per ref via a new `getGameKitDeclaration` on the GitHub client, truncated at 80k chars). Counter-intuitively this made the lane *faster* — mean 14.5s → 7.0s on the bench — because a model that can see what it must satisfy writes the short correct thing once instead of spending repair rounds.
- **A tsc-based gate (`type-check.ts`)** checks every candidate in memory before it becomes a document, mirroring the games repo's compiler options and handing at most 6 diagnostics to the repair loop. Benched over 18 edits it caught all five candidates that compiled and then failed the player — three of which failed without throwing (NaN-radius draws paint nothing).

`typescript` moves from devDependencies to dependencies because the gate now runs in production; the lockfile commit updates its dev flag so `npm ci --omit=dev` keeps it in the image.

Verified: type-check green across workspaces; type-check (7), remix (22) and code-lane (13) suites pass.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -40,6 +40,7 @@
     "genaicode": "^2.1.0",
     "google-auth-library": "^10.9.0",
     "jose": "^6.2.4",
+    "typescript": "^5.5.4",
     "web-push": "^3.6.7",
     "zod": "^3.23.8"
   },
@@ -48,7 +49,6 @@
     "@types/web-push": "^3.6.4",
     "@types/ws": "^8.18.1",
     "tsx": "^4.16.5",
-    "typescript": "^5.5.4",
     "vitest": "^3.2.7",
     "ws": "^8.18.0"
   }
--- a/apps/api/scripts/_gatecheck.ts
+++ b/apps/api/scripts/_gatecheck.ts
@@ -1,0 +1,18 @@
+import { readFile } from 'node:fs/promises';
+import { typeCheckGame } from '../src/type-check.js';
+
+async function main() {
+  const kit = await readFile('/Users/gtanczyk/src/www.gamedev.pl-games/shared/game-kit.d.ts', 'utf8');
+  const cases: Record<string, string> = {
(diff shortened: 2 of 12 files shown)


gtanczyk:
Closing — everything here is already on master, and merging it would take code backwards.

Both halves landed as `29fa3ab` this afternoon, straight to master rather than through this branch. `apps/api/src/type-check.ts` and its test diff to **zero** against master, `getGameKitDeclaration` is in `github-client.ts`, and `typescript` is already in `dependencies` rather than `devDependencies`, so `npm ci --omit=dev` keeps it in the image.

The risk is in the other direction. This branch was cut before that commit, so head-to-head against master it **removes 194 lines and adds 62** across `code-lane.ts` and `github-client.ts` — reverting `7ec90d8` (the repair round carrying the code it is fixing) and the summary fix from #494, which is what stopped players being told "Fixed type error in startGame by removing invalid property reference on RaceScene3D". That is what the merge conflict is pointing at.

The one file not on master is `apps/api/scripts/_gatecheck.ts`, and it is not worth rescuing: it hard-codes `/Users/gtanczyk/src/www.gamedev.pl-games/shared/game-kit.d.ts`, so it runs on exactly one machine. Its four cases are already covered by `type-check.test.ts`, which runs in CI — an unknown kit member, a call to one, a known member as control, and an unknown member on a game's own type, plus two the script never had (no-kit stand-down, and not reporting problems inside the declaration itself). The permanent version of this probe already exists.

Nothing is lost by closing this. The bench numbers it reports are still the reason the work shipped.

---

PR #505, opened 2026-08-03
byoca: make the consent screen a consent screen

gtanczyk:
It rendered, and that was all it did.

> "A coding agent is asking to access your gamedev.pl account."

A sentence every client produces identically, about an account it never named, granting something it never described, on a page that looked nothing like gamedev.pl.

## It could be reached without being seen

`POST /oauth/authorize` mints a durable grant, and its only protection was the session cookie's `sameSite: 'lax'` — which does not cover a top-level cross-site form POST. A page the creator never visited could hand a coding agent standing write access to their games with the consent screen never rendering.

**Every existing test in the suite submitted exactly that shape and passed**, which is the clearest evidence the hole was real — they all had to be changed to go through the form.

Consent is now bound to the screen that issued it by an HMAC over `(uid, client_id, code_challenge)`. Stateless, and covering the uid means a token cannot be lifted from one creator's form onto another's, nor replayed against a different client or PKCE challenge.

## It says who, what, and as whom

- **Who** — from the `clientName` already on the record and previously discarded, falling back to `client_id` so a nameless client shows a URL rather than a generic phrase.
- **As whom** — the account being granted from, which matters most for anyone signed into more than one Google account.
- **What** — rounds, daily quota, sources, publishing. Not "access".
- **How long, and where to undo it.**

The redirect URI is still shown, but reframed from *"you will be sent back to"* into the thing to check.

## It looks like the site

The page set no background at all, so it shipped on the browser default white while gamedev.pl is dark (`--bg: #0f1418`). That is not decoration: a consent screen is exactly where someone decides whether a page really belongs to the site it claims to, and one that looks foreign removes the only cue they have — while training people that unfamiliar-looking gamedev.pl pages are normal.

## Deliberately not included: the waitlisted state

Catching beta access at consent rather than at the agent's first call is right, and I built it. Then a test failed and showed the gate was wrong: `auth.ts` allows on `betaAllowedUids` **OR** `betaAllowedEmails` **OR** `isWaitlistApproved`, and I had checked only the third. It would have refused every creator allowlisted by environment — most of the people using this, very likely including the owner.

Shipping a wrong gate on a consent screen is worse than shipping it later. Doing it properly means extracting that predicate so both call sites share it rather than copying a third of it, and that is its own change.

## Verification

Full gate by exit code: `type-check`, `lint`, `test`, `build` all 0.

A/B against the unfixed source shows the six pre-existing OAuth tests failing on the CSRF check — the bypass they were unknowingly exercising.

---

### DIFF
--- a/apps/api/src/creator-agent-key-routes.test.ts
+++ b/apps/api/src/creator-agent-key-routes.test.ts
@@ -13,6 +13,7 @@ import { mintSessionToken, SESSION_COOKIE_NAME } from './auth.js';
 import { buildApp } from './app.js';
 import type { CatalogGameEntry, GameSources, GitHubClient, LinkedPullRequest } from './github-client.js';
 import { mintMcpSessionKey } from './mcp-session-key.js';
+import { consentToken } from './oauth-as.js';
 import { pkceChallengeS256 } from './oauth-pkce.js';
 import { InMemoryStore } from './store.js';
 import type { CreatorAgentKeyRecord } from './store.js';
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #510, opened 2026-08-03
feat: browse Creator Kit files over MCP (no full-kit dump)

gtanczyk:
## Problem

ChatGPT Apps can call MCP successfully, but `get_kit` still points agents at a GCS signed URL (`curl | tar`). Code-execution sandboxes often cannot resolve `storage.googleapis.com`, and returning the whole ~1.4 MB kit over MCP would blow up agent context.

## Solution

Keep `get_kit` as **metadata** (`engineRef`, `sha256`, `entry`, optional `kitUrl`/`unpack`, plus a `browse` map of tool names). Add four read tools that serve the pinned kit from the server-side tarball:

| Tool | Purpose |
| --- | --- |
| `list_kit_files` | Paths + size + text/binary; `prefix` / simple `glob`, paginated |
| `search_kit_files` | Case-insensitive substring search over text files (capped hits) |
| `read_kit_file` | Whole file ≤48 KiB (binaries need `encoding=base64`) |
| `read_kit_file_fragment` | Line/byte windows for larger files (`nextOffset`; bytes → base64) |

Channel routes (same auth as other build reads):

- `GET /api/agent/build/kit/files`
- `GET /api/agent/build/kit/search`
- `GET /api/agent/build/kit/file`
- `GET /api/agent/build/kit/file/fragment`

Pass `engineRef` from `get_kit` on every browse/read call (N/N−1 pin). Caps: list ≤500, search ≤40 hits, whole-file ≤48 KiB, fragments ≤32 KiB / 200 lines.

## Review follow-ups

Addressed Copilot + Codex feedback on this branch:

- Escape `?` in simple globs; treat non-finite `limit`/`offset` as absent
- Pin browse to `engineRef` so a mid-round registry bump cannot mix kits
- Reject overlong line windows (use `unit=bytes`) instead of truncating
- Require `encoding=base64` for byte fragments (no UTF-8 split)

## Tests

- Unit coverage in `kit-files.test.ts`
- Channel coverage in `agent-build-reads.test.ts`
- MCP tool listing / workflow / readOnly annotations in `mcp-server.test.ts`

[...]

PR #506, opened 2026-08-03
Add GET /api/agent/build/media endpoint for gate-produced media (BY-28)

gtanczyk:
## Summary
Implements a new endpoint for agents to retrieve gate-produced media (screenshots and gameplay video) from completed deliveries. This addresses the use case where agents cannot run games themselves and need visual evidence to iterate on submissions and show creators.

## Key Changes

- **New HTTP endpoint** `GET /api/agent/build/media`: Returns signed URLs for gate-produced capture PNGs and gameplay MP4, plus the opening frame embedded as base64 PNG data
  - Requires valid agent build token (same auth as other channel reads)
  - Supports optional `version` query parameter to fetch specific delivery versions
  - Returns `available: false` gracefully before any delivery or when gate stored no media
  - Enforces filename allowlist from validated `media/metadata.json` (same rule as published-media route)
  - Falls back to manifest's `gate.screenshot` for runs that failed mid-capture
  - Implements terminal receipt access: readable after green closes the round, but only for the delivery the token owns

- **New MCP tool** `get_gate_media`: Wraps the HTTP endpoint for Claude/ChatGPT clients
  - Returns structured content with signed URLs and inline opening frame as image block
  - Marked as read-only (non-destructive)
  - Documented in session workflow to guide agents on when/how to use it

- **Security & validation**:
  - Version IDs shape-checked to prevent path traversal
  - Manifest lookup proves ownership (version must exist under the job's slug)
  - Signed URLs scoped to `games/{slug}/versions/{version}/media/` paths
  - Terminal receipt enforces single-delivery access (retired keys cannot read other versions)
  - Returns 503 if media store (GamesStore/GcsObjectStore) is unconfigured

- **Test coverage**: Comprehensive test suite covering auth, allowlisting, fallbacks, terminal receipt, version validation, and store configuration scenarios

## Implementation Details

- Reuses `parseGameMedia()` from github-client to parse metadata structure
- Inline frame limited to reasonable size (maxShotBytes) to avoid bloating responses
- Opening shot selection: prefers 'opening' name, falls back to first screenshot
- Respects existing `DEFAULT_SIGNED_URL_TTL_SECONDS` for URL expiry
- Integrates with existing `resolveBuild()` auth flow with `allowTerminalReceipt: true` flag

https://claude.ai/code/session_01PrXBSADXMndFNkj9FfEkMn

PR #516, opened 2026-08-03
byoca: make every agent artifact readable without egress (BY-28a)

gtanczyk:
## Why

Owner test, 2026-08-03: **a ChatGPT-side connector agent cannot reach external hosts at all.** It can call our MCP tools and nothing else — no shell, no `curl`, no fetch.

That falsifies an assumption running through the whole WS3 tool contract, which says "signed URL" in five places and never asks whether the reader can follow one. For a fetchless client a URL is not a degraded experience — it is a blank one. Three tools were affected:

| Artifact | Before | Now |
| --- | --- | --- |
| Creator Kit | tarball URL only | `list_kit_files` / `read_kit_file` / `search_kit_files` (**#510**, already merged) |
| Gate media | one frame inline, rest by URL | frames as MCP image blocks (**this PR**) |
| Exemplar games | tarball URL only | `list_example_files` / `read_example_file` (**this PR**) |

## What changed

**Gate media — carry the frames, don't point at them**
- `frames` argument: `opening` (default, one), `all` (up to 3), `none` (URLs only).
- Budget ≤3 frames / ≤1.4 MB decoded; anything dropped is reported as `framesOmitted` rather than silently truncated.
- URLs stay for everyone — shell-capable agents want them, and the fetchless agent can still hand one to the *creator*, who has a browser.
- The mp4 stays URL-only (nothing renders an inlined video) with a `videoNote` saying so, so an agent that cannot fetch doesn't report it broken.

**Exemplars — the last link-only tool**
- `list_example_files` / `read_example_file` serve the same `examples/<slug>.tgz` the publisher already writes, through the tool.
- Narrower than the kit browser on purpose: a kit is a toolchain needing search and fragments; an exemplar is one game directory, so list + read covers it.
- Same allowlist, now enforced on three verbs instead of one — a slug sitting in the bucket doesn't become readable because a different route reaches the same object. Members outside `games/<slug>/` are dropped at unpack; `..` is refused rather than normalized; oversized files are refused rather than truncated.

## On context cost

Deliberately bounded, since inlined bytes *are* context:
- capture frames are 900×900 (`deviceScaleFactor: 1`) ≈ **1k image tokens each** — cheap for a few, not for all eight by default;
- base64 rides a typed `image` content block and **never the JSON body** — as text the same bytes cost ~100× more. A test asserts the base64 does not appear in the text content so nobody merges it back in later.

## Verification

`type-check`, `lint`, `build -w @gamedevpl/api` (the production build that caught the last merge-tree break) and the full API suite: **2096 tests green**. New coverage for frame modes and omission reporting, exemplar path normalization/traversal, allowlist parity across the three example verbs, root-escape at unpack, and token requirements.

Branch restarted from `master` after #506 merged, rather than stacked on merged history.

Docs: [ops#58](https://github.com/gamedevpl/www.gamedev.pl-ops/pull/58) records the finding as **BY-28a** plus a standing rule — *a URL may be an optimisation, never the only path to content an agent must read* — so a future tool gets checked against it rather than rediscovering this one at a time.

### DIFF
--- a/apps/api/src/agent-build-media.test.ts
+++ b/apps/api/src/agent-build-media.test.ts
@@ -190,8 +190,66 @@ describe('GET /api/agent/build/media (BY-28)', () => {
       expect(name).toMatch(new RegExp(`^games/${SLUG}/versions/${VERSION}/media/`));
     }
     // The inline frame is the opening shot, so a client that cannot fetch a URL still
-    // sees the game.
-    expect(body.openingShot).toEqual({ file: 'opening.png', png: PNG.toString('base64') });
+    // sees the game. Default is one frame, not all of them.
+    expect(body.frames).toEqual([{ file: 'opening.png', name: 'opening', png: PNG.toString('base64') }]);
+    // The video is URL-only, and the payload says so — the agent that needs to know is
+    // the one that cannot test a URL to find out.
+    expect(body.videoNote).toMatch(/only as a URL/i);
+  });
+
+  it('attaches every frame on frames=all, one on the default, none on frames=none', async () => {
+    // A ChatGPT-side connector can call our tools and nothing else — no shell, no fetch
+    // (owner test, 2026-08-03). For that client a signed URL is a blank experience, so
+    // the bytes have to ride the reply. Bounded, because each frame costs context.
+    const store = new InMemoryStore();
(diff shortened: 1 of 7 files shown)


gtanczyk:
@codex

PR #521, opened 2026-08-03
fix(mcp): stop refusing ChatGPT when Mcp-Session-Id drifts

gtanczyk:
## Summary

ChatGPT Apps could `create_game` / `start` but then every later tool failed with `sessionKey is bound to a different Mcp-Session-Id`. That is **not** a multi-game limit.

### Cause

1. `sessionKey` was refused when the `Mcp-Session-Id` header differed from the id embedded at mint time. The id is already inside the signed key, so the check added no theft resistance.
2. Unknown correlators 404’d when missing from this instance’s in-memory map — Cloud Run is multi-instance (`maxInstanceCount` > 1) and ChatGPT does not pin to one revision.
3. `start()` reminted a new id when the client’s header was absent from the local map, then the client kept sending the old one → guaranteed mismatch.

### Fix

- Accept a cryptographically valid `sessionKey` even if the header drifted (log `mcp_session_id_drift`).
- Adopt well-formed unknown correlators instead of 404 (multi-instance).
- Bind `start()` to the client’s presented id when well-formed, without requiring a prior local `initialize`.
- After Codex P2: `DELETE /api/mcp` tombstones the correlator on this instance so concurrent/retry POSTs get 404 instead of re-adopting the terminated id (foreign never-seen ids still adopt).

### Reproduce / verify

- ChatGPT: open `/studio/legions-real-time-battles`, `start({ slug })` then `get_brief` — should work after deploy.
- Multiple games in one chat are fine: each `start` returns a per-job `sessionKey`; use the matching key per game.

### Test plan

- [x] Accept sessionKey under drifted `Mcp-Session-Id`
- [x] Bind start to foreign well-formed correlator (no local initialize)
- [x] DELETE tombstones correlator (404 on reuse); foreign never-seen id still adopts
- [x] MCP server + session-key unit tests green

<div><a href="https://cursor.com/agents/bc-f848e546-4d56-4944-b20b-8b6460ebdeea?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f848e546-4d56-4944-b20b-8b6460ebdeea&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -595,7 +595,9 @@ describe('POST /api/mcp (BY-05)', () => {
     expect(JSON.stringify(res.structured)).toContain('finished');
   });
 
-  it('rejects a sessionKey presented under a mismatched Mcp-Session-Id', async () => {
+  it('accepts a valid sessionKey even when Mcp-Session-Id has drifted', async () => {
+    // ChatGPT Apps (and Cloud Run multi-instance) often present a different correlator
+    // than the one start() embedded. The sessionKey is the capability; the header is not.
     const store = new InMemoryStore();
     await seedJob(store);
     app = await createApp(store);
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #540, opened 2026-08-04
Automate account deletion sweep setup

gtanczyk:
## Summary

- derive the account-deletion sweep audience and caller service account in both deployment paths
- add an idempotent setup helper that creates or updates the OIDC Scheduler job and ensures its caller identity
- invoke that helper from the normal GCP bootstrap
- replace the manual rollout commands with the automated setup path
- add source-contract tests that prevent the deployment wiring from drifting back to a required GitHub variable

## Why

The deletion grace-window PR deployed the endpoint, but provisioning still depended on two manual commands: creating a GitHub Actions variable and creating the Scheduler job. A fresh environment could therefore accept deletion requests without ever running final cleanup.

This change makes those values deterministic and brings the Scheduler resource under the repository's existing idempotent shell-script infrastructure.

## Verification

- `npm install`
- `bash -n infra/setup-account-deletion.sh infra/setup-gcp.sh infra/deploy-api.sh`
- `npm run type-check`
- `npm run lint`
- `npm run test` (full suite; 3 optional isolated-vm cases skipped)
- `npm run build`


gtanczyk:
@codex

PR #542, opened 2026-08-04
fix(studio): open an improvement round's thread with the request that started it

gtanczyk:
Follow-up to #539, which fixed the draft-feedback half of this. Same underlying question — *who wrote the words on the creator's side of the thread* — asked at the post-publish boundary.

## The bug

Publishing is terminal, so an improvement is a **new job with an empty thread**. The request that started it lived only in the brief the agent reads (`setSubmissionBrief`); nothing ever called `appendCreatorMessage`.

So a creator sends a change request from Studio, lands on the new round, and finds no trace of what they just asked for. The page's own optimistic echo (`pendingRevisions`) covers it until the first reload, which clears on token change — and the improve response deliberately hands back a *new* token, so the handoff itself drops it.

## The fix

`startImprovementRound` seeds the new job's thread, but only when a person or their agent actually put the request into words. That is the new `requestedBy`:

| Caller | `requestedBy` | Result |
|---|---|---|
| Studio improve composer (`submissions.ts`) | `creator` | Opens with the creator's own words — no label, no translation |
| MCP `open_round` (`mcp-server.ts`) | `agent` | Labelled "Summarized by your agent" and translated, exactly as #539 treats `continue_draft` |
| Approved player-evidence suggestion (`suggestion-inbox.ts`) | *omitted* | Stays out of the thread |
| Autonomous sweep (`suggestion-sweep.ts`) | *omitted* | Stays out of the thread |

The `open_round` row is worth calling out: that was the same defect as #539 in its post-publish form. The agent's relay was previously *invisible* rather than mislabelled, so it never showed up in the original report — but it's the same class, and it now gets the same labelled-and-translated treatment.

The suggestion paths pass nothing on purpose. `buildImprovementBrief` assembles those out of player evidence — nobody typed them, and putting one on the creator's side of the conversation would be the bug #539 fixed wearing a different hat. There's a test pinning that so a future caller doesn't quietly opt in.

## Why `delivered`

`appendCreatorMessage` feeds two things: the Studio thread and the agent's pending inbox (`listPendingCreatorMessages`, which rides every mutating channel reply as `pendingMessages`). For an improvement round the brief already carries the same text to the agent, so queueing it as well would hand over **one instruction that reads like two** — the agent sees a pending note and takes it for something new to act on, mid-round.

So seeded messages are written already collected. They belong in the thread (it is what the creator asked for) without also being an inbox item.

## Tests

- `submissions.test.ts` — the improve round's thread opens with the creator's words, `origin` absent, instrumentation block stripped, and nothing left pending.
- `mcp-open-round.test.ts` — `open_round` seeds one message marked `origin: 'agent'`, also nothing pending.
- `suggestion-inbox.test.ts` — an approved suggestion dispatches with no `requestedBy`.

## Verification

`npm run type-check && npm run lint && npm run test && npm run build` — all green (1172 tests across the workspaces).

---
_Generated by [Claude Code](https://claude.ai/code/session_01JjkGreoXoZWCENEWVh2nFm)_

### DIFF
--- a/apps/api/src/mcp-open-round.test.ts
+++ b/apps/api/src/mcp-open-round.test.ts
@@ -335,6 +335,17 @@ describe('MCP open_round (BY-24 / BY-27b)', () => {
     expect(job?.spec).not.toBe('');
     // The change request is free text, not a clarifications block.
     expect(job?.qa).toEqual([]);
+
+    // It also opens the round's thread — as the agent's relay, not the creator's words,
+    // so Studio labels and translates it the way continue_draft's relay is.
+    const messages = await store.listCreatorMessages(jobId);
+    expect(messages).toHaveLength(1);
+    expect(messages[0]).toMatchObject({
+      text: 'Make the title screen background a solid coloured rectangle.',
+      origin: 'agent',
+    });
+    // Already delivered: the brief above is how the agent receives it.
+    expect(await store.listPendingCreatorMessages(jobId)).toEqual([]);
   });
 
   it('admits only one concurrent open_round per slug', async () => {
--- a/apps/api/src/mcp-server.ts
+++ b/apps/api/src/mcp-server.ts
@@ -138,6 +138,8 @@ export interface McpServerOptions {
     log: { error: (context: object, message: string) => void };
     builder?: BuilderKind;
     openedBy?: 'creator' | 'agent';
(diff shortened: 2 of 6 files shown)


gtanczyk:
@codex review

---

PR #550, opened 2026-08-04
raise the serve cap to 534_800, catching up with games-repo main

gtanczyk:
Brings this side up to the ceiling the games repo already builds against. **Rewritten** from the 533_800 version — see below, the reason for the change moved while it was open.

## What changed under it

`contract:games-repo` reads the **live** games-repo main, and it failed:

```
MAX_BUNDLE_BYTES mismatch: games-repo=534800 website MAX_PROJECT_BYTES=533800
Website is behind the build ceiling — raise GAMEKIT_PLATFORM_BYTES / MAX_PROJECT_BYTES (website-first).
```

While this PR was open, games-repo main raised its own half to 330_000 for `carjack-city`'s world-services expansion ([games-repo #477](https://github.com/gamedevpl/www.gamedev.pl-games/pull/477)). So the drift is main's, and 329_000 was aiming at a number that had already moved.

**This is therefore no longer a website-first raise for [games-repo #480](https://github.com/gamedevpl/www.gamedev.pl-games/pull/480).** That PR needs no ceiling change at all any more: on the new base `carjack-city` assembles to **528_198** against 534_800, with **6_602 bytes** to spare. #480 now carries only its own trim, not a ceiling move.

## The number

313_000 → **330_000**, cap **534_800**, matching main exactly. Catching up rather than leading, which is the direction the check tolerates — it fails only when the games repo builds above what this side would serve.

## What moves with it

- `GAMEKIT_PLATFORM_BYTES` and its doc comment, rewritten to record that this followed #477 rather than #480.
- `fixtures/games-repo/shared/assemble-contract.json`, kept in lockstep.
- The pinned literals in `assemble.test.ts` and `games-repo-contract.test.ts`.
- **Both** synthetic extractor sources — including the `a + b` allowance one, whose parts are re-summed (`13_061 + 316_939`) rather than the assertion being loosened, so it still tests that the extractor evaluates expressions instead of testing a number I just changed.

## Verification

`apps/api`: **147 suites, 2233 tests, all passing** (needs `npm run build:packages` first — without it 47 suites fail to resolve `@gamedevpl/game-generator`, unrelated).

https://claude.ai/code/session_01Jd2LMxMfsvR1XnaXjHfZRz

### DIFF
--- a/apps/api/fixtures/games-repo/shared/assemble-contract.json
+++ b/apps/api/fixtures/games-repo/shared/assemble-contract.json
@@ -27,8 +27,8 @@
     "editor"
   ],
   "authorBudgetBytes": 204800,
-  "platformCeilingBytes": 329000,
-  "maxProjectBytes": 533800,
+  "platformCeilingBytes": 330000,
+  "maxProjectBytes": 534800,
   "audio": {
     "musicField": "string",
     "musicCatalogRequiredKeys": ["version", "tracks"],
--- a/apps/api/fixtures/games-repo/shared/assemble-contract.json
+++ b/apps/api/fixtures/games-repo/shared/assemble-contract.json
@@ -27,8 +27,8 @@
     "editor"
   ],
   "authorBudgetBytes": 204800,
-  "platformCeilingBytes": 313000,
-  "maxProjectBytes": 517800,
+  "platformCeilingBytes": 329000,
+  "maxProjectBytes": 533800,
   "audio": {
     "musicField": "string",
     "musicCatalogRequiredKeys": ["version", "tracks"],
--- a/apps/api/src/assemble.test.ts
+++ b/apps/api/src/assemble.test.ts
@@ -20,7 +20,7 @@ describe('the size cap', () => {
     // Sourced from games-repo-contract.ts; CI re-checks the live games-repo
     // MAX_BUNDLE_BYTES when GAMES_REPO_TOKEN is set (issue #247). Website-first
     // budget raises update this number before games-repo main catches up.
-    expect(MAX_PROJECT_BYTES).toBe(533_800);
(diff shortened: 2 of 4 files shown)


gtanczyk:
Closing — superseded by #553, which landed the identical change while this was open.

Master now has `GAMEKIT_PLATFORM_BYTES = 330_000` / cap 534_800, and #553 covers everything this branch did, including the part I would have checked for: the synthetic `a + b` extractor source is **re-summed** to `13_061 + 316_939` rather than the assertion being loosened, so it still tests that the extractor evaluates expressions. Nothing here is lost by closing.

For the record, since this PR changed reason twice as the ground moved:

1. Opened to raise 313_000 → 329_000 **website-first**, so [games-repo #480](https://github.com/gamedevpl/www.gamedev.pl-games/pull/480) could raise its half — `carjack-city` was 29 bytes under the cap and #480's `input.bridged()` tipped it.
2. Rewritten to 330_000 after `contract:games-repo` failed here: it reads the live games-repo main, which had already gone to 330_000 for carjack's own expansion (games-repo #477). So the drift was main's, not #480's, and I had been aiming at a number that had moved.
3. Now redundant: #553 did exactly that catch-up.

**#480 needs no ceiling change at all.** On the current base `carjack-city` assembles to 528_680 against 534_800, with ~6 KB to spare, so that PR carries only its own work.

One thing worth leaving behind rather than losing with this branch: before #477 and #553, the margin was **29 bytes**. At that width the next author to touch anything under `shared/` pays for it whoever they are and whatever they are doing — which is exactly what happened to #480. The number worth watching is the headroom, not the per-feature deltas. That note is recorded in the games repo's `docs/platform-byte-ledger.md`.

---

PR #559, opened 2026-08-04
Proposals: contribute a change to a game somebody else owns

gtanczyk:
Builds the "pull request for games" feature. The product word is **proposal** — `PR` appears nowhere in the API, the schema, or the UI.

Design and phasing live in the ops repo (`game-pull-request-research.md`, `game-pull-request-execution-plan.md`, ops PR #70).

## What a proposal is

An immutable candidate version written into the target game's own prefix and marked `deliveryMode: 'proposal'`, plus a record of who sent it, what it is based on, and where the review stands. Almost nothing here is new machinery: the gate that checks a proposal is the gate that checks a creator's own delivery, acceptance is the existing improvement-round shape pointed at a version that is already green, and publication is untouched.

## The three invariants, and where each is enforced

**A proposal never publishes.** `isPublishableMode` reads `deliveryMode` off the manifest, and the operator publish route refuses a proposal version even though it has never heard of the proposal registry. `adoptProposalVersion` — the only door out of that mode — requires an accepting owner *and* a green gate. There is an explicit refusal test rather than an absence of call sites.

**A proposal creates no job.** This one changed the design. `creatorOwnsSlug` reads "newest non-abandoned submission for this slug" as ownership, so a proposal job owned by the proposer against the target's slug would have *transferred the game to whoever proposed to it*. Proposals carry `issueNumber: 0` (`PROPOSAL_NO_JOB`) and keep provenance in `manifest.proposal`. Found by writing the test.

**A creator is never surprised by one.** Contributions are off per game by default, and a proposal that has not passed the gate is invisible to its reviewer — so reaching a creator at all costs a change that compiles, runs, and passes the same gate their own deliveries pass.

## Surfaces

Remix gains its third exit. Save-as-yours and share are unchanged and still never publish; **Propose** asks the owner, who may say no. It appears only once a change has landed and only for a game whose owner opted in. Store-lane only for now, and the 409 says so plainly — a repo-era game's code lives on the ref and the session holds only its declaration.

Beyond that: a `/proposals` tracker (state chips, reviewer text quoted as data, withdraw, rebuild-on-latest guidance), one review card serving both the creator's Studio and the ops console, the contributions switch with its block list, and an `/admin/proposals` section. en + pl throughout.

## Notable

Two of the repo's own guards caught real gaps: `moderation-metrics.test.ts` required one rejection report per branch, which moved moderation logging into the domain layer where the field and uid are known; and a Studio test caught a list-shape assumption in the new API client that would have thrown on `.length`.

Whole monorepo green — 3,484 tests, lint and type-check clean.

## Not in this PR

Diff service, notifications, the inline-diff review detail, byline credit, and all of the catalog repo-lane merge-back (archive-backed base, repo gate, apply-bot bridge, MCP proposal rounds). The loop is complete end to end for store-lane games; repo-lane proposals can be opened and reviewed but not yet merged back.

Shipping to real users still waits on the policy work tracked in the ops repo — owner-of-record taxonomy, the contributions decision, B1 question 8, the age-rating questionnaire, and user-level block. Nothing here front-runs those; the build sits behind the existing closed beta.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D58Gbds1eQNoi2PBgqahVk)_

### DIFF
--- a/apps/api/scripts/run-gate.ts
+++ b/apps/api/scripts/run-gate.ts
@@ -95,8 +95,12 @@ async function main(): Promise<void> {
   const harnesses: string[] = [];
   const health = process.argv.includes('--health');
   const preview = process.argv.includes('--preview');
-  if (health && preview) {
-    console.error('--health and --preview are mutually exclusive');
+  // The proposal lane: the full acceptance check, except that a behavioural-golden
+  // mismatch is recorded as a finding rather than refusing the candidate. See
+  // GateRunOptions.proposal for why a proposal needs that and a delivery does not.
+  const proposal = process.argv.includes('--proposal');
+  if ([health, preview, proposal].filter(Boolean).length > 1) {
+    console.error('--health, --preview and --proposal are mutually exclusive');
     process.exit(2);
   }
 
@@ -132,7 +136,7 @@ async function main(): Promise<void> {
     },
     // Health asks about today's engine, so the manifest's pin is exactly the thing to
     // ignore. An acceptance run passes nothing and lets the pin (or `main`) decide.
-    health ? { engineRef: 'main' } : preview ? { preview: true } : {},
+    health ? { engineRef: 'main' } : preview ? { preview: true } : proposal ? { proposal: true } : {},
   );
 
   if (health) {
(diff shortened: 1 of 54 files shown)


gtanczyk:
@codex

PR #560, opened 2026-08-04
Embed every music track a game can reach, not just the autoplay one

gtanczyk:
Paired half of **gamedevpl/www.gamedev.pl-games#502** (rule 6a), which adds an optional `audio.musicTracks` array to `GAME.json` so a game can change score mid-round — ambient exploration into a firefight — without a fetch, which the games repo forbids outright.

## Why this half is required
Serve-time assembly narrowed the injected catalog to `{ [manifest.music]: track }`. Without this change, a published game calling `playMusic('combat-theme')` finds nothing in `__GAME_MUSIC_TRACKS__` and the switch **silently no-ops** — the worst kind of failure, since nothing errors and the game just quietly never changes its score.

`github-client` now parses the optional list and injects the default plus every extra. `__GAME_AUDIO_MUSIC__` still names only the track that autoplays, and the whole catalog is still never shipped.

Parsing mirrors games-repo validate Check 3 — non-empty when present, kebab-case, no duplicates, never a repeat of `audio.music` — so a manifest that would fail the games-repo gate fails here too, rather than serving something subtly different from what CI blessed.

## Changes
- `fixtures/games-repo/shared/assemble-contract.json` refreshed with `audio.musicTracksField`, keeping the vendored copy in lockstep.
- `MUSIC_CONTRACT` documents the new field alongside the existing one.
- The lockstep test now pins **both** audio field shapes, not just the module list and byte ceilings — the audio injection shape is half of what that fixture exists to pin, and it was previously unasserted.
- Two new `github-client` tests: extras are embedded (and the rest of the catalog still is not), and a `musicTracks` repeating `audio.music` is rejected.

## Merge ordering
Contract version stays `2` and the key is additive, so either repo may merge first. Until this lands, a game using `musicTracks` serves with its default track and the swap is a no-op rather than an error.

## Verification
- `npm run lint` — clean
- `npx vitest run apps/api` — 2246 tests, 147 files, all passing

`npm run contract:games-repo` skipped its live cross-repo fetch (`GAMES_REPO_TOKEN` unset in this environment); the unit tests guard the website half.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jk4ff5V1gTbAfxcpfGVUQy)_

### DIFF
--- a/apps/api/fixtures/games-repo/shared/assemble-contract.json
+++ b/apps/api/fixtures/games-repo/shared/assemble-contract.json
@@ -31,6 +31,7 @@
   "maxProjectBytes": 534800,
   "audio": {
     "musicField": "string",
+    "musicTracksField": "string[]",
     "musicCatalogRequiredKeys": ["version", "tracks"],
     "injectedGlobals": {
       "assets": "__GAME_AUDIO_ASSETS__",
--- a/apps/api/src/games-repo-contract.test.ts
+++ b/apps/api/src/games-repo-contract.test.ts
@@ -72,8 +72,14 @@ describe('games-repo-contract (website half)', () => {
       gameKitModules: string[];
       authorBudgetBytes: number;
       platformCeilingBytes: number;
+      audio: { musicField: string; musicTracksField: string };
     };
     expect(fixture.version).toBe(2);
+    // The audio injection shape is half of what this fixture exists to pin: `music` is the
+    // autoplay track and `musicTracks` the extras a game switches to. Serve-time assembly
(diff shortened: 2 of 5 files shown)


gtanczyk:
@codex

gtanczyk:
@codex review

Paired half of gamedevpl/www.gamedev.pl-games#502. Worth focusing on the `parseGameManifest` validation for `audio.musicTracks` — it is meant to mirror games-repo validate Check 3 exactly (non-empty when present, kebab-case, no duplicates, never a repeat of `audio.music`), so any divergence between the two halves is the bug worth catching here.

---

PR #562, opened 2026-08-04
Match a music track by own property, not by truthiness

gtanczyk:
Follow-up to #560. Codex raised this on the paired games-repo PR (gamedevpl/www.gamedev.pl-games#502) and **the same hole is here, already merged** — so this fixes it rather than leaving it to be discovered in production.

## The bug
`parseMusicTracks` returns a `JSON.parse` result, which inherits `Object.prototype`. So `tracks.constructor` is a **function**, not `undefined`, and sails through:

```ts
const track = tracks[name];
if (track === undefined) throw new Error(...)   // never fires for an inherited name
```

`constructor` is also lowercase, so it clears the `isKebabCaseName` filter above it. A game declaring `"musicTracks": ["constructor"]` was therefore accepted — and then embedded **nothing**, because `JSON.stringify` drops a function value. `playMusic('constructor')` would fail at runtime on a build every gate had called clean.

`__proto__` is worse: `selected['__proto__'] = …` sets a prototype instead of adding a key, so `Object.keys()` comes back empty and the whole tracks map serializes as `{}` — a game with no music at all, and no error anywhere.

## The fix
Membership is now `Object.hasOwn`, and the accumulator is `Object.create(null)` so a hostile name cannot reach a prototype in the first place. This mirrors the games-repo fix in `aac90d9`, which moved **all six** of its catalog checks off `in` — the four covering `audio.music` and `audio.sounds` had the identical hole and predated the feature.

## Verification
The new test **fails on master and passes with the fix** — confirmed by reverting only the source change while keeping the test:

```
× getGameSources > rejects an inherited Object.prototype name as a track
  Tests  1 failed | 50 passed (51)     ← fix reverted
  Tests  51 passed (51)                ← fix restored
```

`npm run type-check` clean, `npm run lint` clean, `npx vitest run apps/api` 2247 tests passing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jk4ff5V1gTbAfxcpfGVUQy)_

### DIFF
--- a/apps/api/src/github-client.test.ts
+++ b/apps/api/src/github-client.test.ts
@@ -671,6 +671,42 @@ describe('getGameSources', () => {
     expect(sources?.gameJs).not.toContain('unused-theme');
   });
 
+  it('rejects an inherited Object.prototype name as a track', async () => {
+    // The catalog comes from JSON.parse, so `tracks.constructor` is a function rather than
+    // undefined and survives a truthiness check — and `constructor` is lowercase, so it
+    // clears the kebab-case filter too. Accepting it would embed nothing (JSON.stringify
+    // drops a function) and playMusic would fail at runtime on a build CI called clean.
+    const files = new Map<string, string | Uint8Array>([
+      ['games/raid/index.html', '<canvas id="game"></canvas>'],
+      ['games/raid/game.ts', 'const game: { update(): void } = { update() {} }; GameKit.mount(game);'],
+      ['games/raid/style.css', '.game { color: teal; }'],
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex review

You found this class of bug on the paired games-repo PR (gamedevpl/www.gamedev.pl-games#502) and it was already merged here in #560, so this is the fix for the website half.

Worth checking specifically: whether `Object.hasOwn` plus an `Object.create(null)` accumulator closes every path, and whether the same `!== undefined` / truthiness pattern against a `JSON.parse` result exists anywhere else in the serve path — the sounds and manifest reads nearby are the obvious neighbours.

---

PR #573, opened 2026-08-04
Add MCP Apps views (SEP-1865) Phase 0 spike

gtanczyk:
Implements Phase 0 of MCP Apps views support (SEP-1865), enabling MCP clients to render UI cards for tool results. This is a controlled spike that maintains backward compatibility—clients that don't declare the extension see the exact same contract as before.

## Summary

Adds a new `mcp-ui.ts` module that implements the MCP Apps extension (`io.modelcontextprotocol/ui`), allowing the server to serve HTML views to capable clients. The implementation includes:

- A self-contained round-status card that displays gate verdicts from `get_gate_verdict` tool results
- Capability negotiation during `initialize` to detect and respond to UI-capable clients
- A `ui://` resource registry with `resources/list` and `resources/read` support
- Tool metadata (`_meta.ui`) attachment for tools that have associated views

## Key Changes

- **mcp-ui.ts** (new): Core module containing:
  - `MCP_UI_EXTENSION` and `MCP_UI_MIME_TYPE` constants for SEP-1865
  - `clientDeclaresUi()` to detect UI-capable clients from initialize params
  - `mcpUiEnabled()` flag check (off by default, requires `MCP_UI=true`)
  - `ROUND_STATUS_HTML`: self-contained HTML card with embedded JavaScript for lifecycle management
  - Resource descriptor and read functions for the ui:// registry

- **mcp-server.ts**: Integration with MCP server:
  - Track `uiCapable` per transport session to remember client capabilities across `start` and `open_round` calls
  - Emit `resources` capability and `extensions` in initialize response only for UI-capable clients
  - Add `_meta.ui` to tool descriptors only when client negotiated the extension
  - Implement `resources/list` and `resources/read` handlers (gated by flag and capability)
  - Preserve UI capability when re-setting transport sessions mid-round

- **mcp-ui.test.ts** (new): Comprehensive unit tests for:
  - Flag parsing and client capability detection
  - Resource descriptor and content serving
  - Self-contained HTML validation (no external resources)
  - View lifecycle protocol compliance

- **mcp-server.test.ts**: Integration tests covering:
  - Non-declaring clients remain on pre-views contract
  - UI-capable clients receive extension and resource metadata
  - Resource serving via `resources/list` and `resources/read`
  - Flag-off behavior (no views offered even to declaring clients)
  - Capability persistence across `start` (which re-registers the session)
  - Foreign correlators (multi-instance) treated as not view-capable

- **.env.example**: Documentation of `MCP_UI` flag

- **Backward compatibility**: Clients that don't declare the extension receive zero view-related metadata, maintaining byte-for-byte compatibility with the pre-views contract
- **Security**: The round-status card is self-contained with no external resources, relying on host CSP (`default-src 'none'`) and inline script/style allowance
- **Multi-instance safety**: Correlators adopted from other Cloud Run instances are treated as not view-capable, preventing assumptions about what they can render
- **Session lifecycle**: The `noteTransportSession()` helper preserves `uiCapable` across mid-round session re-registration in `start` and `open_round`
- **Phase 0 scope**: Static card only (no polling), read tools only (no mid-delivery rendering), no app-only tools yet

https://claude.ai/code/session_01MmKqqxSCF6xdNwD1Ke487B

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,13 @@
 # ADMIN_UIDS=
 # DAILY_GENERATION_QUOTA=20
 
+# --- Remote MCP --------------------------------------------------------------------
+# MCP_UI=true turns on MCP Apps views (SEP-1865) on /api/mcp: a ui:// resource registry
+# and _meta.ui on the tools that have a view. Off by default, and even when on it only
+# ever reaches a client that declared the io.modelcontextprotocol/ui extension — every
+# other client sees the same tools/list and initialize result as before.
+# MCP_UI=false
+
 # --- Hosting -----------------------------------------------------------------------
 # PORT=3001
 # HOST=127.0.0.1
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -1877,3 +1877,141 @@ describe('POST /api/mcp (BY-05)', () => {
     });
   });
 });
+
+describe('MCP Apps views (SEP-1865, Phase 0)', () => {
+  let app: FastifyInstance | null = null;
+  const previousFlag = process.env.MCP_UI;
+
+  afterEach(async () => {
+    await app?.close();
+    app = null;
+    if (previousFlag === undefined) delete process.env.MCP_UI;
+    else process.env.MCP_UI = previousFlag;
+  });
+
+  const UI_EXTENSION = 'io.modelcontextprotocol/ui';
+  const UI_MIME = 'text/html;profile=mcp-app';
+
+  async function initializeWith(instance: FastifyInstance, declaresUi: boolean) {
+    const res = await mcpCall(instance, 'initialize', {
+      protocolVersion: '2025-11-25',
+      capabilities: declaresUi ? { extensions: { [UI_EXTENSION]: { mimeTypes: [UI_MIME] } } } : {},
+      clientInfo: { name: declaresUi ? 'view-capable-host' : 'plain-agent', version: '0' },
+    });
+    expect(res.statusCode).toBe(200);
+    return {
+      sessionId: String(res.headers['mcp-session-id']),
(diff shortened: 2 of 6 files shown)


gtanczyk:
Addressed all three review points in `84463b6`. Both were fair — the first contradicted the compatibility claim this change is built around.

**Resource methods were on a weaker gate than `_meta.ui`** (comments on L3499/L3501). `resources/list`, `resources/templates/list` and `resources/read` answered for any session once `MCP_UI` was on, so a legacy client probing them would have seen a new surface instead of the `-32601` it got before views existed. They now sit on exactly the same gate as the tool metadata: the flag **and** a client that negotiated the extension. Tests cover a non-declaring session and a caller with no session at all being refused on all three.

I considered leaving `resources/read` open on the grounds that the URI is only obtainable from `_meta.ui` and the payload is a static first-party template with no authority — but the uniform rule is easier to reason about and to test, and the debuggability it cost is worth little.

**Media-type comparison was too strict** (comment on `mcp-ui.ts` L65). Parameters legally carry surrounding whitespace and quoted values, so `text/html; profile=mcp-app` and `text/html;profile="MCP-App"` would have been read as not view-capable. `clientDeclaresUi` now canonicalises before comparing; tests pin the accepted spellings alongside the ones that must still be rejected (`text/html` with no profile, the profile on another base type).

One limitation the first fix makes sharper, now recorded in a comment at the gate: capability lives on the transport correlator, so a session adopted by another instance reads as not capable and would be refused a resource read mid-round. Irrelevant today (`--max-instances 1`, and a spike runs against one process), but a durable view surface can't key off the correlator alone — that's a Phase 1 design input rather than something to patch here.

CI was green on the pre-fix rebased head (all 7 checks); this push re-runs it.

---

PR #578, opened 2026-08-04
Show share link for published games in Studio

gtanczyk:
## Summary
Updated the share control in Creator Studio to remain visible for games that are live in the catalog, while removing the draft toggle since published games have no draft state to toggle.

## Key Changes
- **Share control visibility**: Changed from hiding the share button entirely for live games to keeping it visible but in a simplified "live" mode
  - Removed the `canShareDraft` check that hid the control for published games
  - Introduced `canShare` (shows control if game has slug and isn't abandoned) and `shareIsLive` (indicates if game is published)
  - Share button now appears for both draft and published games

- **DraftShareControl component**: Added `live` prop to conditionally render UI based on game state
  - When `live=true`: hides the draft toggle switch, shows only the permalink and copy button
  - When `live=false`: shows the existing draft toggle and conditional hint text
  - Updated title and hint text to reflect whether sharing a draft or published game

- **UI/UX improvements**:
  - Added `.is-live` CSS class to style the live share state differently
  - Updated aria-labels and section labels to use appropriate titles for draft vs. live games
  - Permalink remains visible and copyable in both states

- **Localization**: Added new translation keys for live game sharing
  - `studioPanel.share.liveTitle`: "Share the game"
  - `studioPanel.share.liveHint`: Explains that the game is live and anyone with the link plays the latest published version
  - Added Polish translations for both new keys

- **Tests**: Updated test expectations to verify share control remains visible for published games and that the toggle is hidden in live mode

## Implementation Details
The change addresses a UX gap where published games had no share affordance in the Studio interface. The permalink is the same before and after publication, so creators still need a way to share it even after going live. The solution keeps the share control but removes the draft toggle since there's nothing to toggle for published games.

https://claude.ai/code/session_01MRpsXtLNv4CbWNV52dVNC8

### DIFF
--- a/apps/web/src/CreatorStudioView.test.tsx
+++ b/apps/web/src/CreatorStudioView.test.tsx
@@ -504,9 +504,18 @@ describe('CreatorStudioView', () => {
     });
     expect(setDraftShared).toHaveBeenCalledWith('token-draft', true);
 
-    // Published games already live at /play/<slug> — no draft-share head control.
+    // Published games have no draft switch — the link is already public — but the head
+    // control stays so the permalink is still one click away.
     await rerender({ selectedGame: 'live-game' });
-    expect(container.querySelector('[data-testid="studio-head-share"]')).toBeNull();
+    const liveShare = container.querySelector<HTMLButtonElement>('[data-testid="studio-head-share"]');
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex

PR #583, opened 2026-08-05
byoca: preview deliveries carry frames, and say which lane took them

gtanczyk:
## Why

The platform half of preview stills. Pairs with [games#537](https://github.com/gamedevpl/www.gamedev.pl-games/pull/537).

A preview delivery used to store a playable document and nothing else, so an agent that cannot run the game had prose and a URL it can't fetch. The live staged preview shows the **creator** a playable game — it does nothing for the agent.

## What changed

- **`gate-runner`** asks for `--preview-stills` on preview runs and stores whatever the harness captured, alongside `preview.html`. Never `bundle.html` — the publish seal is untouched, and the caller still writes `previewGate`, not `gate`.
- **`previewGate.screenshot`** names the first frame, so the media read doesn't infer it from a bucket listing.
- **`/api/agent/build/media`** serves either lane and reports **`gate.lane: 'preview' | 'publish'`**. **Publish wins when a delivery has both** — it's the later, fuller run, and its media is what a creator would be shown.

The lane field is load-bearing: preview proves the game typechecks, smokes and assembles, **not** that it's sealed. Without it an agent reading `green` would report a game publish-ready when it isn't.

## Cost and control

- On by default: the marginal cost is a few seconds on a build that already installed Chrome in order not to use it.
- Bounded by `SELF_BUILD_DELIVERY_CAP` — 20 deliveries per round, shared by preview and publish, so this cannot run unbounded.
- **`GATE_PREVIEW_STILLS=0`** turns it off with no deploy, and the harness stage is advisory — a runner with no browser loses frames, never a verdict.

## Verification

`type-check`, `lint`, full API suite: **2323 tests green**. New coverage for preview-lane media, publish-wins precedence, and the kill switch. The existing gate-runner test that pinned the old preview args is updated rather than relaxed.

`byoca-mcp` skill updated in the same commit, as its contract requires.

## Merge order

games#537 first (harmless alone — nothing requests the flag until this ships), then this.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PrXBSADXMndFNkj9FfEkMn)_


gtanczyk:
@codex

PR #586, opened 2026-08-05
Thread MCP_UI through both deploy paths

gtanczyk:
MCP Apps views (SEP-1865) shipped in #573 behind `MCP_UI`, but nothing carried that flag to Cloud Run — so there was no way to turn views on in production without the change reverting itself.

`--set-env-vars` replaces the whole env map, so a value set by hand on the service, or threaded by only one of the two supported deploy paths, is dropped by the next deploy from the other one, silently. That is the failure this repo already paid for on 2026-08-04, when a hand-set `TRANSLATE_BUILD_LOG=false` vanished under an unrelated deploy and the spend leak it had fixed resumed unnoticed. `deploy-api.sh` records that incident in its own comments; this follows the rule it states.

## Changes

- **`.github/workflows/deploy.yml`** — threads `MCP_UI` from the repo variable, alongside `SEED_DISPATCH` / `EDITOR_ASSIST` / `CODE_LANE`.
- **`infra/deploy-api.sh`** — adds `MCP_UI` to the shared flag loop, so a deploy from the script cannot drop what the workflow set. Documented in the script's env header.
- **`apps/api/src/mcp-ui-flag.test.ts`** — pins both directions, and asserts neither path pins the flag on with a literal. Mirrors the guard `remix-debug-flag.test.ts` already carries. (The literal check caught a real slip while writing this: the env-header doc line originally read `MCP_UI=true`.)

## Blast radius

Absent, or anything other than `"true"`, leaves the endpoint exactly as it runs today. Even switched on, it is inert for any client that does not negotiate the `io.modelcontextprotocol/ui` extension — those clients get the same `initialize` result and `tools/list` they got before views existed, which #573's tests pin. So the exposure is limited to hosts that explicitly opted in.

## Verification

`bash -n` on the script, YAML parse on the workflow, new tests plus the existing MCP suites green, lint clean. Production probed before the change: `initialize` returns only `{"tools":{"listChanged":false}}`, confirming views are off there today — that same probe is the after-check once this deploys with the variable set.

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -196,8 +196,9 @@ jobs:
           # rather than hand-set for the same reason as everything above. Turning it on is
           # inert for every client that does not declare the extension — they get the same
           # initialize result and tools/list they got before views existed — so the blast
-          # radius is limited to hosts that opted in. Absent or not "true" is that
-          # pre-views state, which is how the endpoint has always run in production.
+          # radius is limited to hosts that opted in. Enabled by "true" or "1",
+          # case-insensitive (mcpUiEnabled, apps/api/src/mcp-ui.ts); absent or any other
+          # value is the pre-views state the endpoint has always run in production.
           MCP_UI_VAL="${{ vars.MCP_UI }}"
           if [ -n "$MCP_UI_VAL" ]; then
             ENV_VARS="${ENV_VARS}|MCP_UI=${MCP_UI_VAL}"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -191,6 +191,18 @@ jobs:
             ENV_VARS="${ENV_VARS}|CODE_LANE=${CODE_LANE_VAL}"
           fi
 
+          # MCP Apps views (SEP-1865) on /api/mcp: a ui:// resource registry and _meta.ui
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

gtanczyk:
Both Copilot comments were the same point and both were right — fixed in `793c12d`.

The deploy comments claimed only `"true"` opens views, but `mcpUiEnabled` accepts `"true"` **or** `"1"`, case-insensitively. Docs for a production toggle being narrower than the parser is exactly how someone sets `MCP_UI=1`, sees it work, and then can't explain why. Both paths now name both values and point at the parser.

Fixing the wording exposed something worse in my own guard, so that got reworked too. The `is never pinned on` test originally matched `MCP_UI=(true|"true")` anywhere in the file, which false-positived on the env-header doc line. My first fix anchored it to end-of-line — which passes, but is nearly useless: in the workflow a real pin sits **mid-string** inside the `ENV_VARS` list, so the anchored version would have sailed past the thing it exists to catch. It now strips comment lines (a pin is code; the header legitimately names the enabling values) and matches unanchored on either spelling.

Verified rather than assumed: injecting `|MCP_UI=true` into `ENV_VARS` makes the test fail, and removing it makes it pass again.

Codex reviewed `b74337c` and found no issues. CI was green on that commit; this push re-runs it.

---

PR #589, opened 2026-08-05
Carry MCP view capability in the correlator, not in per-instance memory

gtanczyk:
First piece of Phase 1, landed on its own because it is a correctness fix rather than a feature: it removes the one constraint review on #573 flagged as blocking a durable view surface.

## The problem

Phase 0 recorded *"did this client negotiate views?"* on the in-process transport session map. Cloud Run is multi-instance and clients do not pin to a revision — that is precisely why the adopt path in `handleJsonRpc` exists. A request adopted by another instance read as **not** capable, so `tools/list` would drop `_meta.ui` and `resources/read` would refuse, breaking a view mid-round with no signal to the user.

`--max-instances 1` hides this today. It is not a safe foundation for Phase 1's dashboard, which reads resources repeatedly over a round's lifetime.

## The fix

`initialize` mints the session id with an HMAC suffix that any instance can verify with the shared secret. Capability travels in the correlator the client already echoes on every request:

- **Stateless** — no shared store, no coordination.
- **Unforgeable** — the marker is an HMAC over the base id; a transplanted, tampered, or wrong-secret marker fails.
- **Invisible to everyone else** — a client that never negotiates gets an unchanged id, and an id without a valid marker is simply not view-capable.

The transport map goes back to being a TTL/adoption record with no capability field.

One subtlety worth flagging for review: the marker is base64url and can itself contain the `-u` separator, so it is **sliced from the end** rather than found by scanning. Scanning would split in the wrong place for a minority of ids — the kind of bug that shows up as an intermittent, unreproducible view failure.

## Tests

The one that matters: a correlator negotiated on one app instance is honoured by a **second app instance that never saw the `initialize`** — `tools/list` still carries `_meta.ui` and `resources/read` still serves the card. I verified it fails when the marker is stubbed out, so it is testing the mechanism rather than passing by construction.

Alongside it: forged, tampered, transplanted and wrong-secret markers all refused; unmarked correlators get `-32601` on `resources/list`; and the marked id stays a legal `Mcp-Session-Id` under the existing regex.

Full API suite green (2357), lint clean.

---

### DIFF
--- a/apps/api/src/mcp-presence.ts
+++ b/apps/api/src/mcp-presence.ts
@@ -25,6 +25,10 @@ const NO_PULSE = new Set([
   'patch_source_file',
   'clear_staged_sources',
   'ack_inbox',
+  // The round view polls this while a creator watches. A human with a chat window open
+  // is not an agent working: pulsing here would refresh the heartbeat, hold off the
+  // quiet stall, and keep self→platform handoff locked for as long as the tab is open.
+  'get_round_status',
 ]);
 /**
  * Closed vocabulary for Studio thought headlines. Client translates via
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -2065,16 +2065,45 @@ describe('MCP Apps views (SEP-1865, Phase 0)', () => {
     );
   });
 
-  it('treats a correlator adopted from another instance as not view-capable', async () => {
+  it('treats an unmarked correlator as not view-capable', async () => {
     process.env.MCP_UI = 'true';
     const store = new InMemoryStore();
     await seedJob(store);
     app = await createApp(store);
 
-    // Cloud Run is multi-instance: this correlator was negotiated somewhere else, so
-    // this instance cannot know what it renders. Degrade to text, never assume UI.
+    // No signed marker, so nothing claims this client negotiated views.
     const foreign = 'fedcba9876543210fedcba9876543210fedc';
     const tools = await listTools(app, foreign);
     expect(tools.every((tool) => tool._meta === undefined)).toBe(true);
+
+    const listed = await mcpCall(app, 'resources/list', {}, { 'mcp-session-id': foreign });
+    expect(listed.json().error?.code).toBe(-32601);
+  });
+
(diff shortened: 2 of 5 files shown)


gtanczyk:
@codex

gtanczyk:
All four findings were real. Fixed in `f9d41bb`.

**Codex — `open_round` has no credential for the view.** Correct, and it was my bug. `open_round` returns `{jobId, slug, alreadyOpen}` and takes no `sessionKey`; its own description sends the agent to `start()` for one. A card opened there would have sat on "Reading round status…" indefinitely. Removed it from the launchers — `start()` follows immediately and opens the view properly. I did not take the other option offered (minting a key for that path), because `open_round` deliberately does not issue one and changing that would bend the agent contract for a UI convenience.

Chasing it down showed the bug was **wider than the report**: the view treated `sessionKey` as a precondition for polling, but Bearer-authenticated clients never pass one to *any* tool, so the same hang was reachable through `get_gate_verdict` and `submit_sources` too. The key is now an optimisation — the view polls with whatever credential the host's connection already carries, stops after two failed attempts with an honest message instead of a permanent spinner, and retries once if a key turns up late.

**Copilot — stale screenshot.** Right. Bytes only arrive when the frame changes, so the old code could leave a previous frame on screen with an out-of-date caption, and never cleared it when a round had no shot at all. Now three explicit cases: new frame with bytes, unchanged frame (keep the image, refresh the caption), no shot (hide and clear).

**Copilot — `get_round_status` reachable by name.** Right, and worth closing: absent from `tools/list` is not the same as unreachable, and `visibility: ["app"]` is a contract. It is now refused server-side for any client that has not negotiated views. It refuses in exactly the situations views are already unavailable, so it cannot break a working view. The test sends a **valid round credential**, so it proves the app-only gate is doing the refusing rather than the missing-credential challenge — my first attempt at that test got a 401 and never reached the guard at all.

**Copilot — the vacuous test.** Fair, and I had even written a comment admitting it. It now fails when the search finds no candidate rather than silently skipping its assertion.

Re-verified in headless Chromium after the polling changes: the first call goes out keyless, the second carries `sessionKey` and `sinceShotId`, both renders correct. Full API suite green (2362), lint clean.

---

PR #591, opened 2026-08-05
Stop the round view flashing a failure, and polling a settled round

gtanczyk:
Two bugs in the live round view, both found by watching the card in Claude rather than in tests. A screen recording of a real round caught the first; the second was visible in a still.

## The card flashed an error that was never true

The card went **"Reading round status…" → "Could not read round status here." → the full verdict**, all within about a second.

My own review fix on #589 caused it. I made the view poll as soon as it initializes, reasoning that a Bearer-authenticated host would carry the connection's credential into an app-only call. **Claude does not** — it refuses the keyless call. I was rendering that refusal as a user-visible failure. The session key then arrives with the opening tool's result, the retry succeeds, and the card fills in correctly — so the card was right within a second, having shown the user an error that was never true.

A poll made before the view has a key is now **speculative**: a refusal is logged, not painted. Only if no key arrives within 20s does it become a real failure worth showing — which keeps the honest message for the case it was written for (a host that genuinely refuses app-only calls).

This also answers a question the code was guessing at: **a view does not inherit the connection's credential in Claude.** The session key is required, not an optimisation.

## The card polled forever

`preview_passed` is not terminal, so a card left open kept polling every 30 seconds even though the agent had stopped and the verdict had settled — roughly 120 requests an hour per open tab, for state that cannot change until someone acts, and that act opens a fresh card. It now stops when the agent has ended and the gate is no longer pending.

## Verification

Reproduced in headless Chromium against a host that refuses keyless calls the way Claude does: the summary goes `Reading round status…` → the verdict with **no error in between**, and the call count stops rising once the settled read lands.

Worth noting how nearly this verification failed: my first harness sampled the summary text from inside the page and silently observed nothing, because the sandboxed frame blocks `contentDocument` — it reported "no flash" from an empty list. Sampling now runs from the driver side, so an empty observation cannot be mistaken for a clean one.

Full API suite green (2362), lint clean.

---

### DIFF
--- a/apps/api/src/mcp-ui.test.ts
+++ b/apps/api/src/mcp-ui.test.ts
@@ -206,6 +206,15 @@ describe('ui resources', () => {
     expect(html).toContain('Could not read round status here.');
   });
 
+  it('retries when the key lands while the keyless poll is still in flight', () => {
+    // noteSessionKey's own poll() is dropped as already-in-flight, so the refusal
+    // handler has to retry or the card waits out the whole give-up timer holding a
+    // usable credential. Reproduced in the browser harness: without this, the view
+    // makes exactly one call and never recovers.
+    const html = readUiResource(ROUND_STATUS_RESOURCE_URI)?.text ?? '';
+    expect(html).toMatch(/if \(speculative\) \{[\s\S]{0,400}?if \(sessionKey\) \{[\s\S]{0,80}?poll\(\);/);
+  });
+
   it('stops polling once the agent has stopped and the gate has settled', () => {
     // Nothing further can arrive, so an open tab must not cost a request every 30s
     // for as long as it stays open.
--- a/apps/api/src/mcp-ui.test.ts
+++ b/apps/api/src/mcp-ui.test.ts
@@ -195,6 +195,24 @@ describe('ui resources', () => {
     expect([...MCP_UI_APP_ONLY_TOOLS].every((name) => name.startsWith('get_'))).toBe(true);
   });
 
+  it('does not paint a failure for the poll it makes before it has a key', () => {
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

gtanczyk:
Both reviewers found the same race independently, and they were right. Fixed in `be41202`.

`noteSessionKey` calls `poll()`, but `poll()` returns early while a request is in flight. So when the host delivers the key **before** the keyless call's refusal comes back — the ordinary ordering in Claude, since both are async host messages — that `poll()` was dropped on the floor, and the refusal handler then armed the 20-second give-up timer while already holding a usable credential. The card would sit on "Reading round status…" for twenty seconds and then show a failure it had everything it needed to avoid.

The refusal handler now retries immediately when a key has since arrived. It stays bounded: the retry is no longer speculative, so a second failure takes the normal give-up path rather than looping.

**Verified in both directions**, because a passing test proves nothing on its own here. I reproduced the exact ordering in the browser harness — delivering the key while deliberately holding the keyless call open:

| | calls made | summary |
|---|---|---|
| without the fix | `["keyless"]` | stuck on "Reading round status…" |
| with the fix | `["keyless", "keyed"]` | recovers to the verdict |

Two smaller things came out of this:

- A comment claimed a view inherits the connection's credential. The Claude run disproved that — the keyless attempt is a hopeful first try, not the expected path. Corrected, since a comment asserting the opposite of observed behaviour is worse than none.
- The compiler caught me writing backticks inside that comment, which terminate the enclosing template literal. Worth noting for anyone editing this file: the whole view is one TS template string, so backticks and `${` in comments break the build.

Full API suite green (2362), lint clean.

---

PR #590, opened 2026-08-05
Remix: save remixed game into your Studio

gtanczyk:
Adds the missing remix exit: after you change a published game and like the result, **Make it mine** forks it into a private Studio draft under a new slug you own.

## What it does

- Earned CTA in the Remix panel (only after a change has landed) — en + pl
- `POST /api/remixes/:id/save` copies remixed sources into a new job/slug
- Works for **both catalog eras**:
  - **Store-era** — sources already in the remix session
  - **Repo-era / catalog** — loads a full delivery set at save via new `getGameDeliverySources` (TS import graph + fixed files like `SPEC.md` / `index.html`)
- Bakes client params / painted content into `EDITOR.json` (+ regenerated `game/editor-content.ts`)
- Writes `preview.html` so Studio can play immediately
- Provenance: `origin: 'remix'`, `forkedFrom: { slug, version }` (store version or games-repo ref SHA)
- Spends a normal creation quota / creation-gate slot (same as submitting a new game)

## What it deliberately does not do

- **Never publishes** to the catalog — preview-lane only, no `gate.green`, so the operator publish path refuses it
- Not contribute-back (that’s #559 Propose) — this is a private owned draft, the opposite direction

## Tests

- `remix-save.test.ts` — change detection + EDITOR.json bake
- `remix.test.ts` — store-era save, repo-era/catalog save, `no_changes` refusal
- `github-client.test.ts` — `getGameDeliverySources` joins graph + fixed files

Related: separate from #559 / ops #70 (proposals). Catalog derivative publish remains legal-gated later work.

[...]

PR #592, opened 2026-08-05
Round card: stop showing a stale verdict, say what it means, offer the next move

gtanczyk:
Everything here came from watching real rounds in Claude, not from tests. The live view works — a gate verdict reached the creator after the agent's turn had ended, with no polling and no tokens — but three things it said were wrong, and it never said what to do about any of them.

## A fresh round showed the previous round's verdict

The worst of the three. A new round opened with the **last** round's `kit_outdated` on the card — same delivery id, same timestamp — telling a creator their fresh round had already been refused.

`get_round_status` asked the channel for the gate, and the channel answers for the job's latest delivery whatever round produced it. A round with `roundDeliveryCount === 0` has no verdict of its own, so none is reported now.

## The card contradicted itself

*"The agent is building."* directly above *"The agent has stopped."* The headline led with the job phase, which lags; the note came from the stall state, which is current. Phase is now the last resort: a delivery sitting in a pending gate reads "Delivered — the gate is checking it", and an agent that stopped without delivering says exactly that. The stall row is dropped when it would only repeat the headline.

## The headline was instructions the creator can't act on

`kit_outdated` returns the same long agent-facing text as **both** summary and report, so the card printed it twice — once as its headline, where "submit_sources({ fromLatestDelivery: true, mode, kitEngineRef })" is not something a human can do. Long summaries now go to the detail block and the headline uses our own copy; the detail is skipped when it would duplicate the headline.

## And the missing half: what now?

The card knows things Studio doesn't. Studio says only *"your agent finished this round, send a note to continue"* — it doesn't know the gate refused the delivery. So the card now offers one contextual action per outcome: rebuild against a rotated kit, ask for a fix on red or a failed preview, ask to publish after a green preview.

Sent as `ui/message` in the shape SEP-1865 defines — `{ role, content: { type, text } }` — **read from the spec rather than guessed**, since a wrong shape means a button that silently does nothing. It appears only once the agent has stopped, because a live agent acts on a red gate itself and a second instruction would talk over it. If a host won't post for us, the text stays on screen to copy rather than leaving a dead button.

## Verification

Headless Chromium against a fake host: the button posts the documented payload and settles to "Sent to the agent"; `kit_outdated` renders a short headline with the detail below it once, not twice. Full API suite green (2365), lint clean.

## Unrelated finding worth recording

ChatGPT **does** discover the template — `ui://gamedevpl/round-status` is listed under its Plugins → Templates with no Apps-SDK-specific metadata, which answers the open Q3 about whether pure SEP-1865 is enough. It does flag that a widget CSP and a unique widget domain are required *for app submission*. Not addressed here: I would rather look up the exact metadata contract than guess at field names.

---

### DIFF
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -2007,7 +2007,6 @@ describe('MCP Apps views (SEP-1865, Phase 0)', () => {
     // The tools that open the round view, and nothing else.
     expect(tools.filter((tool) => tool._meta?.ui?.resourceUri !== undefined).map((tool) => tool.name)).toEqual([
       'start',
-      'submit_sources',
       'get_gate_verdict',
     ]);
   });
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -2110,6 +2110,9 @@ describe('MCP Apps views (SEP-1865, Phase 0)', () => {
       retryAfterSeconds: number;
     };
     expect(status.phase).toBeTruthy();
+    // Nothing delivered this round, so there is no verdict for it. The channel would
+    // happily answer with the previous round's, which read as "this round was refused".
+    expect((first.structured as { gate: unknown }).gate).toBeNull();
     expect(status.note?.text).toContain('wiring the HUD');
     expect(status.shot?.png).toBe(TINY_PNG);
     expect(status.retryAfterSeconds).toBeGreaterThan(0);
--- a/apps/api/src/mcp-server.test.ts
+++ b/apps/api/src/mcp-server.test.ts
@@ -2082,6 +2082,40 @@ describe('MCP Apps views (SEP-1865, Phase 0)', () => {
     expect(after?.agentEndedAt).toBe(before?.agentEndedAt);
   });
 
+  it('still reports the verdict that closed the round, when read as a terminal receipt', async () => {
+    process.env.MCP_UI = 'true';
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex

gtanczyk:
Rebased onto `master` (which had moved past #590/#591), and both review findings are fixed in `2929698`.

**Codex — terminal-receipt reads.** This one was a genuine regression inside my own stale-verdict fix, and I checked the premise before acting: `store.ts` resets `roundDeliveryCount` to 0 in the `closes` branch. So "delivered nothing, therefore no verdict" also hid the **green verdict that had just closed the round** — read through a one-behind session key, which is precisely what terminal receipts exist to allow. The card would have shown green from its opening payload, overwritten it with nothing, and gone back to polling a round that was already finished.

Terminal receipts are now exempt, because a receipt is a closed round being read rather than a fresh one with no deliveries. Verified by mutation rather than by a passing test: removing the exemption fails the new test, restoring it passes.

**Copilot — focus state.** Fair. The button had hover and disabled styling but nothing for keyboard focus, and its custom colours can leave the platform's default ring invisible against this card. Added `:focus-visible` with the accent.

On the rebase: the branch still carried the pre-squash commits from #591, so a plain rebase conflicted with content already in `master`. Replayed only the genuinely new commits with `--onto` instead, then checked the result rather than trusting it — no conflict markers, `submit_sources` still absent from the launchers, the action button intact, full suite green (2366).

---

PR #593, opened 2026-08-05
Declare the view's CSP and origin, and tell the model what the view learned

gtanczyk:
Three follow-ups, each checked against the specs rather than assumed.

## 1. The ChatGPT warnings — fixed

ChatGPT's plugin settings flagged our template with *"Widget CSP is not set"* and *"Widget domain is not set… required for app submission."* The resource now declares both, using the modern standard form:

```json
"_meta": { "ui": { "csp": { "connectDomains": [], "resourceDomains": [], "frameDomains": [] }, "domain": "https://www.gamedev.pl" } }
```

Empty by design and identical in effect to the deny-all host default — the card inlines everything and reaches the network only through the host — but **stated rather than assumed**, which is what ChatGPT wants before accepting a template for submission. `frameDomains` stays empty until Phase 2 actually needs a nested frame; opting in is documented to trigger stricter app review, so it isn't something to declare speculatively.

The `domain` identifies the view and grants it nothing on our site: the card is served as inline HTML and never runs with our origin's authority.

## 2. Tool visibility — we were marking it correctly, with one redundancy

Checked against the spec: `_meta.ui` is the right key, and `visibility` **defaults to `["model", "app"]` when omitted**. So spelling that out on a launcher said nothing. Launchers now carry just `resourceUri`; the app-only tool keeps its explicit `["app"]`, which is the part that isn't the default.

## 3. `ui/update-model-context` — yes, and it closes the loop

This turned out to be the missing piece of watching-without-polling. By the time a gate answers, the agent has usually ended, so **nothing in the conversation knows the outcome** — the creator can see it on the card, but the model can't.

The view now pushes the verdict when it lands. Per the spec the host holds it until the next user message rather than waking the model, so it costs no turn: a creator who then says "fix it" is talking to an agent that already has the verdict instead of one that must spend a call discovering it. Sent once per distinct verdict — polling must not re-announce the same thing — and never while the gate is still pending.

## Verification

Headless Chromium: exactly one context update carrying the verdict, and the action button still posting its `ui/message`. Full API suite green (2369), lint clean.

## One thing still unverified

Screenshots travel as `data:` URIs, and no real host has yet rendered a card that had one — every Claude round so far has had no frame at the moment the card polled. If a host's CSP blocks `data:` images, that path fails silently. Not fixable by declaring domains (a `data:` URI has none); it would need observing a round that actually has a frame.

Sources: [MCP Apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) · [OpenAI Apps SDK reference](https://developers.openai.com/apps-sdk/reference)

---

### DIFF
--- a/apps/api/src/mcp-server.ts
+++ b/apps/api/src/mcp-server.ts
@@ -3796,7 +3796,8 @@ export async function registerMcpServerRoutes(app: FastifyInstance, options: Mcp
               const uiMeta = appOnly
                 ? { visibility: ['app'] }
                 : uiResourceUri
-                  ? { resourceUri: uiResourceUri, visibility: ['model', 'app'] }
+                  ? // visibility defaults to ["model", "app"], so a launcher says nothing.
+                    { resourceUri: uiResourceUri }
                   : null;
               return {
                 name,
--- a/apps/api/src/mcp-ui.test.ts
+++ b/apps/api/src/mcp-ui.test.ts
@@ -152,6 +152,31 @@ describe('ui resources', () => {
     expect(readUiResource('')).toBeNull();
   });
 
+  it('declares its CSP and origin rather than relying on the host default', () => {
+    // Same effect as the deny-all default, but stated: ChatGPT will not accept a
+    // template for submission without it, and an empty frameDomains is a deliberate
+    // signal that nothing is nested yet (declaring one triggers stricter review).
+    for (const meta of [uiResourceDescriptors()[0]?._meta, readUiResource(ROUND_STATUS_RESOURCE_URI)?._meta]) {
+      expect(meta).toMatchObject({
+        ui: {
+          csp: { connectDomains: [], resourceDomains: [], frameDomains: [] },
+          domain: 'https://www.gamedev.pl',
+        },
+      });
+    }
+  });
+
+  it('tells the model what the view learned, once per verdict', () => {
+    // When a verdict lands the agent is usually gone, so nothing in the conversation
+    // knows it. The host holds this until the next user message.
+    const html = readUiResource(ROUND_STATUS_RESOURCE_URI)?.text as string;
+    expect(html).toContain("request('ui/update-model-context'");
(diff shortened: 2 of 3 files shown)


gtanczyk:
@codex

gtanczyk:
All three findings were right. Fixed in `e176726`, and fixing them surfaced a fourth that was worse than any of them.

**Codex — the report was missing from model context.** This was the important one. The app-only status result never reaches the transcript, so a follow-up "fix it" left the next agent knowing *that* a verdict landed but not *what broke*, forcing another call to rediscover it — which is most of what this feature exists to avoid. The report now travels with the verdict, truncated at 4k with a pointer to `get_gate_verdict` for the remainder, since this lands in the model's context whether or not it is ever read.

**Copilot — shallow freeze.** Correct: one `VIEW_CSP` object is handed out on every `resources/list` and `resources/read`, and `Object.freeze` doesn't reach the nested arrays. Frozen individually, with a test that a push throws.

**Copilot — type erasure.** Also correct, and my shortcut. Widening to `Record<string, unknown>` dropped the guarantee that a descriptor carries `uri`/`name`/`description`/`mimeType` and a read carries `uri`/`mimeType`/`text`. Restored as explicit types extended with a typed `_meta` — which also removed an `as string` cast the tests had needed.

**The fourth, found while fixing the third.** I wrote `'\n'` inside the view, and TypeScript consumed it and emitted a **real line break** into the view's JavaScript, breaking the string it sat in. `tsc` passed. Lint passed. Every string-matching test in this file passed. The card would simply have rendered blank in production.

The browser harness caught it; nothing in CI did. So there is now a test that extracts the embedded script and parses it, which fails when the escape is reverted. This file puts an entire JS program inside a TS template literal, so backticks, `${`, and escapes in *comments* are all live hazards — this is the second time one has bitten (the first was a backtick in a comment, caught by the compiler because it happened to terminate the literal).

CI was green on the previous head; this push re-runs it. Full suite green locally (2371), lint clean.

---

PR #595, opened 2026-08-05
Stop declaring ui.domain — Claude validates it, and ours was wrong

gtanczyk:
A live regression from #593, surfaced by the Claude console:

```
[mcp-ext-apps-host] ui.domain validation failed for connector "https://www.gamedev.pl/api/mcp"
```

## What I got wrong

I read "widget domain: string (origin)" and declared our origin. **It isn't ours.** Claude derives the expected value from the connector URL *the user typed*:

```
sha256(<connector URL>).hex.slice(0, 32) + '.claudemcpcontent.com'
```

Which means it varies per user, not just per host:

| connector URL as configured | expected domain |
| --- | --- |
| `https://www.gamedev.pl/api/mcp` | `f05e7b02…claudemcpcontent.com` |
| same, trailing slash | `96230950…claudemcpcontent.com` |
| apex spelling | `8935003b…claudemcpcontent.com` |

A static resource cannot know which one a given creator configured, so **any hardcoded value is wrong for somebody**. Omitting it is what worked before #593, and the only thing it gates is ChatGPT plugin submission — owner-gated and not imminent. When that day comes it has to be derived per request from the URL the client actually called. The formula and the three-way table are recorded at the site so nobody rediscovers this the hard way.

## Also: two ChatGPT compatibility keys

Separately, ChatGPT **rendered the card** for the first time — a real step forward — but showed a `CSP off` badge and *"Could not read round status here."*, i.e. it would not let the view call the app-only tool.

Both of its documented compatibility keys are now declared beside the standard ones:

- `openai/widgetCSP` on the resource (it showed `CSP off` against `_meta.ui.csp` alone)
- `openai/widgetAccessible` on the app-only tool
- `openai/outputTemplate` on the launchers

They say exactly what `_meta.ui` already says. Declaring both beats guessing which one a host actually reads — and this is a hypothesis to test in ChatGPT, not a confirmed fix.

Full API suite green (2371), lint clean.

---

### DIFF
--- a/apps/api/src/mcp-server.ts
+++ b/apps/api/src/mcp-server.ts
@@ -3817,13 +3817,21 @@ export async function registerMcpServerRoutes(app: FastifyInstance, options: Mcp
                   ? // visibility defaults to ["model", "app"], so a launcher says nothing.
                     { resourceUri: uiResourceUri }
                   : null;
+              // ChatGPT rendered the card but could not call the app-only tool from it.
+              // Its own compatibility keys say the same things as `_meta.ui`, so declare
+              // both rather than guess which one a host actually reads.
+              const openAiMeta = appOnly
+                ? { 'openai/widgetAccessible': true }
+                : uiResourceUri
+                  ? { 'openai/outputTemplate': uiResourceUri }
+                  : null;
               return {
                 name,
                 description: tool.description,
                 inputSchema: tool.inputSchema,
                 outputSchema: tool.outputSchema,
                 ...(tool.annotations ? { annotations: tool.annotations } : {}),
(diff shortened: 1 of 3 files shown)


gtanczyk:
Fixed in `a6fd8db`. `uiResourceMeta()` stopped returning `ui.domain` in this PR, but its doc comment still promised "its CSP and the origin it belongs to".

Small, and worth more than its size. A comment asserting the opposite of the code is what put the wrong value there in the first place — I read "widget domain: string (origin)" and declared ours. This file has now produced two such comments in a day, so the new one states what is returned *and* why the origin is deliberately absent, pointing at the block that explains Claude's per-connector-URL hash.

Nothing else changed. Lint clean, suite green.

---

PR #609, opened 2026-08-05
Make creator thumbnails open game pages

gtanczyk:
## What changed

- creator-profile thumbnails now open the canonical `/:handle/:slug` game page
- title and thumbnail share the same route and SPA navigation behavior
- the Play button still opens the theater directly, and owner-only Studio links are unchanged

## Why

The thumbnail looked like a browse affordance but sent visitors straight into `/play/:slug`, unlike the title. The card now has one consistent browse destination while retaining an explicit Play action.

## Measurement

No telemetry vocabulary changes. This preserves the existing visit funnel: thumbnail clicks produce the normal game-page `route_viewed`, and only the explicit Play action produces `play_started`. Questions 1–3 (first minute, session depth, acquisition) remain answerable from visit telemetry.

## Validation

Independent clean checkout:

- `npm run type-check`
- `npm run lint`
- `npm run test`
- `npm run build`

### DIFF
--- a/apps/web/src/CreatorProfilePage.test.tsx
+++ b/apps/web/src/CreatorProfilePage.test.tsx
@@ -66,15 +66,15 @@ afterEach(() => {
   document.body.querySelectorAll('.claim-handle-modal-card, .modal-backdrop').forEach((node) => node.remove());
 });
 
-async function renderPage() {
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #617, opened 2026-08-06
Play button, show_media, and the card fixes from a live ChatGPT session

gtanczyk:
Five commits, all driven by the owner testing the card in ChatGPT.

## 1. Play button

The step the flow exists for — the card shows what the agent built, the theater is where it gets played. **A link, not an embedded game**: the theater does fullscreen, pointer lock and touch; a chat card cannot. That is the reasoning that cut in-chat play from Phase 2, and this replaced a six-question spike with a link.

Only offered when something is genuinely playable.

## 2. `show_media` — because the agent cannot show what it can see

The owner's diagnosis, and the sharpest finding in this batch: **a tool result is input to a model, not output to a person.**

`get_gate_media` attaches frames as image blocks. Those reach the *model* — it looks at them, describes them, reasons about them — and there is no path back out. Asked to display them, ChatGPT answered *"the image attachments apparently didn't render in your view"* and the creator saw nothing.

A view is the only surface that puts pixels in the conversation without going through the model. `show_media` opens the card and names which delivery's frames to fetch — the part the existing gallery could not do, since it hung off the *current* round's gate and "show me the screenshot" opens a round that has delivered nothing.

It carries **no bytes**: the card fetches them over the app-only `get_round_media`, keeping a megabyte of base64 out of the model's context.

## 3. The card contradicted itself

The pill read `DISPATCHED` above *"The agent has stopped without delivering."*, and *"The agent has stopped."* repeated as a Note row. The call ledger confirms the card was factually right — both defects were presentational, and both were half-fixed already: the summary ranks the phase last (it lags reality) and the pill was left on the old rule.

The stall dedup needed **the escape doubled** — `/\.$/` written once in this template literal emits `/.$/`, *"strip any last character"*. It happened to look correct only because every stall line ends in a period.

## 4. Codex's two findings — the first would have shipped a dead button

**P1:** `playUrlFor` took the first entry of `WEB_ORIGIN`. That is a **CORS allowlist**, and production starts it with the Cloud Run service URL — so every production Play link pointed at an origin `redirect_domains` does not permit, and ChatGPT would refuse it. **Dead in production, fine everywhere we tested.** Now `canonicalAppBaseUrl()`, with the view's allowlist derived from the same function so they cannot drift again.

**P2:** `needs_changes` was treated as playable. A publish run that fails before assembly lands there with neither bundle nor preview, sending the creator to the exact "not available yet" page the button exists to avoid. Removed — a red gate is precisely when there may be nothing to play.

## 5. Sizing, and it was my own earlier fix

DevTools showed ChatGPT's wrapper pinned at `height: 400px` around ~250px of content.

`reportSize` sent `documentElement.scrollHeight`, which *was* the content height until `applyContainerDimensions` (#606) started setting `html{height:100%}` to fill a host-declared frame. After that the two became the same number: the card filled 400px, reported 400px, and the host had no reason to shrink.

Measuring the card breaks the loop. **Mutation-checked in the browser** against a fixed 400px frame: the old code reports `520`, this reports `287`.

Also as requested: **the wordmark opens gamedev.pl and the game name opens that round in Creator Studio.** Studio rather than `/play` because the Play button already covers playing, and Studio is valid whether or not anything is playable — none of the "not available" risk from P2. Both are click handlers rather than anchors, since the view is sandboxed without `allow-top-navigation` and an `href` is inert.

## Verification

Full API suite green (**2563** tests, 162 files), lint clean.

Browser harness: all three links post `ui/open-link` with the right URLs; the card reports content height inside a fixed frame; the Play button renders only on a playable round.

**Still unproven in a real host:** that ChatGPT opens these links on click. The `CSP off` badge clearing is strong evidence `redirect_domains` is in force, but a click is the proof.

---

### DIFF
--- a/.claude/skills/byoca-mcp/SKILL.md
+++ b/.claude/skills/byoca-mcp/SKILL.md
@@ -188,8 +188,14 @@ cache), so a resume cannot keep showing “finished this round” next to live p
   MCP tools refuse an already-issued key with a reconnect instruction. Do not restore a
   per-game compatibility path in UI, API routes, or MCP tool handling.
 - An opener is checked by `start`; later calls use the returned round-scoped `sessionKey`.
-- **`show_round` is the only tool that opens the creator's status card**, and it exists for
-  nothing else. Agents do not call it on their own — ChatGPT never did until the creator
+- **`show_media` is how the creator sees the game.** `get_gate_media` attaches frames as
+  image blocks, which reach the _model_ — it can look at them and describe them, and there is
+  no path back out. Asked to display them, ChatGPT answered "the image attachments apparently
+  didn't render in your view" and the creator saw nothing (2026-08-06). A tool result is input
+  to a model, not output to a person, so a view is the only surface that puts pixels in the
+  conversation. `show_media` carries no bytes: the card fetches them over the app-only
+  `get_round_media`, keeping a megabyte of base64 out of the model's context.
+- **`show_round` opens the creator's status card**, and exists for nothing else. Agents do not call it on their own — ChatGPT never did until the creator
   asked (2026-08-05) — so a view-capable session that has not opened one gets a bounded
   `warnings.code=card_unopened` nudge, the same remedy `call_end` needed. Never sent to a
   client with no views. It used to hang off `start` and `get_gate_verdict`, which made the card a side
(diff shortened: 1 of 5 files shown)


gtanczyk:
@codex

PR #679, opened 2026-08-08
Managed agent backend: a vendor-neutral provider seam

gtanczyk:
## Summary
Adds the abstraction layer for running the platform builder on a hosted agent platform, designed so the vendor can be replaced without touching anything above it.

- `managed-agent.ts` — the seam: start a session, observe it, list and read its outputs, stop it. No HTTP shapes, SDK types, beta headers, or vendor state words.
- `managed-backend.ts` — an `AgentBackend` over any provider. Delivery either comes from the agent over MCP (`submit_sources`, the rail every other round already uses) or is **pulled** from the session's output directory.
- `managed-provider-anthropic.ts` — the first adapter, and the only file that knows one vendor's wire format.
- `agent-backend-env.ts` — `MANAGED_AGENT_VENDOR` selects the adapter and takes the platform slot; builder routing, the job machine, the gate, Studio and self builds are untouched.
- `docs/managed-agent-backend.md` — the design record, how to exercise it, and how to add a vendor.

## The architecture objection, and why it is settled
Codex read the contract in `AGENTS.md` / `games-repo.md` — hosted agent execution removed for legal reasons, not a future phase — as blocking this. The owner's decision is that it does not apply, because the credential is a **metered first-party API key** rather than a seat on a subscription licensed to a human, and the sandbox is the vendor's. Both findings behind the original removal were properties of those two facts: `B0` needed a container of ours holding a credential, and `B1`/`B2` needed subscription terms that do not cover reselling a seat. `docs/games-repo.md` now draws that distinction, and `AGENTS.md`, the Copilot instructions and `docs/README.md` point at it.

## Review findings addressed
- **Bounded downloads.** The provider surface splits into `listOutputs` (paths, sizes, opaque handles) and `readOutput`. Caps refuse on the declared sizes *before* a byte is fetched, and a per-file budget re-checks the bytes on arrival for vendors that omit sizes.
- **A failed pull no longer spends the round.** `completed` with no candidate is how the job machine recognises an agent that finished without delivering, and it fails the job. When it is the pull that failed, the observation reports `in_progress`, so the work survives to the next poll.
- **Durable at-most-once.** `ManagedDeliveryLock` is an injectable durable guard: the holder delivers, the loser reports `in_progress` and picks up the candidate later, and a failed attempt releases the lock.
- **Token-billed sessions no longer carry a credit placeholder**, through one helper shared by both store implementations.
- Output paths reject traversal, absolute, Windows and NUL-containing paths before prefix stripping; zero-token usage objects are preserved.

## Convergence with architecture A, and its limit
Copilot does not move under `ManagedAgentProvider`, and `docs/managed-agent-backend.md` says why: of the seam's four capabilities it fits one. Its workspace is a branch staged with git writes rather than a list of files, it bills in credits and never reports tokens, and it has no outputs to list because delivery is a push. Widening the seam to fit would either make every capability optional — which stops it constraining anything — or make it carry branches, credits and output directories for implementations that each ignore two thirds. `AgentBackend` is already the shared abstraction.

What *is* shared is now shared, which fixed a defect:

- `AgentTaskState` moves to `agent-state.ts`. The neutral vocabulary was defined inside GitHub's client, and the job machine, the store and the managed seam all imported one vendor's file to name a state they share.
- `buildPrompt` moves to `build-prompt.ts` and takes a `DeliveryContract`. **The managed backend had been telling pulled rounds to `npm run submit` over the build channel while it read their output directory.** A pulled round is now told where to write, and told plainly that it has no progress channel, no inbox and no gate verdict it can still act on.
- The prose from the moved code becomes `docs/build-brief.md` rather than re-baselining the comment ratchet; `copilot-backend.ts` ratchets from 1251 to 976 words, `agent-tasks.ts` from 1183 to 1135.

## The probe, and the two bugs it caught
The backend is **not reachable from the running app**: `app.ts` builds the platform slot from `createPlatformBackendFromEnv`, so `MANAGED_AGENT_VENDOR` on a deployed service does nothing. Wiring the registry in is the same change as extracting the sink. So `npm run managed:probe -w @gamedevpl/api` drives one whole round — dispatch, two observes, harvest, cleanup — through the real backend and real caps against a stub vendor, with a printing sink. `--mcp`, `--prompt`, `--out` and `--vendor anthropic` cover the other shapes.

Running it found two things no unit test had, because the fakes only ever returned tidy paths:

1. **A `scratch/notes.md` was delivered into the game's source tree.** The path mapping treated anything that was not another game's directory as game-relative. Only `games/<slug>/` is a delivery now.
2. **A `media/cover.png` reached the sink**, which `forbiddenDeliveryPathReason` refuses on upload — media bytes are produced by our gate, never delivered. The harvest now applies the same rule, so a pull cannot store what an upload would reject.

The second one **drops** the offending files rather than refusing the delivery, and the doc explains why that is not a weakened check: an upload can make the agent fix the path and retry, while a pull happens after the session is gone, so failing the round over a stray screenshot would lose a game the gate would have accepted. The invariant — non-source files never reach the store — holds either way, what was dropped is logged, and the brief now tells a pulled round "sources only" so the drop is the backstop rather than the first defence.

The doc is also explicit that `managed-provider-anthropic.ts` is written from a *description* of the Managed Agents API rather than against it — the session paths, beta headers and files query are inferred, and the spike's two measured rounds were driven from the Console. The first real-key run should be a wire-format check, and is expected to fail on request shape.

## Validation
- [x] Full local green gate: `type-check`, `lint` (including the comment-prose ratchet), `test`, `build`
- [x] New tests for listing-time refusal, masked state on a failed pull, lock contention and release, MCP-only configuration, path selection, non-deliverable paths, and the delivery contract each mode promises
- [x] `managed:probe` green in both shapes

## Scope
No behaviour change for existing rounds: without `MANAGED_AGENT_VENDOR` the platform slot stays Copilot, and the app does not read it yet. Next change is extracting the channel's `submit_sources` route into a shared function — it supplies the sink, unblocks the registry wiring, and gives A and C one definition of "delivered".

<div><a href="https://cursor.com/agents/bc-4cf2e9e5-dd15-4c3a-8323-8e026e20823b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4cf2e9e5-dd15-4c3a-8323-8e026e20823b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


gtanczyk:
@codex

PR #682, opened 2026-08-08
Managed rounds: stop overriding the vendor agent, and make its config code

gtanczyk:
## The bug, found by running it
Two live rounds ended after a single 73-token reply with the session idle. I blamed the prompt; it was our code. `--mcp` made the Anthropic adapter send an `agent_with_overrides` block that **replaces** the configured agent's `tools` and `mcp_servers` wholesale. The real agent declares an `agent_toolset_20260401` (with `web_search`/`web_fetch` deliberately disabled) plus an `mcp_toolset`; the override flattened all of it, so the session had nothing it could call and did the only thing left — answered briefly and ended the turn.

Overriding is now opt-in (`overrideTools`, `--override-tools`). The same bounded run with the agent's own config left alone reaches `start` at 4.9s, unpacks the kit at 20s, and writes `render.ts` / `runtime.ts` / `index.html` continuously until the cap.

## Where the two minutes actually go
From the transcript of the run that worked:

| Phase | Time | Verdict |
| --- | --- | --- |
| `find / -iname "*copilot-instructions*"`, `ls /mnt/session` | 0–15s | **Our prompt caused it.** It described `games/<slug>/`, `shared/` and `.github/` as if a checkout existed. That sandbox has none. |
| `start` → error → `create_game` → `start` | 23–33s | The slug did not exist yet. A real job would already have one. |
| `get_brief` + `get_kit` + `show_round` | 43–46s | Correct, and parallel. |
| Unpack kit, then read ~78 KB of it via bash | 52–94s | Kit spelunking. Context 57k → 97k. |
| One 31s model request | 94–125s | Writing the game. We cut it off. |

So ~67s of a 125s round was avoidable overhead, and the cap landed on the generation step.

## What this PR changes
- **`overrideTools` is opt-in.** The vendor agent's configuration wins by default.
- **The clocked lane stops lying about the sandbox.** It says the unpacked kit is the only copy of the repository, and asks for a delivered rough draft over a better undelivered one, rather than reciting a tool sequence the agent's own system prompt already carries.
- **`infra/managed-agent.json` + `npm run managed:agent:apply`.** The agent definition is version-controlled rather than Console state; the script prints the resulting `MANAGED_AGENT_ID` and version, so a Console edit shows up as a version this file did not produce. `--dry-run` prints the request. The manifest is seeded from the live agent's current configuration, not invented.
- **`--keep` and a transcript dump in the probe.** Cleanup used to delete the session, which is why the last diagnostic got a 404. The probe now prints the tool timeline on exit.
- **An MCP-configured backend no longer requires a delivery sink.** `MANAGED_AGENT_MCP_URL` is enough, because that agent submits through the channel itself — which is what makes a UI test possible before the `submit_sources` extraction lands.

## The next lever, with a number on it
The remaining overhead is API archaeology: eight separate `grep`/`sed` passes over `game-kit.d.ts` looking for `input(`, `poly` vs `polygon`, `hud(`, at roughly 5–15s each. That is exactly what AB-06's kit digest was built to remove, and the seam already has the socket for it — `systemPrompt` is documented as "cacheable prefix, typically the published Creator Kit digest". We publish the digest and then never feed it. Wiring it is the next change and should be worth 45–60s of a two-minute round.

## Validation
- [x] Full green gate: `type-check`, `lint` (including the comment-prose ratchet), `test`, `build`
- [x] Live bounded round against the real API: no override, agent builds until the cap, transcript captured
- [x] New tests for both override modes, the clocked lane's copy, and the absent-checkout wording

## Note
The verification round left session `sesn_01RuGambVJeoMo1j4vJzBUSV` alive on purpose (`--keep`); delete it when you have finished looking at it.

<div><a href="https://cursor.com/agents/bc-4cf2e9e5-dd15-4c3a-8323-8e026e20823b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4cf2e9e5-dd15-4c3a-8323-8e026e20823b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -14,6 +14,7 @@
     "gate:run": "tsx scripts/run-gate.ts",
     "remix:probe": "tsx scripts/remix-lane-probe.ts",
     "managed:probe": "tsx scripts/managed-probe.ts",
+    "managed:agent:apply": "tsx scripts/managed-agent-apply.ts",
     "beta:approve": "tsx scripts/beta-approve.ts",
     "beta:invite": "tsx scripts/beta-invite.ts",
     "token:mint": "tsx scripts/access-token.ts mint",
--- a/apps/api/scripts/managed-agent-apply.ts
+++ b/apps/api/scripts/managed-agent-apply.ts
@@ -1,0 +1,45 @@
+// Applies infra/managed-agent.json. See docs/managed-agent-backend.md.
+
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+const args = process.argv.slice(2);
+const flag = (name: string) => args.includes(`--${name}`);
+const value = (name: string) => (args.includes(`--${name}`) ? args[args.indexOf(`--${name}`) + 1] : undefined);
+
+const apiKey = process.env.MANAGED_AGENT_API_KEY?.trim() ?? process.env.ANTHROPIC_API_KEY?.trim();
+if (!apiKey) {
+  console.error('needs ANTHROPIC_API_KEY (or MANAGED_AGENT_API_KEY)');
+  process.exit(1);
+}
+
+const manifestPath = value('manifest') ?? fileURLToPath(new URL('../../../infra/managed-agent.json', import.meta.url));
+const manifest = JSON.parse(readFileSync(manifestPath, 'utf8')) as { agent: Record<string, unknown> };
+const existing = value('agent-id') ?? process.env.MANAGED_AGENT_ID?.trim();
+
+if (flag('dry-run')) {
+  console.log(existing ? `POST /v1/agents/${existing}` : 'POST /v1/agents');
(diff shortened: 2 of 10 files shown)


gtanczyk:
@codex

PR #708, opened 2026-08-09
Add a switch to disable the managed (platform) coding agent, with honest UI

gtanczyk:
## Summary

Creators can always fall back to BYOCA (`self`), but until now an unconfigured or paused platform (Gamedev.pl) backend failed silently: a submission asking for `builder: 'platform'` just sat in `queued` forever with no explanation, and the builder picker kept offering an option that could never actually dispatch.

This adds an explicit, operator-controlled switch for the managed agent, with four distinguishable "why not" reasons, and makes the UI say so honestly instead of hiding the option.

- **`managed-availability.ts`** — resolves whether `platform` can be offered right now:
  1. operator switch `off` → *outage*
  2. operator switch `coming_soon` → *coming soon*
  3. no platform backend configured in this environment → *coming soon*
  4. shared or per-creator daily cap spent → *global limit* / *user limit*

  Reads from the same `opsConfig/creationLimits` document the creation breaker already uses (~60s propagation, no redeploy). Bots bypass it, same as the breaker.
- **Server-side enforcement** on new submissions, resumes, and handoffs targeting `platform`, refusing with `platform_builder_unavailable` + reason. An **in-flight platform round keeps running** — only new dispatches are gated. Availability is echoed on `GET /api/me/quota` and the submission status payload so the UI never has to guess.
- **UI never hides the option.** `BuilderChoice`, `BuilderModeBadge`, and the Studio "switch to platform" control show `platform` visibly disabled with a reason badge and note. Uses `aria-disabled` rather than the `disabled` attribute — a truly disabled button shows no tooltip in Chrome/Safari and drops out of the tab order, which would hide the reason from keyboard/screen-reader users. The create wizard blocks advancing past the builder step (and submitting from review) until the creator explicitly picks BYOCA, rather than switching them over automatically.
- **`CreationLimitsPanel`** (admin console) gains the mode selector plus both daily caps, so pulling the switch during an incident is a console click, not a deploy.

One thing worth flagging: on the next deploy, production will show *Coming soon* on the Gamedev.pl option unless a platform backend is actually configured there (`AGENT_TASKS_TOKEN` or the `MANAGED_AGENT_*` vars) — the "unconfigured reads as coming soon" behavior is deliberate per the requester's explicit confirmation.

## Test plan

- [x] `apps/api`: `tsc --noEmit` clean; full vitest suite (2846 tests) passes, including new `managed-availability.test.ts` and enforcement tests in `submissions.test.ts`
- [x] `apps/web`: `tsc --noEmit` clean; full vitest suite (1210 tests) passes, including new coverage in `BuilderChoice.test.tsx`
- [x] `apps/web`: i18n key-parity test passes for the new `builder.platform.unavailable.*` and `qa.builderBlocked` keys (en + pl)
- [x] Both packages build cleanly (`vite build` / `tsc -p tsconfig.json`)
- [x] ESLint clean on all changed files
- [ ] Manual click-through in a running Studio session (not done in this session)

https://claude.ai/code/session_01TBimyB3MuY7ugDHC7RMnF5

---
_Generated by [Claude Code](https://claude.ai/code/session_01TBimyB3MuY7ugDHC7RMnF5)_

### DIFF
--- a/apps/api/fixtures/games-repo/shared/assemble-contract.json
+++ b/apps/api/fixtures/games-repo/shared/assemble-contract.json
@@ -5,6 +5,7 @@
     "input",
     "collision",
     "world",
+    "grid",
     "path",
     "ai",
     "gameplay",
--- a/apps/api/src/admin.test.ts
+++ b/apps/api/src/admin.test.ts
@@ -483,7 +483,14 @@ describe('/api/admin/creation-limits', () => {
 
     expect(res.statusCode).toBe(200);
     const body = res.json() as CreationLimitsResponse;
-    expect(body.effective).toEqual({ paused: true, globalDailySubmissionCap: 25 });
+    expect(body.effective).toEqual({
+      paused: true,
+      globalDailySubmissionCap: 25,
+      managedBuilderMode: 'auto',
+      managedDailyCap: null,
+      managedDailyUserCap: null,
+      hasPlatformBackend: false,
+    });
     // A pause left on by accident is this feature's own failure mode, so the record of
     // who set it is part of the deliverable.
     expect(body.stored).toMatchObject({ paused: true, updatedBy: 'g:boss' });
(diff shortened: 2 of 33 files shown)


gtanczyk:
@codex

PR #710, opened 2026-08-09
Add get_kit_api endpoint and improve kit digest selection

gtanczyk:
## Summary

This change adds a new `get_kit_api` MCP endpoint that serves a compacted Creator Kit API digest to agents, and significantly improves how kit declarations are selected and presented when space is constrained. Previously, agents had no in-band way to discover what capabilities the platform supports (party/multiplayer, zones, commons, presence, etc.), leading them to search the public web for documentation that doesn't exist there.

## Key Changes

- **New `get_kit_api` endpoint** (`/api/agent/build/kit/api`): Serves the compacted kit digest with a larger budget (90 KiB) than the platform prompt lane (20 KiB), since it's paid once per round by agents that explicitly request it. Supports optional `engineRef` parameter, defaulting to the registry's current entry.

- **Improved kit digest selection algorithm**: Replaced simple line-pattern filtering with a sophisticated whole-block selection strategy that:
  - Splits declarations into top-level blocks and preserves them whole rather than orphaning individual lines
  - Ensures every module family (party, zone, commons, presence, etc.) gets representation before any family gets a second block
  - Reserves budget for oversized core declarations like `GameKitApi` and abridges them member-wise rather than dropping them entirely
  - Names omitted declarations in the output instead of silently dropping them

- **New declaration parsing utilities**:
  - `splitDeclarationBlocks()`: Parses `.d.ts` files into top-level declaration blocks
  - `selectApiBlocks()`: Implements the tiered selection algorithm with family-aware prioritization
  - `elideDeclaration()`: Abridges oversized declarations by dropping less-important members while preserving factory methods
  - `moduleFamiliesOf()`: Extracts module families from the digest's own catalog with fallback for older kits

- **Increased default digest size limits**:
  - `DEFAULT_KIT_DIGEST_MAX_BYTES`: 100 KiB → 150 KiB (games repo digest was already at 99.5% of old limit)
  - New `DEFAULT_MCP_DIGEST_MAX_BYTES`: 90 KiB for MCP lane

- **MCP surface changes**: Kit browse tools (`list_kit_files`, `search_kit_files`, `read_kit_file`, etc.) and the new `get_kit_api` are now advertised to models, replacing the previous hidden-tool approach. Example and proposal tools remain unadvertised.

- **Updated workflow guidance**: Clarified that the platform and kit are not on the public web, directing agents to use `get_kit_api` or browse tools instead of web searches.

- The selection algorithm uses a "reserve" strategy: oversized core declarations get a dedicated budget slice to prevent them from being starved by earlier passes
- Module factories are ranked above other members when abridging, ensuring `createParty`, `createZone`, `createCommons`, and `createPresence` survive even tight budgets
- Omission notes are included in the output so agents can see what was dropped and know to use the full kit if needed
- The `Engine modules` catalog section is now placed first in the compacted digest, answering "what can this platform build" before the API details

https://claude.ai/code/session_0155jrUy76vVwwcdPLm2Vh3u

### DIFF
--- a/.claude/skills/byoca-mcp/SKILL.md
+++ b/.claude/skills/byoca-mcp/SKILL.md
@@ -108,14 +108,45 @@ declaration file into whole top-level blocks (`splitDeclarationBlocks`), gives e
 module family a block before any family gets a second one (`selectApiBlocks`), and abridges
 a block too large to fit whole (`GameKitApi`, ~44% of the entire surface — this is where
 `createParty` / `createZone` / `createCommons` / `createPresence` all live) member-wise
-rather than dropping it, ranking module factories first. Whatever a budget still can't fit
-is named in an `// Omitted for length (...)` note instead of vanishing. The digest also
-gained a `## Engine modules` catalog (`digestEngineModules` in the games repo's
-`tools/lib/pack-kit.ts`), generated from each `shared/modules/*.ts` file's own header
-comment — so a module is listed the moment it exists and the catalog cannot drift.
+rather than dropping it, ranking module factories first (`elideDeclaration`). Whatever a
+budget still can't fit is named in an `// Omitted for length (...)` note instead of
+vanishing. The digest also gained a `## Engine modules` catalog (`digestEngineModules` in
+the games repo's `tools/lib/pack-kit.ts`), generated from `GAME_KIT_MODULES` +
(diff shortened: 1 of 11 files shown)


gtanczyk:
@codex

PR #741, opened 2026-08-10
Run the games catalog's full behavioral gate here (free public Actions)

gtanczyk:
## Summary

- `gamedevpl/www.gamedev.pl-games`#685 (merged) capped its billed `validate` job's full-mode gate at 3 minutes, dropping the catalog-wide behavioral sweep (trace, check:webgl, accept, `agency --all --strict`, agent-play, playtest) from that path — it was routinely landing at 24-45 minutes on a single billed runner.
- That sweep needs to keep running somewhere. This repo is public, so its Actions minutes are free — adds `games-catalog-gate.yml` to run it here instead, against a checkout of the games repo pulled in with the existing `GAMES_REPO_TOKEN` (same PAT the snapshot bake and `games_repo_contract` CI job already use).
- `playtest` is upgraded from `--suite default` (10 games) to `--all` (full catalog) now that it isn't blocking a billed merge.

## Trigger

- `repository_dispatch` (`games-catalog-verify`) fired from the games repo's `publish` job on every push/schedule to its `main` — already merged and live, so once this merges the sweep runs roughly per-push instead of the old weekly-only schedule.
- A daily `schedule` here as a fallback for a missed/failed dispatch, mirroring `publish-games.yml`'s existing pattern.
- `workflow_dispatch` with a `ref` input as a manual escape hatch.

## Notes

- Best effort in one direction only: nothing here can block or fail a push in the games repo — the dispatch step there already tolerates this failing.
- `GAMES_REPO_TOKEN` is contents:read only, so a red run here has no way to post back to the games repo (no issue/status). It's only visible in this repo's own Actions tab. Flagging in case cross-repo failure notification is wanted later — that would need the token's scope widened.

https://claude.ai/code/session_01Am7adgAyGpWpE6dqPUmuZx

---
_Generated by [Claude Code](https://claude.ai/code/session_01Am7adgAyGpWpE6dqPUmuZx)_

### DIFF
--- a/.github/workflows/games-catalog-gate.yml
+++ b/.github/workflows/games-catalog-gate.yml
@@ -1,0 +1,111 @@
+name: Games catalog gate
+
+# The full-catalog behavioral sweep for gamedevpl/www.gamedev.pl-games — trace,
+# check:webgl, accept, agency --all --strict, agent-play, and playtest --all
+# across every game. This used to run inline in that repo's `validate` workflow
+# on every push/schedule to its main branch and was routinely landing at
+# 24-45 minutes on a single billed runner (private repo => billed Actions
+# minutes). This repo is public, so its Actions minutes are free — the sweep
+# runs here instead, against a checkout of the games repo pulled in with
+# GAMES_REPO_TOKEN (same contents:read PAT the snapshot bake and the
+# games-repo contract check already use).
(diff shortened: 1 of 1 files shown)


gtanczyk:
@codex

PR #742, opened 2026-08-10
listings: document the tools, and what a client is not offered

gtanczyk:
GitHub approved `pl.gamedev/creator` for the MCP Registry and suggested one improvement: a section naming the tools the server exposes, so developers can judge it before connecting. This is that.

## What's here

A table in `listings/mcp/README.md` — 30 tools, grouped by where they fall in a build round, with each tool's own `readOnlyHint` / `destructiveHint` as the third column rather than a judgement written by hand.

I generated it from what `tools/list` actually returns against production, not from the source map. Those differ, and that turned out to be the interesting part.

## Nine tools are registered and will never appear

- `get_round_status` and `get_round_media` are offered only to a client that negotiated the UI extension — a client with no views would hand them to its model, which `visibility:["app"]` exists to forbid.
- The proposal tools (`open_proposal_round`, `submit_proposal`, `get_proposal_status`) and the exemplar tools (`list_examples`, `get_example`, `list_example_files`, `read_example_file`) are in `MCP_UNADVERTISED_TOOLS`: *"Callable for REST clients, never advertised, never named to a model."*

A list that silently omitted them would read as complete, and a reader would have no way to discover the difference until something they wrote never fired. So the section names them and says why they're absent.

That is not hypothetical. I recommended `open_proposal_round` as the tool for Test Case 4 of the ChatGPT plugin submission, which is already filed. A model cannot see that tool, so it cannot call it — the test will not do what its expected-output text says. My error, from reading the tool map instead of the visibility list. Documenting the split is the fix that stops it recurring.

## Pinned, not trusted

`MCP_VISIBLE_TOOLS` is now exported, and `plugin-manifests.test.ts` asserts:

- the table equals `MCP_VISIBLE_TOOLS` minus the app-only pair,
- no `MCP_UNADVERTISED_TOOLS` entry is listed as available,
- both app-only tools are still *mentioned*, so they can't be dropped silently.

A tools doc drifts the moment someone adds a tool, which is exactly when nobody rereads it. Verified by deleting a row and watching the assertion fail.

Also confirmed the table matches the live server exactly: 30 documented, 30 returned by `tools/list`, sets equal.

## Notes

- `npm run type-check` fails on `src/refine.ts` (`Property 'thinking' does not exist on type 'RequestBuilder'`) on master with these changes stashed — pre-existing, local `node_modules` drift against the lockfile, untouched by this PR.
- Comment prose kept under the file's existing baseline; no `--force`, no reseal.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GjXeZdWfyTSwUdr4S9Lcit)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,29 +46,21 @@ The interesting engineering is written up in [`docs/`](./docs) — start with
 ## Connect your agent (MCP)
 
 The server is remote and needs no install: point an MCP-capable client at
-`https://www.gamedev.pl/api/mcp` and sign in with OAuth, or add it as a plugin with
-`/plugin marketplace add gamedevpl/www.gamedev.pl`.
+`https://www.gamedev.pl/api/mcp` and sign in with OAuth. Claude users can install it as a
+plugin instead — [`listings/mcp/claude-plugin/README.md`](./listings/mcp/claude-plugin/README.md)
+has the exact sequence, including the connector approval that installing does not do on its
+own.
 
-Thirty tools are advertised, and they follow the shape of one build round:
+Thirty tools cover one build round: opening or rejoining it, reading the brief and the
+engine kit, staging and delivering source files, checking the quality gate, and exchanging
+progress, screenshots and messages with the creator. Nine more exist and are deliberately
+never advertised, so an agent will not discover them.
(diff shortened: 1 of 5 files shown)


gtanczyk:
@codex

gtanczyk:
@codex

PR #745, opened 2026-08-10
fix(seed): type-check generated GameKit calls

gtanczyk:
## Summary

- Keep `shared/game-kit.d.ts` available to the seed validator without rendering it into model prompts.
- Run the esbuild bundle check and in-process TypeScript check together.
- Combine bundle and type diagnostics into one bounded repair round.
- Preserve `compiles` as the esbuild-only preview flag and expose `typeChecked`/`typeErrors` for observability.
- Fail open when the declaration is absent, the checker errors, or the post-check budget is exceeded.

## Why

The seed pipeline previously treated transpilation as sufficient validation, so incorrect GameKit calls could reach the generated preview. The checker now validates against the real GameKit declaration while keeping prompt context and seed behavior backend-agnostic.

## Validation

Run locally before publishing:

- `npm run type-check`
- `npm run lint`
- `npm run test`
- `npm run build`
- Focused seed/context tests


gtanczyk:
@codex

PR #756, opened 2026-08-11
Fix overlapping Studio chat actions

gtanczyk:
## Summary

- keep the Studio chat pop-out and close controls in one explicit flex action row
- add consistent sizing, spacing, hover/focus treatment, and localized tooltips
- add component and stylesheet regression coverage

## Root cause

The chat close button reused the generic `.modal-close-btn` class, which positions controls absolutely. The neighboring pop-out link stayed in normal flex flow, so both controls tried to occupy the same top-right space.

## Impact

The two actions are now distinct and legible, including with long Polish titles, compact phone layouts, keyboard-height viewports, and bottom update banners. Both controls expose localized accessible names and visible hover/keyboard-focus tooltips.

Existing navigation and close behavior is unchanged. None of the product instrumentation questions are affected, so no telemetry event was added.

## Validation

- `npm run type-check`
- `npm run lint`
- `npm exec --workspace @gamedevpl/web -- vitest run src/StudioChatRail.test.tsx src/styles.studioChatRail.test.ts`
- `npm run build`
- browser checks at desktop, phone, and keyboard-height viewports, including long Polish copy and an update banner

### DIFF
--- a/apps/web/src/StudioChatRail.test.tsx
+++ b/apps/web/src/StudioChatRail.test.tsx
@@ -1,0 +1,53 @@
+// @vitest-environment jsdom
+
+import { act } from 'react';
+import { createRoot } from 'react-dom/client';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import i18n from './i18n/index.js';
+import { StudioChatRail } from './StudioChatRail.js';
+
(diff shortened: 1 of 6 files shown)


gtanczyk:
@codex

PR #762, opened 2026-08-11
Code surface: GameKit-aware autocomplete, hover, and advisory diagnostics (GA-02..GA-08)

gtanczyk:
## Summary

Adds a worker-backed TypeScript language service to the Code surface's CodeMirror editor, giving creators real completions, hover info, and live advisory diagnostics against their game's actual types and the pinned GameKit declaration — not just syntax highlighting.

Full design/rationale lives in `creator-code-gamekit-autocomplete-plan.md` in the internal ops repo (GA-01..GA-08; GA-01 shipped earlier in #752).

- **`tsWorker.ts`** — a Web Worker running a `@typescript/vfs` TypeScript environment. TypeScript's `lib.*.d.ts` files are bundled as lazy per-file chunks via Vite (`import.meta.glob`), so nothing is fetched from a CDN and only the handful of lib files actually needed ever load.
- **Lockstep test** — `apps/api/src/type-check-web-lockstep.test.ts` deep-equals the browser's `COMPILER_OPTIONS` against the server's authoritative `type-check.ts`, so the editor can never suggest a member the server's typecheck gate would then reject. Both packages pin `typescript@5.9.3` exactly.
- **`CodeSurface.tsx`** — owns the worker's lifecycle: creates it once the surface is editable, seeds it from the fetched sources plus the kit declaration (GA-01's route), and keeps sibling files in sync on autosave.
- **`CodeMirrorEditor.tsx`** — wires `tsSync`/`tsAutocomplete`/`tsHover` through a CodeMirror `Compartment` once the worker is ready, reconfigured live rather than remounting (preserves cursor/focus if the worker finishes loading mid-keystroke). A custom hover renderer surfaces a kit type's JSDoc doc comment, not just its signature.
- **Advisory diagnostics** — the worker's own live TypeScript diagnostics render as amber warnings via `tsLinter()`, visually distinct from the server typecheck's authoritative red errors. Nothing the worker produces ever gates or blocks — it's advisory only.
- **Degradation** — a failed worker chunk load, a failed kit-declaration fetch, or a read-only (agent-round) surface all fall back to plain CodeMirror editing rather than a blocked panel; covered by `CodeSurface.languageService.test.tsx`.

### Bugs found and fixed by real-Chromium verification

The jsdom suite stayed green through all of these — they were only visible driving an actual worker in an actual browser:

1. A `sources`-by-object-reference effect dependency in `CodeSurface.tsx` could race its own cleanup and silently prevent the worker from ever starting (no error, no console output). Fixed by keying the effect on `editable`/`slug` instead.
2. `@typescript/vfs` roots its virtual filesystem at `/` — lib files and game/kit files seeded without the leading slash silently failed to resolve (`Math` itself read as unknown). Fixed with one `toVfsPath()` choke point.
3. The original "remount when the worker becomes ready" approach could blur the editor mid-keystroke. Replaced with a live `Compartment` reconfigure.
4. `comlink` was a static import reachable from the eager (non-lazy) part of the bundle, leaking its bytes into the main chunk for every visitor. Now dynamically imported, its own chunk.

## Test plan

- [x] `apps/web` — 149 test files / 1251 tests passing
- [x] `apps/api` — 195 test files / 3071 tests passing (includes the new GA-03 lockstep test)
- [x] `tsc --noEmit` clean in both `apps/web` and `apps/api`
- [x] `eslint` clean on all changed/new files
- [x] Production `vite build` succeeds; verified the worker/typescript/vfs/comlink/codemirror-ts weight lands in lazy chunks, not the main bundle
- [x] Verified against a real headless Chromium instance (local API + kit fixture): `ctx.gfx.` completion lists the kit's exact members (`fillRect`, `fillStyle`, `restore`, `rotate`, `save`, `translate`); hovering the `ctx` parameter and its type both show tooltips, including the kit interface's own JSDoc doc comment; the worker's advisory diagnostics render as amber warnings alongside the server's red ones

https://claude.ai/code/session_01Y2WrkNc5Ww3LAMVMJjBT9a

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y2WrkNc5Ww3LAMVMJjBT9a)_


gtanczyk:
@codex

PR #770, opened 2026-08-11
fix(api): commit builder switch only after dispatch/resume succeeds

gtanczyk:
## Summary
- `resumeBuild` wrote `record.builder` (via `store.setRoundBuilder`) *before* calling the target backend's `dispatch`/`resume`. When that call throws, the job is left durably marked with a builder it never actually reached, while `dispatch.refs` still ends in the previous builder's ref.
- Every later status-poll/reconcile then resolves the wrong backend from `record.builder`, calls `observe()` with the stale ref, gets a vendor error, and silently swallows it (`reconcileNativeJob`'s catch — best-effort by design) — so nothing ever corrects the record.
- Client-side, a job stuck reading `builder: 'platform'` with no active platform work hides `BuilderModeBadge` entirely (`canChooseBuilder`/`isAgentWorkActive`/`showBuilderBadge` in `SubmissionStatusView.tsx`), so the creator has no visible way to switch to BYOCA/self-build.

Root-caused against a real production incident: job 1000053 (`mistrz-trampoliny`) hit exactly this — a 2026-08-10 self→platform handoff attempt failed with `anthropic managed agents ... 400 Invalid session ID: self:1000053`, and the round has been stuck reading `builder: platform` ever since, even though the creator's own BYOCA/MCP sessions kept building the game underneath via the unaffected self-round-credential path.

## Fix
Move the `store.setRoundBuilder` commit to after a successful `dispatch`/`resume` result, so a failed vendor call leaves the record's `builder` field untouched.

## Test plan
- [x] Added a regression test (`does not commit the requested builder when its dispatch fails`) that dispatches against a backend stub whose `dispatch`/`resume` always throw, and asserts `record.builder` and `dispatch.refs` are unchanged afterward.
- [x] Verified the new test fails against the pre-fix code (reproduces the exact `builder: 'platform'` wedge) and passes after the fix.
- [x] `npx vitest run apps/api/src/submissions.test.ts` — 182/182 passing.
- [x] `node eslint-rules/comment-prose-check.mjs apps/api/src/submissions.ts` — within baseline.

Job 1000053 itself still needs a one-off Firestore repair (outside this PR's scope) to clear its stale `builder`/`builderHandoff` fields — tracked separately.

### DIFF
--- a/.claude/skills/verify-agent-work/SKILL.md
+++ b/.claude/skills/verify-agent-work/SKILL.md
@@ -179,6 +179,12 @@ Two concrete instances of that (observed 2026-07-23):
   in a second worktree and run the same file. Pair it with a control case that must
   pass on both, so a green-everywhere result reads as a broken probe rather than a
   clean bill of health.
+- **A rollback-after-dispatch fix can still fail before the provider returns.** Observed
+  (#770, 2026-08-12): managed providers can launch an MCP session before the awaited
+  `dispatch()` resolves. Moving a builder write after that call fixed vendor rejection
+  recovery but left the session's first tool call reading the old builder. When an
+  awaited vendor call can call back into the app, publish routing state before the call,
+  roll it back only when that call rejects, and have the stub read the store in-flight.
 - **A PR that changes an error's SHAPE has changed agent behaviour even when no test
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #772, opened 2026-08-12
Fix wasted Code header space

gtanczyk:
## Summary

- compact the desktop Code header while preserving the global escape action
- keep the global escape on its own row on narrow phones
- replace the mobile scrolling filename rail and native select with a compact current-file disclosure
- open mobile file navigation as a focused bottom sheet with touch-sized rows, selected-file state, locked directories, backdrop dismissal, and Escape support
- preserve the desktop file rail and existing file-open telemetry

## Measurement / telemetry

- no new event or PII
- existing `code_step: file_opened` remains the signal for both desktop rail selection and mobile sheet selection
- the admin coding funnel remains comparable before and after the UI change

## Validation

- focused Code surface + CSS tests: 22 passed
- web type-check, lint, and production build passed
- visual checks at 320px and 390px
- all web tests passed (151 files, 1,292 tests); package and rules tests passed
- 22 unrelated API/world websocket tests fail under the sandbox with `listen EPERM`

### DIFF
--- a/apps/web/src/CodeSurface.test.tsx
+++ b/apps/web/src/CodeSurface.test.tsx
@@ -137,7 +137,7 @@ describe('CodeSurface', () => {
     expect(textarea.value).toContain('export const boot');
   });
 
-  it('uses a native file picker instead of a scrolling mobile rail', async () => {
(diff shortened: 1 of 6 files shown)


gtanczyk:
@codex

PR #779, opened 2026-08-12
refactor: extract shared seed-branch staging for both Copilot implementations

gtanczyk:
## Summary

`copilot-backend.ts`'s `stageSeed` and `managed-provider-copilot.ts`'s `stageSeed` independently implemented the same "delete any stale seed branch, then commit the draft to a fresh disposable branch" logic — same seed-branch naming (`seed/job-<issueNumber>`), same fail-open-to-null behavior on any staging error. Two implementations of that logic means a GitHub-behavior surprise (`agent-tasks.ts` already documents several places where GitHub's API contradicts its own docs) gets found and fixed once, or silently only once, instead of both places getting the fix.

- Extracts `stageAndCleanupSeedBranch` and `seedBranchName` into `agent-tasks.ts`, since both callers already depend on it and it's where the equivalent GitHub-quirk documentation for this dispatch surface already lives (checked for a better home first — no other shared module fit as well).
- Both callers now delegate to it, still injecting their own `github` client. `copilot-backend.ts` also passes its own failure-logging callback via a new `onStageError` hook, preserving the one real behavioral difference between the two callers: the direct backend logs a staging failure, the managed provider does not.
- `copilot-backend.ts` re-exports `seedBranchName` from `agent-tasks.ts` to keep its existing public export working.

## What did not change

**No dispatch-visible behavior changed:** same seed-branch naming, same fail-open semantics, same stale-branch-delete-before-create ordering. `copilot-backend.test.ts` and `managed-provider-copilot.test.ts` both pass **unmodified** — they mock the `github` client directly rather than the extracted helper, so they exercise the new code path with zero test changes needed. That's the evidence this refactor didn't touch behavior: the assertions are unchanged, only the implementation underneath moved.

**The trivial post-round branch deletion was left alone.** The brief that prompted this cleanup also flagged `copilot-backend.ts`'s `cleanup` and `managed-provider-copilot.ts`'s `deleteWorkspace` as duplicated. After reading both, they're each a single delegated call to the shared `GitHubClient.deleteBranch` (one has an `if (!workspace) return` guard the other doesn't need) — there's no logic there to duplicate a bug in. Wrapping that one-line delegation in a "shared helper" would add an abstraction with nothing to test, which the repo's own conventions argue against, so I left it as-is rather than force an extraction for its own sake. Flagging this explicitly per the task brief's request to state reasoning rather than silently deciding.

## Out of scope

Per the cleanup brief this PR is drawn from: this does not retire, delete, or feature-flag `copilot-backend.ts` (the games-repo dedup work is a separate, gated decision — internal task "MP-04"), and does not change which backend any submission routes to. `managed-backend.ts` and `managed-agent.ts` are untouched.

## Test plan

- [x] `npm run lint` (including the comment-prose check) — green
- [x] `npm run type-check` — green
- [x] `npm run test` — green: `copilot-backend.test.ts` (29 tests) and `managed-provider-copilot.test.ts` (8 tests) both pass unmodified, plus the full monorepo suite (3122 API-workspace tests, 1288 web tests, 20 world tests, 36 eslint-rule tests — all green)
- [x] `npm run build` — green

---
_Generated by [Claude Code](https://claude.ai/code/session_01PAsj5Z8XRHPmrrBsQzzs84)_

### DIFF
--- a/apps/api/src/agent-tasks.test.ts
+++ b/apps/api/src/agent-tasks.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import {
   createAgentTasksClient,
   creditsFromUsageAmount,
@@ -6,6 +6,8 @@ import {
   normalizeModel,
   parseAgentTask,
   resolveTaskBranch,
+  seedBranchName,
+  stageAndCleanupSeedBranch,
 } from './agent-tasks.js';
 
 /** A settled task, shaped exactly as the live API answered on 2026-07-29. */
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex


---

gtanczyk:
Good suggestion — `stageAndCleanupSeedBranch` was only exercised indirectly through the two callers' tests. Pushed direct coverage in `agent-tasks.test.ts`: branch naming, the delete-before-create ordering, a stale-branch delete failure not blocking staging (the common case on a job's first dispatch), and the fail-open-to-null behavior with and without `onStageError`. (Your own commit `aa4a553` didn't reach the branch since this checkout had no way to push it — I wrote equivalent coverage from scratch.) Full gate green: lint, type-check, and the affected test files (61 tests).


---

PR #776, opened 2026-08-12
docs: fix agent-adapters.md's stale PR-merge delivery model

gtanczyk:
## Summary

`docs/agent-adapters.md`'s header, common contract, and Hosted/Interactive-agent sections described PR-merge as the delivery mechanism for a platform game build. That stopped being true when Copilot dispatch moved to GitHub's Agent Tasks API and delivery moved to the build channel (`submit_sources`) — see `copilot-backend.ts`'s header comment and `docs/build-brief.md`, which are current and were used as the reference for what's true today.

- Rewrites the doc's "Current model" header and common contract to describe the real flow: gamedev.pl dispatches a coding agent for one round and reads the delivery back over the build channel or a managed backend's own harvest — never a merged pull request.
- Cross-references `build-brief.md` (delivery-contract table) and `managed-agent-backend.md` instead of duplicating their detail.
- Updates the **Hosted agents** section to describe Copilot's actual dispatch path (`copilot-backend.ts` / `managed-provider-copilot.ts`, Agent Tasks API, harness-not-target repo, build-channel delivery) instead of "assign an issue, receive a PR."
- Clarifies **Interactive/local agents** covers ordinary repository-tooling PRs on the games repo itself, distinct from a platform build round.
- Fixes **Task routing**'s "structured issue" language — no game round creates a GitHub issue in production code (`createIssue` is defined but never called from a build-dispatch path), so routing is described as round dispatch rather than issue assignment.
- Keeps everything still true: the common contract (SPEC.md as source of truth, untrusted spec, one-game-directory scope, no runtime network), seeded-workspace behavior, and the "Retired adapter model" history.

This is a documentation-only change. Nothing about dispatch, routing, or which backend is used in production is touched.

## Out of scope

Per the cleanup brief this PR is drawn from: this does **not** retire, delete, or feature-flag `copilot-backend.ts` (a separate, gated decision — internal task "MP-04" — with an explicit owner-approval requirement), and does not change which backend any submission routes to.

## Test plan

- [x] `npm run lint` (includes the comment-prose check) — green
- [ ] `npm run type-check` — not applicable, docs-only change, no TypeScript touched
- [ ] `npm run test` — not applicable, docs-only change
- [ ] `npm run build` — not applicable, docs-only change

---
_Generated by [Claude Code](https://claude.ai/code/session_01PAsj5Z8XRHPmrrBsQzzs84)_

PR #788, opened 2026-08-12
MP-04: retire copilot-backend.ts, the direct Arch A builder

gtanczyk:
## Summary

Per the owner's direction: Arch B (BYOCA/MCP) and Arch C (managed agents on Anthropic/Gemini/Copilot) are the production paths now, and Arch A — the direct `copilot-backend.ts` dispatching through the Agent Tasks API and staging/deleting git branches — has no live traffic left to serve. This executes MP-04 from the ops-repo migration plan (`docs/managed-driver-expansion-plan.md`): retire the legacy backend.

**Stacked on #779.** This branch builds on `claude/arch-a-cleanup-src6jj-task3-dedup` (base of this PR) because that PR already moved the seed-staging/cleanup logic both callers shared into `agent-tasks.ts`. Landing that first means this deletion is a clean subtraction — `managed-provider-copilot.ts` was already the sole forward-looking caller of the shared helper, so nothing needs re-porting. **Please merge #779 before this one.**

## What this does

- Deletes `copilot-backend.ts`.
- `agent-backend-env.ts`: removes `createPlatformBackendFromEnv` and the already-`@deprecated` `createAgentBackendFromEnv` wrapper (verified zero remaining callers of either). `createAgentBackendRegistryFromEnv` no longer falls back to the legacy backend when `MANAGED_AGENT_VENDOR` is unset or invalid — an unconfigured environment now has no platform builder at all, rather than a silent Copilot dispatch nobody configured.
- `agent-backend-env.test.ts`: the one test asserting the old fallback (`'falls back to Copilot only when no managed vendor was selected'`) now asserts its absence instead.

### Test relocation, not blanket deletion

`copilot-backend.test.ts` was 29 tests, and not all of them tested the backend:

| Tests | What they actually exercise | Where they went |
| --- | --- | --- |
| 11 (`buildPrompt` describe) + 1 (seeded-dispatch's prompt assertion) | `build-prompt.ts`'s default (harness-lane) contract | Moved verbatim into `build-prompt.test.ts` |
| 2 (`dispatch through the submissions route`) | `store.ts`'s `recordDispatch`/`getSubmission`, generic to any backend name | Moved verbatim into `store.test.ts` |
| 16 (dispatch / resume / cleanup / observe+cancel / seeded dispatch against the backend) | The backend itself | Deleted with the backend |

The 16 deleted tests are not silently assumed covered — `managed-provider-copilot.test.ts` exercises the equivalent behavior for the surviving Copilot driver (pinned model + no PR, branch-from-task reporting, credit summation, seed staging on the same branch naming, fail-open on seed failure, cancellation honesty, workspace deletion), but two things don't have an exact equivalent and I'm flagging rather than pretending parity:
- `copilot-backend.ts`'s `resume()` carried its own doc comment on why a revision starts a fresh branch instead of continuing the stale one. `managed-provider-copilot.ts` has no `resume()` of its own — that semantics lives in `managed-backend.ts`, untouched by this PR.
- The `cleanup()` guard ("does nothing for a dispatch that never got a branch") doesn't map onto `deleteWorkspace(workspace: string)`'s required-string signature the same way.

Also fixes one now-stale reference in `docs/build-brief.md` naming `copilot-backend.test.ts` as one of two suites pinning the untrusted-spec contract — that job belongs to `build-prompt.test.ts` alone now.

## What this does NOT do

Per MP-04's own brief in the ops repo, this PR is deliberately narrower than "MP-04, done":

- **The availability switch is untouched.** `managed-availability.ts` still only turns the platform builder on/off; it doesn't yet select or fail over between drivers from the ops-config document without a redeploy, which MP-04's brief calls for ("the kill switch's implicit fallback for a disabled managed lane is the legacy Copilot backend — which this task removes"). That fallback is gone now, but the driver-selection dimension the brief also asks for is not built here. Flagging so it isn't mistaken for already covered.
- No change to `managed-backend.ts`, `managed-agent.ts`, or any `managed-provider-*.ts` file's behavior.
- No burn-in period was observed (MP-04's step 3) — per the direction this PR is drawn from, Arch A already has no live traffic, so there's nothing to burn in against.

## Test plan

- [x] `npm run lint` (incl. comment-prose) — green
- [x] `npm run type-check` — green
- [x] `npm run test` — green: 12+96+3112+1288+20+36 = 4564 tests across the monorepo. API workspace went from 3127 (after #779's coverage commit) to 3112 — accounted for exactly by relocating 14 tests and deleting 15 net (29 removed, 14 relocated).
- [x] `npm run build` — green

---
_Generated by [Claude Code](https://claude.ai/code/session_01PAsj5Z8XRHPmrrBsQzzs84)_


gtanczyk:
@codex


---

gtanczyk:
Confirmed — `createManagedPlatformBackendFromEnv` returns `undefined` immediately unless `MANAGED_AGENT_VENDOR` is set, so `AGENT_TASKS_TOKEN`'s presence alone no longer enables dispatch the way `createPlatformBackendFromEnv`'s unconditional fallback used to. Pushed fixes:

- `infra/deploy-api.sh`: deploy log no longer claims dispatch is enabled from the token alone; says it's enabled once `MANAGED_AGENT_VENDOR=copilot` is also set.
- `docs/runbooks/rotate-secrets.md`: fixed the dangling `createAgentBackendFromEnv` reference (deleted by this PR) to point at `createManagedPlatformBackendFromEnv`, noted the vendor requirement, and corrected the verification table's log-line name to the real `managed agent dispatch enabled` (the old `agent dispatch enabled` string no longer exists anywhere in the codebase — checked).

(Your commit `7f6ebc4` didn't reach the branch since this checkout had no way to push it — applied equivalent fixes directly.) `bash -n` clean, lint green.


---

gtanczyk:
@codex

gtanczyk:
Addressed both review passes:

**Codex (P2):** confirmed — `docs/managed-agent-backend.md`'s "How to exercise it" section said "Without a managed vendor, Copilot fills the platform slot only when `AGENT_TASKS_TOKEN` is configured," which is exactly the fallback this PR retires. Rewrote it: `MANAGED_AGENT_VENDOR=copilot` must be set explicitly now, `AGENT_TASKS_TOKEN` alone is no longer sufficient. The other reference you flagged (`docs/agent-adapters.md:49-51`) was already fixed by #776, which has since merged — this branch picked that up on rebase, nothing left to do there.

**GitHub Copilot review (3 comments):**
- `build-prompt.test.ts`: the untrusted-spec-fence test only covered `{kind:'channel'}` and `{kind:'outputs'}`, not the `fast` variant, while `build-brief.md` claims the fence holds for every delivery contract. Extended the test to cover `{kind:'channel', fast:true}` rather than soften the doc — verified it passes, since the fence is emitted before the delivery-specific branch runs.
- `store.test.ts`: a test named "is skipped without a backend..." only asserted a fresh submission has no dispatch record, true regardless of any backend. Renamed to match what it actually checks.
- `rotate-secrets.md`: a `gcloud run services update ...` command's second line had lost its list-item indentation, breaking the inline code span in rendered Markdown. Joined it onto one line.

Full gate green on the rebased branch: lint, type-check (all workspaces), and the five affected test files (100 tests).


---

PR #798, opened 2026-08-12
fix(api): route Copilot's MCP-lane rounds to a separate, content-free repo

gtanczyk:
## Summary
- Copilot's real MCP lane has never worked. Verified live against a real Copilot Agent Task on the games repo (job 1000056): zero `/api/mcp` traffic reached us, and the task committed straight to git instead of using any MCP tool.
- Root cause, both confirmed directly: (a) the repo's GitHub-side MCP connector was never configured (`mcpServers: {}`, no `COPILOT_MCP_CONNECTOR_SECRET`), and (b) `game-builder.agent.md` only ever taught the harness/build-channel workflow, directly contradicting the per-round MCP prompt (it dutifully followed its own standing instructions instead).
- Wiring the connector onto the real private games repo would also put a prompt-injectable session (creator `spec` text) in a checkout with real proprietary source, and — once a connector secret is added there — a real secret value, both readable via the agent's always-available bash tool and relayable to the creator via `report_progress`/`end` (unmoderated, ~1200 chars/call, no output-side filtering).
- Fix: dispatch MCP-lane Copilot rounds against a separate, minimal repo (`gamedevpl/scratchpad`) holding nothing but a dedicated MCP-only custom agent doc (`.github/agents/game-builder-mcp.agent.md`, already committed there). The MCP lane never needs local repo content — `get_brief`/`get_seed`/`get_kit`/`stage_source_file`/`submit_sources` are all HTTP, same as every other MCP-lane path (Gemini, Anthropic, BYOCA creators). Nothing there is worth exfiltrating even under a full prompt injection, and the connector secret's blast radius drops to near zero even if it leaked.
- GitHub's Agent Tasks API is scoped by repo in the URL, not session id, so the harness and MCP lanes now use separate `AgentTasksClient`/`GitHubClient` instances, with in-process session/workspace tracking routing `getSession`/`cancelSession`/`deleteWorkspace` to the right one.
- New env vars (`MANAGED_AGENT_COPILOT_MCP_REPO`/`_BASE_REF`/`_CUSTOM_AGENT`) are all optional — unset preserves today's harness-only behavior exactly, byte-for-byte.

## Remaining manual steps (not code, tracked separately)
- Grant `AGENT_TASKS_TOKEN` access to `gamedevpl/scratchpad`
- Create the `copilot-mcp-connector` secret (deploy-api.sh:196-199 / deploy.yml:94-96 already wire `COPILOT_MCP_CONNECTOR_SECRET` automatically once it exists — no further code change needed)
- Configure `gamedevpl/scratchpad`'s MCP JSON + Agents secret (`COPILOT_MCP_CONNECTOR_SECRET`) via its GitHub settings

## Test plan
- [x] `npx vitest run src/managed-provider-copilot.test.ts src/agent-backend-env.test.ts` — 19/19 pass, including new tests asserting MCP-lane rounds route to `mcpRepo` (not the games repo) and that harness-lane behavior is unchanged when `mcpRepo` is configured but unused
- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npm run lint` — clean on all touched files


gtanczyk:
@codex

PR #801, opened 2026-08-13
feat(admin): flip managed agent vendor at runtime from the Limits tab

gtanczyk:
## Summary
- `MANAGED_AGENT_VENDOR` now only picks the *default* vendor at boot — every configured vendor (anthropic/gemini/copilot) that has valid credentials builds its own backend independently, tracked as a `platformByVendor` map instead of a single `platform` slot.
- The Limits tab of the operator console gained a Vendor selector, writing `managedAgentVendorOverride` onto the existing `creationLimits` document. Clearing it (empty option) falls back to the deployed default.
- The override resolves through the same TTL-cached breaker gate as every other creation limit, so a flip takes effect within the existing cache delay — no redeploy.
- An override naming a vendor with no credentials in this environment falls back to the default (surfaced in the panel's status line) rather than breaking dispatch.
- Workspace cleanup (`releaseWorkspace`) now targets the vendor that actually built the workspace (`record.dispatch.backend`) instead of assuming a single platform backend, since a runtime flip can change which vendor `platformByVendor`'s default points at mid-flight.

## Test plan
- [x] `apps/api`: `store.test.ts`, `agent-backend-env.test.ts`, `managed-availability.test.ts`, `managed-provider-copilot.test.ts`, `admin.test.ts`, `submissions.test.ts` — 290/290 passing
- [x] `apps/web`: `CreationLimitsPanel.test.tsx` — 6/6 passing
- [x] `npm run lint` (eslint + comment-prose) clean repo-wide
- [x] `tsc --noEmit` clean in both `apps/api` and `apps/web` (excluding pre-existing, unrelated failures: two `.thinking` property errors in `chat-agent.ts`/`refine.ts`, and missing `comlink`/`@codemirror/*`/`@valtown/codemirror-ts` type declarations)
- [ ] Manual: flip the vendor in a deployed operator console and confirm a fresh dispatch picks it up without a redeploy (not done as part of this change — `MANAGED_AGENT_VENDOR` stays `anthropic` in prod; this PR only ships the capability)


gtanczyk:
@codex

PR #805, opened 2026-08-13
Replace disabled builder send control with STOP

gtanczyk:
## What changed

- Replace the disabled composer send arrow with an enabled `STOP` button while the Gamedev.pl builder is actively working.
- Route STOP through the existing cooperative handoff to the creator's BYOCA agent.
- Keep handoff controls in the composer instead of the transcript.
- Add English and Polish labels, mobile touch-target styling, and regression coverage.

## Why

A disabled send button gives no clear action while a build is running. STOP makes the intended workflow explicit: stop the current build, switch to BYOCA, then reprompt.

## Validation

- `npm run test`
- `npm run test --workspace apps/web -- SubmissionStatusView.test.ts styles.studioComposer.test.ts`
- `npm run type-check`
- `npm run lint`
- `npm run build`
- Browser QA at desktop and 390px mobile widths


gtanczyk:
@codex

PR #810, opened 2026-08-13
Improve MCP channel: preview delivery warning, multi-patch, and inbox ack

gtanczyk:
## Summary of Changes

This PR incorporates developer experience and reliability improvements observed during live self-build MCP agent sessions:

1. **Fix `must_deliver` False Warning after Preview Delivery (`agent-channel.ts`)**:
   - `control.mustDeliver` and `control.delivered` now check `record.deliveredVersion || record.previewVersion` rather than only `deliveredVersion`.
   - Prevents the channel from continuing to send false "Nothing has been delivered yet" soft warnings after a successful preview submission (`mode=preview`).

2. **Support Batch Multi-Replace in `patch_source_file` (`source-patch.ts`, `agent-channel.ts`, `mcp-server.ts`)**:
   - Added `patches: [{ old, new }, ...]` parameter to `patch_source_file` and the `/api/agent/build/sources/stage/patch` route.
   - Allows agents to apply multiple exact string substitutions to a single file in a single tool call.
   - Tool description notes that replacements apply sequentially, so an earlier pair must not make a later `old` snippet ambiguous.

3. **Multi-file patching (`files[]`)**:
   - `patch_source_file({ files: [{ path, old, new }, { path, patches: [...] }] })` edits several files in one call (up to 100 files, 50 fragments each).
   - **Best-effort apply:** every edit is tried. Ones that apply are staged. Misses are listed in `failed[]` (`path` + `index`) with `warnings.code=patch_incomplete`. The agent retries only the misses — it must not resend the ones that landed.
   - If a file write is refused after several fragments applied in memory, `failed[]` lists every applied index for that file (not just `0`).
   - If nothing in the batch applied, the call is still 400 and `failed[]` lists every miss.
   - Single-file `old`/`new` and unified diffs are unchanged.

4. **Creator Inbox Ack on `end` (`agent-channel.ts`, `mcp-server.ts`)**:
   - Added optional `ackInboxIds: string[]` to `end` tool and endpoint.
   - Allows agents to acknowledge handled creator inbox messages concurrently when ending the round instead of requiring an extra dedicated turn for `ack_inbox`.
   - **Review fix:** acks run only after the `end` is accepted (not on `stopped` or a handoff that did not start). Store failures surface instead of being swallowed.

5. **Environment Clarity in Prompt for Fast MCP Lane (`build-prompt.ts`)**:
   - Clarified in revision instructions for fast-lane MCP sessions that the execution environment is an MCP-only sandbox (instructing models to use `start` then `get_sources` and avoid running bash exploration commands).
   - **Review fix:** if `get_sources` reports nothing delivered yet, the earlier round never finished and the agent is starting the game rather than revising it.

6. **BYOCA playbook** (`.claude/skills/byoca-mcp/SKILL.md`) documents `patches[]`, `files[]`, `end({ ackInboxIds })`, and `patch_incomplete`.

## Test Coverage
- Unit tests added/updated in `source-patch.test.ts`, `agent-channel.test.ts`, `mcp-server.test.ts`, and `build-prompt.test.ts`.
- Inbox ack is covered for: successful `end`, successful builder-handoff ack, rejected `stopped` (no ack), and rejected `handoff_already_acknowledged` (no ack).
- Multi-file patch is covered for: successful multi-file + multi-fragment apply, keeping successful files when a later one misses, keeping earlier `patches[]` when a later fragment misses, 400 + `failed[]` when nothing applied, and reporting every applied index when a file write is refused.

<div><a href="https://cursor.com/agents/bc-d3a07f47-3d62-4fbe-be89-0eb6a4bb8c61?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d3a07f47-3d62-4fbe-be89-0eb6a4bb8c61&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/api/src/agent-channel.ts
+++ b/apps/api/src/agent-channel.ts
@@ -70,8 +70,8 @@ import { BUILD_EVENT_KINDS, BUILD_STEPS, sanitizeCreatorText, type BuildEvent } 
 import { normalizeAtIntake, type IntakeText } from './localize-intake.js';
 import { createTranslatorFromEnv, type Translator } from './translate.js';
 
-// The build channel (docs/agent-live-channel-plan.md).
-// Direct route for progress, staging, submission, and status.
+// The build channel (docs/agent-live-channel-plan.md). Direct route for progress, staging, and status.
+// Invariant: agent input is untrusted, prompt-influenced text — sanitized, escaped on render, never model instructions.
 
 const MAX_EVENT_TEXT = 300;
 
@@ -2032,10 +2032,6 @@ export async function registerAgentChannelRoutes(
         return true;
       };
 
-      if (parsed.data.ackInboxIds && parsed.data.ackInboxIds.length > 0) {
-        await store!.markCreatorMessagesDelivered(issueNumber, parsed.data.ackInboxIds).catch(() => {});
-      }
-
       if (
         record.builderHandoff &&
         record.builderHandoff.awaitsAgentAck !== false &&
@@ -2055,6 +2051,9 @@ export async function registerAgentChannelRoutes(
             ...(await channelState(issueNumber, fresh)),
           });
         }
+        if (parsed.data.ackInboxIds && parsed.data.ackInboxIds.length > 0) {
+          await store!.markCreatorMessagesDelivered(issueNumber, parsed.data.ackInboxIds);
+        }
         const summarized = await recordSummary();
         const state = await channelState(issueNumber, fresh);
         return reply.send({
@@ -2069,6 +2068,10 @@ export async function registerAgentChannelRoutes(
 
       if (stopReason(record)) {
         return reply.send({ accepted: false, rejected: 'stopped', ...(await channelState(issueNumber, record)) });
+      }
+
+      if (parsed.data.ackInboxIds && parsed.data.ackInboxIds.length > 0) {
+        await store!.markCreatorMessagesDelivered(issueNumber, parsed.data.ackInboxIds);
       }
 
       // Submit-ended still records; a prior or legacy end does not.
--- a/apps/api/src/agent-channel.ts
+++ b/apps/api/src/agent-channel.ts
@@ -62,7 +62,7 @@ import { seedPayload } from './seed-status.js';
 import { largeSourceFileHint } from './module-size.js';
 import { gameManifestHint } from './game-manifest-hint.js';
 import { computeStageAdvisories } from './stage-hints.js';
-import { applyExactReplace, applyMultipleExactReplaces, applySourcePatch, SourcePatchError } from './source-patch.js';
+import { applyExactReplace, applySourcePatch, SourcePatchError } from './source-patch.js';
 import { overlayGameSources } from './staged-preview.js';
 import { SourceDeliveryValidationError, type SourceDeliveryService } from './source-delivery.js';
 import { type BuilderHandoff, type CreatorMessage, type Store, type SubmissionRecord } from './store.js';
@@ -322,7 +322,7 @@ const StageSourcePatchInputSchema = z
     old: z.string().max(200_000).optional(),
     new: z.string().max(200_000).optional(),
     patches: z.array(ExactPatchPairSchema).min(1).max(50).optional(),
-    files: z.array(FileExactPatchSchema).min(1).max(20).optional(),
+    files: z.array(FileExactPatchSchema).min(1).max(100).optional(),
     slug: z
       .string()
       .trim()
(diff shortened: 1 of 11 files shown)


gtanczyk:
@codex

apps/api/src/agent-channel.ts
```diff
@@ -2034,6 +2032,10 @@ export async function registerAgentChannelRoutes(
         return true;
       };
 
+      if (parsed.data.ackInboxIds && parsed.data.ackInboxIds.length > 0) {
+        await store!.markCreatorMessagesDelivered(issueNumber, parsed.data.ackInboxIds).catch(() => {});
```
gtanczyk: **Confirmed bug: a rejected `end` still acks the inbox, silently dropping the creator's requests.**

This runs at the top of the handler, ahead of both early-return paths — the `builderHandoff` not-started rejection below, and the `stopReason(record)` rejection after it. So an `end` that is *refused* has still marked the creator's queued change requests as delivered. Nothing re-queues them, and no later round will be handed them.

Verified against the branch with a throwaway test modelled on the existing `drops the summary when the round is already stopped, like a progress report` case:

```ts
const msg = await store.appendCreatorMessage(ISSUE, 'please make the ship faster');
await store.setSubmissionAbandoned(ISSUE, new Date().toISOString());
// ... POST /api/agent/build/end with { ackInboxIds: [msg.id] }
expect(end.json()).toMatchObject({ accepted: false, rejected: 'stopped' });
expect(await store.listPendingCreatorMessages(ISSUE)).toHaveLength(1);
//                                                    ^ got 0, expected 1
```

That existing test encodes the rule this breaks: a rejected `end` must not have side effects. Suggest moving the ack down next to `markAgentEnded(issueNumber)` and into the accepted-handoff branch, gated the same way `recordSummary` is.

Two smaller things on the same line:

- `.catch(() => {})` swallows store failures silently, which puts this out of step with the dedicated `/api/agent/build/ack` route above — that one lets the error surface. Worth dropping the catch, or at least logging it.
- Nothing exercises `ackInboxIds` in either `agent-channel.test.ts` or `mcp-server.test.ts`. Of the three features in this PR it's the only one shipping untested, and it's the one with the bug.

---

apps/api/src/build-prompt.ts
```diff
@@ -80,6 +80,16 @@ export function buildPrompt(brief: BuildBrief, delivery: DeliveryContract = { ki
           '',
         ]
       : []),
+    ...(brief.feedback && !brief.undelivered && channel && fastLane
+      ? [
+          '## Before you change anything',
+          '',
+          'Fetch the version the creator actually played using the MCP tools below (`start` then `get_sources`).',
+          'Do not run bash exploration commands — this execution environment is an MCP-only sandbox with no local checkout.',
+          'Read the returned files, then make the creator’s changes on top of them.',
```
gtanczyk, line 89: The `!fastLane` / `fastLane` split is correct — the two branches are properly mutually exclusive, so no round gets both.

One line worth carrying over from the shell branch, though. That one ends with:

> If it reports nothing delivered yet, the earlier round never finished and you are starting the game rather than revising it.

The MCP branch drops it, so a fast-lane agent whose `get_sources` comes back empty has no guidance for that case — and it's arguably more likely to hit it, since it has no checkout to fall back on and reads a revision brief that assumes a prior delivery exists.

---

apps/api/src/source-patch.ts
```diff
@@ -285,5 +281,33 @@ export function applyExactReplace(input: ApplyExactReplaceInput): ApplySourcePat
   return { content, replacements: 1 };
 }
 
+export function applyMultipleExactReplaces(input: ApplyMultipleExactReplacesInput): ApplySourcePatchResult {
```
gtanczyk, line 284: Logic is fine and the whole thing stays atomic — it throws before anything is staged, so a mid-sequence failure can't leave a half-patched file in the buffer.

Worth a one-liner about the failure mode it creates, though: replacements apply sequentially against the running content, and `applyExactReplace` requires an exactly-once match. So an earlier pair can introduce text that makes a later pair's `old` ambiguous, and the agent gets `matched more than once` for a snippet that *was* unique in the file it read. The MCP tool description says "sequentially", which is the right hint, but that consequence isn't obvious from it.

---

apps/api/src/agent-channel.ts
```diff
@@ -62,32 +62,16 @@ import { seedPayload } from './seed-status.js';
...
- * requests and a control block, so an agent that reports progress gets its instructions
- * back for free. Note the limit: this makes a *working* agent responsive, but it cannot
- * wake a stopped one — nothing is polling between sessions. Creator feedback therefore
- * still goes out as a PR comment, which is both the durable record and the wake-up.
- *
- * Everything the agent sends is untrusted, prompt-influenced text: sanitized here,
- * escaped on render, and never fed back to any model as instructions.
- */
+// The build channel (docs/agent-live-channel-plan.md).
+// Direct route for progress, staging, submission, and status.
```
gtanczyk: The comment trimming here is sanctioned by the ratchet — `.github/copilot-instructions.md` asks for one-line `//` comments with a shrink-only baseline, and `npm run comment-prose` passes on the branch. No objection to the direction.

The one line I'd keep, as a `//` one-liner, is the invariant from the deleted header: everything the agent sends is untrusted, prompt-influenced text — sanitized here, escaped on render, never fed back to a model as instructions. That's a security property of this file rather than prose debt, and it's the kind of thing a future edit here needs to know it must not break.

---

gtanczyk (commented):
Checked out the branch, installed deps, and ran the affected suites — `source-patch`, `build-prompt`, `agent-channel`, `mcp-server`: **185 tests pass**, and eslint (including `comment-prose`) is clean on all four changed source files. The green-gate claim holds.

**One confirmed bug, in feature 3.** `ackInboxIds` is applied at the top of the `end` handler, ahead of both rejection paths, so an `end` that gets refused (`rejected: 'stopped'`, or a handoff that didn't start) still marks the creator's queued change requests as delivered — and nothing re-queues them. Reproduced against this branch; details inline. It's also the only one of the three features with no test. Worth fixing before merge.

**The other two look sound:**

- The preview-version fix is correct and consistent with `managed-backend.ts:502`, which already computed `roundDelivered` as `deliveredVersion || previewVersion`. Since `setSubmissionDeliveredVersion` writes both fields while preview writes only `previewVersion`, the disjunction is the right signal and can't regress the publish case.
- Multi-patch mode arbitration is airtight in both the zod refinement and the MCP pre-check — notably `patches: []` falls through to the "pass something" error rather than a silent no-op. The theoretical 20 MB payload (50 × 200 KB × 2) is capped by the route's existing `bodyLimit: 500_000`.

Everything except the ack ordering is a nit and fine to take or leave.

---

gtanczyk:
@codex

PR #811, opened 2026-08-14
fix(api): fail closed when Copilot's MCP lane has no scratch repo

gtanczyk:
## What happened

A managed round dispatched to Copilot was supposed to run in `gamedevpl/scratchpad` over MCP. It did neither — the task landed in `gamedevpl/www.gamedev.pl-games` on the `game-builder` harness agent, with no MCP and no warning anywhere. The only tell was the task URL.

The cause is a silent fallback in `managed-provider-copilot.ts`:

```ts
if (promptLane === 'mcp' && mcpTasks) { ... }   // scratchpad, game-builder-mcp
// falls through to the harness dispatch when mcpTasks is undefined
```

`mcpTasks` exists only when `MANAGED_AGENT_COPILOT_MCP_REPO` is set. Unset, an MCP-lane round quietly became a harness round in the private games repo — which is precisely the exposure #798 added the separate repo to remove: a prompt-injectable session sharing a checkout with real proprietary source.

## Changes

- `startSession` throws when the lane is `mcp` and no scratch repo is configured, instead of falling through. The dead `promptLane === 'mcp' ? null : …` seed-staging ternary below it goes away with it.
- `agent-backend-env.ts` refuses to build a copilot backend at all when the lane is `mcp` and `MANAGED_AGENT_COPILOT_MCP_REPO` is missing — the same fail-closed shape every usage ceiling in that file already uses. A half-configured lane is worse than a disabled one.
- Documents all three `MANAGED_AGENT_COPILOT_MCP_*` variables. The code comment already pointed at `docs/managed-agent-backend.md` for them, but the env table never listed them. Adds a note that enabling the lane takes two variables (`MANAGED_AGENT_PROMPT_LANE=mcp` **and** the repo), not one.

Two existing tests encoded the old fallback and now assert the refusal instead.

## Not in scope

This is the guardrail, not the rollout. The operator steps from #798 are still outstanding and are what actually enables the lane: setting the repo variables, granting `AGENT_TASKS_TOKEN` access to `gamedevpl/scratchpad`, and creating the `copilot-mcp-connector` secret there. After this change a deploy missing the repo variable disables copilot dispatch loudly rather than serving it from the games repo.

## Validation

- `npm run test -w @gamedevpl/api` — 3171 tests, 198 files, all pass
- `npx eslint` on changed files, `--max-warnings 0` — clean
- `npm run comment-prose` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01N56phUDE38ZDGRa7PanE5P)_

### DIFF
--- a/apps/api/src/agent-backend-env.test.ts
+++ b/apps/api/src/agent-backend-env.test.ts
@@ -119,7 +119,8 @@ describe('createAgentBackendRegistryFromEnv', () => {
     );
   });
 
-  it('can select the Copilot MCP lane without changing backend selection', () => {
+  // Half-configured is worse than off: the MCP round would hit games.
+  it('fails closed when the Copilot MCP lane has no scratch repo', () => {
     setEnv({
       MANAGED_AGENT_VENDOR: 'copilot',
       AGENT_TASKS_TOKEN: randomBytes(32).toString('hex'),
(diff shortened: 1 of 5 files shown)


gtanczyk:
@codex

PR #816, opened 2026-08-14
Give the working copy its own rail pane

gtanczyk:
## Summary

The own-IDE checkout shipped inside the **Agent access** pane. In use that turned out to be effectively undiscoverable: two clicks deep, behind a lock icon that promises secrets — not what a creator is looking for when they want to open their game in an editor.

This gives it its own pane in the Details rail, between **Screenshots** and **Agent access**, with a download glyph.

## Changes

- **`CreatorStudioView.tsx`** — new `workspace` pane; the checkout panel moves out of the credentials pane into it. The rail entry appears only when the game has a slug and is not abandoned: no slug means no sources, and the route would refuse the request, so offering the pane would open an empty one.
- **`PixelIcon.tsx`** — adds a `download` glyph (arrow into a tray). The nearest existing candidate, `collapse`, is a resize chevron pair that reads as *shrink*, not *save*.
- **i18n** — `studioPanel.rail.workspace`: "Your own IDE" / "Twoje IDE".
- **`docs/own-ide-checkout.md`** — status was "Built, not yet released", which is no longer true. The feature is live and the round trip has been exercised against production; the note now says so while flagging that it is not announced to creators and the archive README still carries a terms placeholder.

## Notes for review

- **The gating test got sharper rather than just relocated.** It previously opened the credentials pane and looked for the panel; now the assertion for a game with nothing to check out is that the rail icon is *absent* — the pane cannot be reached at all. Its name was also stale ("beside the agent keys") and is corrected, since left alone it would misdescribe the behaviour for the next reader.
- **Rebased across the Studio refactor.** ~288 commits landed on master since this work began, including a large `CreatorStudioView` rework. The rail/pane architecture survived, so the change still applies cleanly in intent — but the diff conflicted on reindentation plus a new `onSwitchToPlatform` prop. Rather than resolve hunk-by-hunk (the way a neighbouring prop gets silently dropped) I took master's file wholesale and re-applied the three deltas with exact-match assertions.
- **Comment prose.** The first pass tripped the `comment-prose` baselines on three files. The comments were shortened to one-liners rather than the baselines raised.

## Verification

1345 web tests, `npm run lint` (including the prose gate) and `tsc --noEmit` all clean.

No server-side change — the endpoint, archive composition and rate limit are untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_018ny5RYa1cFtxT4kefc5iua)_


gtanczyk:
@codex

PR #817, opened 2026-08-14
fix(api): cancel Copilot MCP rounds on the repo that actually holds them

gtanczyk:
## The bug

GitHub scopes Agent Tasks per repo *in the URL*, so a session started on one lane is invisible to the other lane's client. `cancelSession` only ever asked the games-repo client:

```ts
const branch = await tasks.getTask(sessionId)   // games repo — an MCP round isn't here
  .then((task) => (task ? resolveTaskBranch(task) : null))
  .catch(() => null);
if (!branch) return { enforced: false };        // reads as "nothing to stop"
```

For an MCP round `getTask` returns null, the method reports `{ enforced: false }`, and nothing is stopped. #798 fixed exactly this for *polling* — `getSession` falls back to `mcpTasks` — but left cancel behind.

**This is live.** Since production cut over to the MCP lane this morning, both ceilings the managed backend enforces on observation have been silent no-ops:

- `managed-backend.ts:417` — the credit budget stop
- `managed-backend.ts:447` — the wall-clock expiry

Each calls `cancelSession`, each believed the `{ enforced: false }`, and a round that blew its ceiling kept running and kept spending. This is the failure mode the "no ceiling, no dispatch" guard in `agent-backend-env.ts` exists to prevent, arriving through the back door.

## The fix

`cancelSession` now resolves the owning lane the same way `getSession` already does — the tracked lane when the session id is known, otherwise probe both — and cancels workflow runs through *that* lane's GitHub client. Adds `deps.mcpGithub` so the scratch-repo client is injectable, which is what makes this testable at all.

## Validation

Both new tests fail against the previous implementation and pass against this one (verified by reverting the lane resolution and re-running — 2 failed / 15 passed):

- `cancels an MCP-lane session on the scratch repo, not the games repo`
- `cancels an untracked MCP session after a provider restart`

Full gate:

- `npm run test -w @gamedevpl/api` — 3194 tests, 198 files, all pass
- `npx eslint` on changed files, `--max-warnings 0` — clean
- `npm run comment-prose` — clean

## Context

Found while mapping the harness lane for removal. That refactor collapses these two clients into one and makes this class of bug structurally impossible — but it is a larger change, and this ceiling should not stay broken while it lands.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N56phUDE38ZDGRa7PanE5P)_

### DIFF
--- a/apps/api/src/managed-provider-copilot.test.ts
+++ b/apps/api/src/managed-provider-copilot.test.ts
@@ -38,6 +38,14 @@ function run(overrides: Partial<WorkflowRun> = {}): WorkflowRun {
   return { id: 1, path: COPILOT_AGENT_WORKFLOW_PATH, status: 'in_progress', ...overrides };
 }
 
+function emptyTasks() {
+  const getTask = vi.fn(async () => null);
+  return {
+    client: { startTask: vi.fn(), getTask, listTasks: vi.fn(async () => []) } as unknown as AgentTasksClient,
+    getTask,
+  };
+}
+
 function tasks(result: AgentTask = task()) {
   const startTask = vi.fn(async (_input: AgentTaskInput) => result);
   const getTask = vi.fn(async () => result);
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #823, opened 2026-08-14
TA-01/TA-02: ghost-text autocomplete (API + web)

gtanczyk:
## Summary

The LLM ghost-text autocomplete track (TA-*), following CE-36/GA-01..09 (deterministic autocomplete, shipped separately). Both the API route and the web client that calls it.

### TA-01 — API

- `POST /api/me/studio/games/:slug/sources/complete` — the ghost-text proposal endpoint, same owner-auth/404 posture as every other Code Surface route.
- `tab-complete.ts`: `VertexTabCompleter` assembles the client's prefix/suffix window into one prompt (Vertex Gemini has no native FIM `suffix` field, so this is prompt-based FIM), reusing the existing Vertex AI integration (`gemini-3.7-flash`) — no new provider, no new credential. Thinking disabled and output capped at 300 tokens (see "Verification" below for why).
- `TAB_COMPLETE` kill switch, off by default — the opposite default from `CODE_SURFACE`.
- Two independent spend controls: a per-creator daily call quota (`tabCompletes`) and a global 2,000,000-token/day hard cap (`createTabCompleteGate`, peek-before-call/spend-after). Operator toggle for both wired into the admin panel.
- Privacy policy (en/pl) updated to disclose the new data flow.

### TA-02 — Web

- Idle-debounced (300ms) ghost-text extension in `CodeMirrorEditor.tsx`, wired to the route through `CodeSurface.tsx`.
- Hand-rolled rather than the research doc's first-pick library (`codemirror-extension-inline-suggestion`): a real-Chromium test against a real model call found the library's state field resets on *any* transaction lacking its own effect, so this app's own autosave/typecheck dispatches silently erased a shown proposal within ~400ms of it rendering. The hand-rolled version (StateField + widget decoration + `Prec.highest` Tab/Esc keymap) only resets on an edit, a bare cursor move, or explicit dismiss.
- Suppressed while the CE-36 `.`-triggered tooltip is genuinely showing (`completionStatus === 'active'`, not `'pending'` — pending fires on nearly every keystroke and would have starved ghost text entirely), mid-selection, or over an open hover/lint tooltip.
- Tab inserts the proposal as a normal buffer edit through the existing autosave/staging path — no special-cased accept path needed.

## Verification

- Real, non-mocked verification: spun up the API + web dev server locally with a real `gemini-3.7-flash` client (API-key path, local-only) and drove real Chromium against it.
- This caught two real bugs before they shipped: the output-token cap needed thinking disabled first (Gemini's extended-thinking tokens share the same budget and were starving the visible answer), and the first-choice ghost-text library's decoration didn't survive the app's own background dispatches.
- Confirmed end to end: proposal renders with correct, contextual completions; survives autosave + typecheck; Tab accepts into the real buffer; Escape dismisses.

## Test plan

- [x] `apps/api/src/tab-complete.test.ts` — prompt assembly, token accounting, markdown-fence stripping, kill-switch default, thinking disabled (regression-pinned via a mocked Vertex client), output-token cap
- [x] `apps/api/src/creator-code.test.ts` — new route: kill-switch 404, ownership 404, happy path, per-creator quota 429
- [x] `apps/web/src/CodeSurface.languageService.test.tsx` — `fetchGhostText` prop wiring
- [x] `npx tsc --noEmit` clean on both `apps/api` and `apps/web`
- [x] `npx eslint` clean on all touched files
- [x] Full API suite (3200 tests) and full web suite (1358 tests) passing
- [x] `apps/web` production build succeeds
- [x] Repo-wide comment-prose gate at baseline
- [x] Real Chromium + real Gemini call, screenshots attached to the session (proposal render, persistence through autosave/typecheck, Tab-accept)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -226,6 +226,13 @@ jobs:
             ENV_VARS="${ENV_VARS}|CODE_LANE=${CODE_LANE_VAL}"
           fi
 
+          # TA-01's ghost-text completion route: off unless set, same reasoning as
+          # EDITOR_ASSIST above — a hand-set value would vanish on the next deploy.
+          TAB_COMPLETE_VAL="${{ vars.TAB_COMPLETE }}"
+          if [ -n "$TAB_COMPLETE_VAL" ]; then
+            ENV_VARS="${ENV_VARS}|TAB_COMPLETE=${TAB_COMPLETE_VAL}"
+          fi
+
           # The Code surface (creator-code-editing-execution-plan.md CE-02): manual
           # editing of a game's own sources in the Studio. A kill switch, not a cohort
           # gate — the owner decision is all-creators, so it ships on by default and this
--- a/apps/api/src/admin.test.ts
+++ b/apps/api/src/admin.test.ts
@@ -536,6 +536,8 @@ describe('/api/admin/creation-limits', () => {
         configuredVendors: [],
         defaultVendor: null,
       },
+      tabCompletePaused: false,
+      globalDailyTabCompleteTokenCap: 2_000_000,
     });
     // A pause left on by accident is this feature's own failure mode, so the record of
     // who set it is part of the deliverable.
@@ -570,6 +572,30 @@ describe('/api/admin/creation-limits', () => {
         configuredVendors: [],
         defaultVendor: null,
       },
+      tabCompletePaused: false,
+      globalDailyTabCompleteTokenCap: 2_000_000,
+    });
+    await app.close();
+  });
+
+  it('sets the tab-complete pause and cap, independent of the other lanes', async () => {
+    const app = await appWith(store);
+
+    const res = await app.inject({
+      method: 'POST',
+      url: '/api/admin/creation-limits',
+      headers: authHeaders('g:boss'),
+      payload: { tabCompletePaused: true, globalDailyTabCompleteTokenCap: 500_000 },
+    });
+
+    expect(res.statusCode).toBe(200);
+    const body = res.json() as CreationLimitsResponse;
+    expect(body.effective.tabCompletePaused).toBe(true);
+    expect(body.effective.globalDailyTabCompleteTokenCap).toBe(500_000);
+    expect(body.effective.paused).toBe(false);
+    expect(await store.getCreationLimits()).toMatchObject({
+      tabCompletePaused: true,
+      globalDailyTabCompleteTokenCap: 500_000,
     });
     await app.close();
   });
--- a/apps/api/src/admin.ts
+++ b/apps/api/src/admin.ts
@@ -167,6 +167,9 @@ const CreationLimitsPatchSchema = z
     // The studio chat breaker rides the same document too.
     chatPaused: z.boolean().optional(),
     globalDailyChatCap: z.number().int().min(0).max(100_000).nullable().optional(),
+    // TA-01's own breaker, denominated in tokens rather than calls.
+    tabCompletePaused: z.boolean().optional(),
+    globalDailyTabCompleteTokenCap: z.number().int().min(0).max(50_000_000).nullable().optional(),
     // Same document: whether the platform builder is offered. See managed-availability.ts.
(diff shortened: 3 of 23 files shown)


gtanczyk:
@codex

gtanczyk:
@codex

gtanczyk:
@claude

gtanczyk:
All 7 findings from the Codex review are addressed in `9149c1a`:

1. **Token reservation race** — `TabCompleteGate.peek()` now reserves a worst-case `TAB_COMPLETE_TOKEN_RESERVATION` (1400 tokens) against the global daily cap *before* the call, and `spend()` reconciles it against actual usage afterward via a new `Store.adjustGlobalTabCompleteTokens` (signed delta, floored at zero, never refuses). Closes the concurrent-burst overshoot window.
2. **Ordering bug** — the global gate is now checked before the per-creator quota is spent in `creator-code.ts`, and every early-exit path (quota refusal, completer error) releases the reservation via `spend(uid, dateStr, 0)`.
3. **`stripFences()` whitespace corruption** — now trims only a detection copy of the text; the returned completion keeps real leading/trailing whitespace. Added a regression test.
4. **Admin panel missing tab-complete controls** — `admin.ts` now exposes `tabCompletePaused`/`globalDailyTabCompleteTokenCap` in both `effective` and `today`, and `CreationLimitsPanel.tsx` got a full "Tab completion (ghost text)" section mirroring the existing pause/cap UI.
5. **`TAB_COMPLETE` not threaded through deploy** — added to both `.github/workflows/deploy.yml` and `infra/deploy-api.sh`, plus a new `tab-complete-deploy-flag.test.ts` parity test (same pattern as the existing MCP_UI test).
6. **Privacy copy overstated an opt-in that doesn't exist** — corrected in both `privacy.en.ts` and `privacy.pl.ts` to describe it accurately as automatic while editing your own game, not an opt-in toggle.
7. **Hardcoded "TypeScript" in the prompt for non-TS files** — ghost text is now restricted client-side to `.ts`/`.tsx` files in `CodeSurface.tsx`, with a regression test confirming non-TS files never fetch.

Full verification before this commit: `tsc --noEmit` clean on both apps, `eslint` clean on all touched files, full API suite (3219 tests) and web suite (1359 tests) passing, `vite build` succeeding, and the repo-wide comment-prose gate at baseline.

---

gtanczyk:
@codex

gtanczyk:
Thanks for the follow-up pass — all 3 new findings are real and fixed in `eca0565`:

1. **Reservations could cross the cap** — `checkAndIncrementGlobalTabCompleteTokens` compared `current >= limit`, so with a small cap (e.g. 100) the very first 1400-token reservation would be admitted and blow past it. Fixed in both `InMemoryStore` and the Firestore store to compare the post-reservation total (`current + tokens > limit`) instead. Added a regression test with a cap smaller than one reservation.
2. **Reconciling a reservation that never happened** — if the reservation write in `peek()` threw, the call was admitted uncounted (no reservation recorded), but `spend()` always subtracted `TAB_COMPLETE_TOKEN_RESERVATION` regardless, corrupting the counter once Firestore recovered. `peek()` now returns whether it actually reserved (`{ allowed, reserved }`), and the caller threads that flag into `spend()`, which only subtracts the reservation when one was really taken. Added a regression test that fails the reservation write and asserts real usage is preserved rather than underflowed.
3. **Tab could accept stale ghost text over an open popup** — the `Prec.highest` ghost-text Tab handler only checked whether a proposal existed, not whether the CE-36 autocomplete popup was active, so a ghost text proposal set before the popup opened could win Tab away from it. Fixed two ways: `ghostTextField`'s `update()` now clears itself once `completionStatus() === 'active'`, and the Tab handler independently rechecks `completionStatus()` before falling back to the built-in popup binding.

Full verification: `tsc --noEmit` clean on both apps, `eslint` clean, full API suite (3221 tests) and web suite (1359 tests) passing, `vite build` succeeding, repo-wide comment-prose gate at baseline.

---

PR #826, opened 2026-08-15
Support sourced audio catalog fallback for game sounds

gtanczyk:
## Summary
Add support for falling back to the sourced audio catalog (`shared/audio/sourced.json` + `shared/audio/sourced/*.mp3`) when a game references sounds that aren't available in the synth catalog (`shared/audio/assets/*.wav`). This allows games to use vendor-generated audio clips that cannot be re-rendered from parameters.

## Key Changes
- **Sound resolution logic**: Implemented a two-tier lookup system that checks the synth catalog first, then falls back to the sourced catalog before treating a sound as missing
- **Sourced catalog interface**: Added `SourcedAudioCatalog` interface to parse `shared/audio/sourced.json` and extract MIME type information
- **Lazy catalog loading**: Sourced catalog is only loaded once and cached if needed, avoiding unnecessary fetches when all sounds are in the synth catalog
- **DrumKit sample embedding**: Extended audio asset resolution to include drumKit samples (kick/hat) referenced from music tracks, which are not directly selected by games but need embedding to avoid silent fallbacks
- **Unified asset handling**: Consolidated audio asset management into a single `assets` object that includes both direct game sounds and indirect drumKit samples

- The `resolveSoundAsset` function encapsulates the fallback logic and is reused for both game sounds and drumKit samples
- If any sound is missing from both catalogs, the entire source generation fails (returns null) rather than silently degrading
- MIME type defaults to `audio/mpeg` if not specified in the sourced catalog
- Asset chunks are reordered to ensure the assets declaration comes first in the generated JavaScript

https://claude.ai/code/session_011AAR4UoEUM63b8aJgnxrVp

### DIFF
--- a/apps/api/src/github-client.test.ts
+++ b/apps/api/src/github-client.test.ts
@@ -904,9 +904,6 @@ describe('getGameSources', () => {
   });
 
   it('falls back to the sourced-audio catalog for a sound missing from the synth catalog', async () => {
-    // shared/audio/sourced.json + shared/audio/sourced/*.mp3: vendor clips that cannot be
-    // re-rendered from parameters the way the synth catalog can. A GAME.json sound name that
-    // misses shared/audio/assets/*.wav should resolve here before being treated as missing.
     const files = new Map<string, string | Uint8Array>([
       ['games/march/index.html', '<canvas id="game"></canvas>'],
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #835, opened 2026-08-15
Studio live-editing, Track 1: rebuild pipeline timing instrumentation

gtanczyk:
## Summary

Second of the stacked series toward live editing in Studio (stacked on #834 — this PR's diff is only `github-client.ts`/`github-client.test.ts`/`staged-preview.ts`/`docs/`; the base compare will shrink once #834 merges).

Adds the measurement layer the fast-lane work (Track 2, next in the stack) needs before it changes anything:

- `GitHubClient.getGameSources` returns an optional `timings` field on a successful assembly: `baseReadMs` (game+engine file reads), `kitModulesMs` (GameKit module compile), `audioMs`, `musicMs`, `bundleMs` (esbuild), `totalMs`. Absent on every existing mock/test double and on any `null` return — purely additive, best-effort telemetry.
- `staged-preview.ts`'s `attempt()` logs one `info` line per successful publish (`'staged preview assembled'`) with the full wall-clock breakdown: overlay build, `getGameSources` (+ its phase breakdown), `assembleGameHtml`, and the Firestore store write. `StagedPreviewOptions.log.info` is optional and the real deployment already wires `app.log` (Pino), so no new plumbing is needed in production.
- `docs/live-editing-latency.md` documents what's instrumented, what's deliberately not (the remix-lane bench explicitly excludes network cost; the client/server scheduler constants are already-known values, not something to measure), and the specific questions Track 2 should answer from these numbers before optimizing.

## Test plan
- [x] New test: every real `getGameSources` assembly reports non-negative per-phase timings
- [x] Fixed one pre-existing test that did a deep `toEqual` between two assembly paths (contents-API vs archive-backed) — excluded `timings` from that comparison since it's process telemetry, not assembly output, and the two paths are expected to differ in speed by design
- [x] Full API suite: 197 files / 3142 tests green
- [x] `eslint` + `comment-prose` clean on all changed files, repo-wide comment-prose budget unchanged
- [x] `tsc --noEmit` clean

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ)_

PR #834, opened 2026-08-15
Studio: CE-17 implicit round-open, diff-vs-live, i18n fix

gtanczyk:
## Summary

First of a stacked PR series toward a live-editing loop in Studio (research context: this fixes a real dead end I hit while editing a published game's code in the Code surface).

- **CE-17, implicit round-open**: `PUT /sources/stage` and `POST /sources/stage/patch` now open a fresh manual round automatically on the first owner staging write into a closed round (e.g. an already-published game), instead of 409ing with "start a new round from the thread first." Mirrors `startImprovementRound`'s job-creation shape but skips `dispatchBuild` — no agent is involved. `resolveOwnedRecord`'s newest-`createdAt`-wins scan picks the new job up automatically; no slug pointer to move. Deliver still requires an active round (unchanged) — staging is what opens one now.
- **Diff vs live**: `GET /sources` now returns each staged file's delivered `base` content alongside the staged content. The Code surface renders a "Diff vs live" toggle (only when a base exists) showing a real LCS line diff instead of forcing the creator to compare mentally.
- **i18n fix**: the deliver route's remaining `no_active_round` message (still reachable if a creator clicks Publish with nothing staged on a closed round) was raw English leaking into an otherwise Polish UI. Now localized in both `en`/`pl` via an error-code check instead of showing the server string verbatim.
- Added a `round_reopened` funnel step across the client/server telemetry lockstep (`visitTelemetry.ts`, `visit-funnel.ts`, `visit-telemetry.ts`) and the operator `VisitFunnelPanel`, to see how often this path fires.

## Test plan
- [x] `npx vitest run` — full API suite (197 files / 3142 tests) and web suite (159 files / 1364 tests) both green
- [x] New tests: CE-17 implicit-open (stage + patch, including base-content resolution against a real published version), diff-toggle rendering, round-opened notice + auto-dismiss, `diffLines` unit tests
- [x] `npm run lint` equivalent (`eslint` + `comment-prose`) clean on all changed files, repo-wide comment-prose budget unchanged
- [x] `npm run type-check` clean for both `api` and `web` workspaces

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ)_

PR #837, opened 2026-08-15
Studio live-editing, Track 2: fast lane (engine cache, sync preview, input-aware swap)

gtanczyk:
## Summary

Third of the stacked series (stacked on #835, which is stacked on #834 — this PR's own diff is scoped to the files listed below; the base compare shrinks as earlier PRs merge).

This is the actual latency win the research called for: taking Code-surface edit→preview from ~20-30s (agent-tuned scheduler + uncached engine reads) toward the ~100-300ms floor the instrumentation in #835 predicted.

**Per-engine-SHA caching** (`github-client.ts`) — `shared/game-shell.css`, the compiled `core.ts` IIFE, every GameKit module compile, and the kit declaration are now cached per resolved commit SHA, shared across every concurrent editor. This also fixes a real staleness bug: the old `gameKitByRef` cache was keyed on the raw (possibly mutable-branch) ref and its own docstring incorrectly claimed refs were immutable — a games-repo push to a branch ref would never be picked up by a warm process. Refs now resolve to a SHA with a 60s TTL, so a cache hit is genuinely safe and a push is visible within a minute. Bounded to 2 warm SHAs (mirrors `kit-files.ts`'s own `KIT_TREE_CACHE_CAPACITY`).

**`POST /sources/preview`** (`creator-code.ts`) — a synchronous rebuild endpoint for the Code surface. Bypasses `stagedPreview`'s debounce/min-gap scheduler entirely (that scheduler's 6s/25s constants exist to survive an *agent's* burst-of-small-writes cadence, and don't fit a single owner click) and mirrors `remix.ts`'s `rebuild()` shape instead: read staged sources → overlay → `getGameSources` → `assembleGameHtml` → return the document inline. No `BuildPreview` write, no polling.

**Input-aware swap policy** (`StudioStage.tsx`) — implements the swap-policy decision from the design discussion: a build that arrives mid-play now auto-applies the instant player input goes idle (400ms window), tracked via the existing bridge `activity`/`pointer` messages in a ref (so tracking it never triggers a render). No more manual "restart on it" / "finish this run" toast — leaving play posture is the unconditional safety net if a swap is still held. Removed the now-dead toast UI, its CSS, and its unreachable i18n strings.

## Codex review fixes
- **The new endpoint had no caller.** `POST /sources/preview` existed but nothing in the web app requested it — the Code surface still went through the old debounced/polled path exclusively. It now requests the sync preview alongside the existing debounced rebuild and shows the result the instant it lands (`useStageSource`'s new `pushPreview`), which is the actual latency win this PR was for.
- **Cache-key staleness under a moving branch.** A cache miss for `getCachedGameShellCss`/`getCachedCoreJs`/`getCachedGameKitModule` read through the mutable `ref` instead of the already-resolved SHA — if the branch moved mid-TTL, newer bytes could get stored under an older SHA's cache key. These now take only the SHA.
- **Idle-detection missed held input.** The 400ms swap-hold window only saw discrete `keydown`/`pointerdown` pings, so a held touch joystick or a mouse drag could look idle and get its run swapped out mid-input. A new bridge message (`held`) reports press/release separately from the existing `activity` signal — kept separate because `GameTheater`'s own chrome-autohide timer deliberately excludes movement from the same signal, for a different reason.

## Test plan
- [x] New tests: engine-cache hit avoids re-reading `shared/` for a second game on the same ref; a branch's SHA cache expires per its TTL and picks up a moved branch; the preview endpoint's config/ownership/agent-lock/incomplete-overlay/success paths; `StudioStage` tests covering hold-while-active → auto-apply-on-idle, immediate-apply-when-already-idle, apply-on-leaving-play-mid-input, held-pointer-blocks-idle-apply, and the sync-preview wiring
- [x] Full suites: API 199 files / 3172 tests, web 163 files / 1415 tests, all green
- [x] `eslint` + `comment-prose` clean
- [x] `tsc --noEmit` clean for both workspaces

Co-Authored-By: Claude Sonnet 5 

https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ptra1eaV7EdeeS8MZP7USJ)_


gtanczyk:
@codex review

---

PR #839, opened 2026-08-15
fix: seed generation and Code lane silently 400ing on gemini-3.7-flash since #803

gtanczyk:
## Summary

Found while live-testing #820: two real production seed-generation attempts on ordinary self-build rounds both came back `seedStatus: unavailable` instead of a draft. Traced to a regression, unrelated to #799/#820, that's been silently breaking two live features since **#803** merged (2026-08-13).

## Root cause

`game-seed.ts`'s picker call and `code-lane.ts`'s only call both set the *raw* Vertex `generationConfig.thinkingConfig = { thinkingBudget: 0 }` directly on client construction, instead of using genaicode's request-level `.thinking()` API every other Vertex caller in this codebase uses.

`#803` bumped `DEFAULT_SEED_MODEL` / `DEFAULT_CODE_MODEL` to `gemini-3.7-flash`. That model rejects a raw `thinkingBudget: 0` outright with `INVALID_ARGUMENT` — the same way it rejects the `MINIMAL` level genaicode's own auto-conversion maps a zero budget to. **`1e37e2d`** (merged the next day) already fixed this exact class of bug at five other call sites, but explicitly exempted these two:

> code-lane.ts and game-seed.ts use `.run()` with no `.json()`/`.thinking()`, so their provider-level thinkingConfig is not silently overwritten — left as is.

That reasoning is correct about a *different* bug (genaicode silently overwriting an *unset* thinking config to a value it then rejects) but doesn't apply here: the literal raw value these two files set was never valid for Gemini 3 to begin with, overwritten or not. Both calls are fail-open by design, so the failure produced no visible symptom anywhere — no error, no alert, just a log line — for two days.

## Blast radius

- **Seed generation**: every round-0 seed attempt (self/BYOCA and, since #820, managed `mcp`-lane rounds) has been failing since the #803 deploy.
- **Code surface** (`code-lane.ts`, in-app AI code editing): every request has been failing the same way — same root cause, same fail-open blind spot.

Confirmed unaffected, by re-checking every other file #803 touched: `refine.ts`, `moderation.ts`, `chat-agent.ts`, `editor-assist.ts`, `feedback-themes.ts` all already use the structured `.thinking()`/`.json()` API (several of them fixed by `1e37e2d` itself) and are fine — verified `refine.ts` live against production during this investigation.

## Fix

Matches `1e37e2d`'s own established pattern exactly: drop the raw `thinkingConfig`, add `.thinking({ level: 'low' })` — the actual accepted floor on `gemini-3.7-flash`, confirmed by the working call sites — to the request chain in both files.

## Testing

- Test fakes for both files needed a `thinking()` method added to their stub chain object (they didn't have one, which is exactly why this regression didn't show up in CI — the fakes were more permissive than the real API).
- New test in `game-seed.test.ts` captures the actual value passed to `.thinking()` and pins it to `{ level: 'low' }`; confirmed it fails against the old code (`.thinking(false)`) before restoring the real fix.
- Full suite: 3165 tests pass, lint (including `comment-prose`) and typecheck clean.

https://claude.ai/code/session_01DLh3ShdDejjF9Xn5HSAJ2Q

---
_Generated by [Claude Code](https://claude.ai/code/session_01DLh3ShdDejjF9Xn5HSAJ2Q)_

### DIFF
--- a/apps/api/src/code-lane.test.ts
+++ b/apps/api/src/code-lane.test.ts
@@ -30,6 +30,7 @@ function stubClient(responses: unknown[], usage = { inputTokens: 100, outputToke
     prompts.push(prompt);
     const payload = responses.shift();
     const chain = {
+      thinking: () => chain,
       temperature: () => chain,
       signal: () => chain,
       run: () => Promise.resolve({ parts: [{ type: 'text', text: JSON.stringify(payload) }], usage }),
--- a/apps/api/src/code-lane.ts
+++ b/apps/api/src/code-lane.ts
@@ -339,7 +339,6 @@ export class VertexCodeLane {
         defaultModel: DEFAULT_CODE_MODEL,
         generationConfig: {
           responseMimeType: 'application/json',
-          thinkingConfig: { thinkingBudget: 0 },
         } as VertexGenerationConfig,
       });
     return this.client;
@@ -360,6 +359,7 @@ export class VertexCodeLane {
     tokens: { input: number; output: number },
   ): Promise<T> {
     const result: GenerationResult = await this.getClient()(prompt)
+      .thinking({ level: 'low' })
       .temperature(0.1)
       .signal(AbortSignal.timeout(timeoutMs))
       .run();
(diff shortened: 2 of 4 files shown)


gtanczyk:
@codex

PR #842, opened 2026-08-15
Fix oversized empty Studio prompt field

gtanczyk:
## Summary

- Clear the auto-grow inline height when the compact feedback composer is empty.
- Add a regression assertion covering the empty composer state.

## Root cause

The resize effect ran on the initial empty render and applied the textarea's measured height, overriding the CSS one-line height and leaving a large blank prompt field.

## User impact

An empty Studio prompt now stays at its compact CSS height. Typed or seeded feedback still auto-grows up to the existing 220px cap.

## Checks

- `npm run type-check` ✅
- `npm run lint` ✅
- `npm run build` ✅
- `npm run test:rules` ✅
- `npm test --workspace apps/web -- --run src/SubmissionStatusView.test.ts` ✅ (70 tests)
- `npm test` ⚠️ 3,156 passed; 8 existing/unrelated API-suite failures/timeouts remain in agent-build-reads, creator-studio, editor-drafts, mcp-server, stage-hints, telemetry, and typecheck-preflight. The web (1,372 tests) and world (20 tests) suites passed.

### DIFF
--- a/apps/web/src/SubmissionStatusView.test.ts
+++ b/apps/web/src/SubmissionStatusView.test.ts
@@ -1696,6 +1696,9 @@ describe('SubmissionStatusView', () => {
     });
 
     const textarea = container.querySelector<HTMLTextAreaElement>('.status-feedback-input');
+    // Embedded Studio uses the compact composer, whose empty state must keep the CSS height.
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #843, opened 2026-08-15
Improve code editor hover and completion

gtanczyk:
## Summary

- add VS Code-like compact hover info and Cmd/Ctrl-expanded documentation
- underline navigable symbols while the modifier is held
- reuse TypeScript completion results while the current identifier grows
- record anonymous completion health and latency separately for TypeScript and ghost text
- expose completion request counts and median/P90 latency in the admin funnel

## Why

The editor had a single hover presentation, no modifier navigation affordance, and no completion latency telemetry. The TypeScript completion adapter also did not provide a `validFor` range, so the worker could be queried again on every identifier character. Ghost-text failures were swallowed as empty responses, making failures indistinguishable from valid no-suggestion responses.

## Validation

- `npm run type-check`
- `npm run lint`
- `npm run build`
- focused editor, telemetry, and funnel tests passed
- full workspace tests were attempted; unrelated websocket/world suites cannot bind localhost in this sandbox (`listen EPERM`), which also causes cascading UI test timeouts

### DIFF
--- a/apps/api/src/admin.test.ts
+++ b/apps/api/src/admin.test.ts
@@ -244,6 +244,35 @@ describe('GET /api/admin/telemetry/visits', () => {
     expect(body.days).toContain(today);
   });
 
+  it('keeps completion diagnostics out of the core event read budget', async () => {
+    const coreEvents = Array.from({ length: 1000 }, (_, index) =>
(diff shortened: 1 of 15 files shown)


gtanczyk:
@codex

PR #845, opened 2026-08-15
Allow sending messages while agent is building

gtanczyk:
## Summary
This change allows creators to send messages to the agent while it's actively building, instead of requiring them to stop the build first. Messages sent during a build are queued to the agent's inbox and the UI now displays delivery status (queued vs delivered) for each message.

## Key Changes

- **Composer behavior**: The textarea and send button are no longer disabled when the agent is working (`agentWorking` state). Messages can now be sent at any time during a build.

- **Message delivery tracking**: Added a `delivered` field to track whether the running agent has picked up a message from its inbox yet:
  - Messages show "Queued for the agent" (muted color) when pending
  - Messages show "Delivered to the agent" (success color) once the agent has read them
  - This replaces the previous "The agent is building now" blocking message

- **Simplified UI copy**: Removed separate messaging for the "busy" state since messages no longer wait for the build to stop. The building state now uses the same copy as before, with delivery status shown inline on each message.

- **API updates**: 
  - Added `delivered?: boolean` field to `CreatorRevision` type in submission status API
  - Updated `BuildProgress` type to include delivery tracking
  - Backend now includes delivery status when returning revision history

- **Removed i18n keys**: Deleted unused translation keys for the old "busy" state (`hintBusy`, `busyComposerHint`, `busy`)

- **Tests**: Updated existing tests to reflect that the composer stays enabled during builds, and added a new test verifying message queuing/delivery status display and continued sending capability during builds

## Implementation Details

The change removes the `agentWorking` condition from send validation and UI disabling logic, allowing messages to flow to the agent's inbox regardless of build state. The delivery status is now the primary way users understand whether their message has been picked up, replacing the previous blocking UX that prevented sending during builds.

https://claude.ai/code/session_01WDUHWZiRVuBPyU15wkxCbG


gtanczyk:
@codex

PR #848, opened 2026-08-15
Add WebMCP tool registration for browser agents in Code surface

gtanczyk:
## Summary

Lets a creator opt into agent-assisted editing on the Studio Code surface, three ways at once — one tool contract behind all of them:

- **WebMCP**, for a browser agent that supports it — registers `document.modelContext` / `navigator.modelContext` tools that mirror the platform's regular BYOCA MCP contract (`apps/api/src/mcp-server.ts`): `get_sources`, `stage_source_file`, `patch_source_file`, `clear_staged_sources`, `submit_sources`, plus `studio_get_agent_guide` for onboarding. Same names, same shapes — one contract, not a parallel one. No browser ships a WebMCP-consuming agent yet, so this is inert everywhere today and additive.
- **A real MCP bridge, for Claude Desktop / Claude Code / Cursor** — the same entry point offers the platform's existing `/api/mcp` connector (creator-wide key, already used by BYOCA), so a full MCP client can connect and edit this round directly, right now.
- **An agent console, for browser agents that can only click and type** — Claude for Chrome can read the page and type into it, but cannot call tools or run scripts, so neither of the above reaches it today. The console is a textarea that takes one JSON command (`{"tool":"patch_source_file","input":{…}}`), a Run button, and a result box it can read back. It dispatches through the *same* tool definitions WebMCP registers, so the guardrails are identical. It adds no capability a human at the same keyboard doesn't already have with the same session, and becomes deletable the day browsers ship WebMCP support.

## Key changes

- **`webmcp.ts`**: one `buildCodeSurfaceTools(slug)` definition list, consumed by both the WebMCP registration and the console dispatcher, over the existing `codeSurfaceApi` routes (same creator-session auth, same working-copy model the human editor uses). `submit_sources` is hard-pinned to `mode: "preview"` — publishing stays a human-only action via the Studio "Opublikuj" button, no tool here can publish, through any of the three paths.
- **Opt-in, not automatic**: an "Agent mode" button opens a modal with the WebMCP toggle, the creator key panel, and the console. Tools no longer register just because the surface is editable.
- **Agent activity indicator**: an "Agent is editing" banner lights up while a tool call is in flight.
- Security invariant unchanged: only the Studio shell registers tools; the game iframe never gets `allow="tools"` (`docs/security-model.md`, `GameFrame.sandbox.test.ts`).

## Review findings addressed

- **Stale editor after agent writes (P1)**: mutating tools now emit a completion event and the editor reloads its working copy, so an agent's staged edits can't be clobbered by the creator's next autosave. The existing poll only ran for read-only rounds, which these owner-authenticated calls are not.
- **Telemetry dropped on the floor (P1)**: the new `code_step` values are now accepted by `CodeStepSchema` and the `CODE_STEPS` aggregate. Without that the API rejected the whole batch with 400 and `flush()` discarded it — losing every other visit event sharing it, not just this one.
- **Opt-in scope (P2)**: moved from `localStorage` to `sessionStorage`, so the grant matches the "in this browser tab" wording and dies with the tab instead of silently re-arming in every future tab. Not tied to a round id — the surface is per-slug throughout, and a tab-lifetime grant seemed the honest bound; happy to tighten to round identity if you'd rather.

## Tests

Tool registration and shapes, console dispatch (including malformed JSON and unknown tools), the preview pin holding through the console, the mutation-triggered reload, activity phases, opt-in persistence, and the modal itself.

https://claude.ai/code/session_015hkmqEDu7W6go34rUNqtnN


gtanczyk:
@codex

gtanczyk:
@codex

PR #852, opened 2026-08-15
Persist code undo history across Play

gtanczyk:
## What changed

- Persist CodeMirror undo/redo state per file in the existing Creator Studio session cache.
- Restore history when returning to Code, while rejecting history whose document no longer matches.
- Add regression coverage for real undo restoration, stale history, and CodeSurface remounts.

## Why

Switching from Code to Play unmounted CodeMirror. The draft text survived, but CodeMirror's internal history stack did not, so returning to the editor made Cmd/Ctrl+Z ineffective.

## Validation

- Focused CodeSurface and editor-state tests: passing
- `npm run lint`: passing
- `npm run type-check`: passing
- `npm run build`: passing
- The full suite's web tests pass; socket-based API/world tests are blocked in this sandbox by `listen EPERM`.


gtanczyk:
@codex

PR #870, opened 2026-08-17
feat: give Party a real destination page at /party

gtanczyk:
## Summary

- Adds `/party` as a real, standalone destination — replacing the old header link that just scrolled the home page to `#party-rail` (removed on the redesign branch for exactly that reason: "just a deeplink to the main page").
- `PartyPage.tsx` reuses the existing multiplayer-filtered catalog rail (no new data), and adds what a bare rail couldn't carry: the "one screen, everyone's phone" framing, a 3-step "how it works" explainer, and a CTA pointing at `/create` for a game built ad hoc for that night's group.
- Wired as a first-class route: new `AppRoute` variant, `partyPath()` builder, reserved handle segment, document title, nav active-state highlighting (desktop + mobile menu), and a `party_page` telemetry `via` value.
- The header's "Party" link and mobile menu item now open `/party` instead of scrolling home.
- Fixes a real bug found while wiring the route: the catalog-fetch effect only loaded data on `home`/`play`/`create`, so `/party` would have been stuck on its empty state forever.
- Closes the "Is a dedicated Party destination worth building?" open question recorded in the ops repo's `home-page-plan.md` (companion PR: gamedevpl/www.gamedev.pl-ops#122).

## Test plan

- [x] `npx vitest run` — 175 test files / 1519 tests pass (includes new/updated coverage for the route, page title, and header nav wiring)
- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npx eslint . --report-unused-disable-directives --max-warnings 0` — clean
- [x] `npm run comment-prose` — at or under baseline
- [x] `npx vite build` — succeeds

---
_Generated by [Claude Code](https://claude.ai/code/session_01BamcxDavd42EQLFTg2H8m9)_


gtanczyk:
@codex review

---

gtanczyk:
All 4 findings were real, fixed in ae2085f4:

- Reserved `party` in the API's `RESERVED_HANDLES` too (was only client-side), and added `/party` explicitly to `isKnownSpaShellPath`.
- Registered `party_page` in the server's `PlayViaSchema`/`PLAY_VIAS` (and the funnel's via-label map) so telemetry batches carrying it stop 400ing.
- Added a `party` case to `routeKind()` on both client and server so `/party` visits stop recording as `home`.
- `PartyPage` now takes `catalogStatus`/`catalogError`/`onRetryCatalog` and shows the same loading/error+Retry treatment as `ArcadeCatalog`, instead of claiming "no party games" during a cold load or a failed fetch.

@codex review

---

gtanczyk:
Both new findings were real too, fixed in 46b77367:

- `PartyPage` now shows the same non-blocking refresh-error banner `ArcadeCatalog` uses when a same-page catalog refresh fails (`ready` + `catalogError`) — the stale rail stays visible with a Retry, instead of silently going stale with no signal.
- `handlePartyNav` now resets scroll to top like `handleCreateNav` does, so opening `/party` from partway down a scrolled page doesn't land mid-page.

Also: rebasing onto current master surfaced a real TS build break (CI's "Production image" and "Lint, Test & Build" jobs were red) — master had moved since the last rebase, including an unrelated fix that stripped a stale `onParty` prop that predated this feature, and a new `/play` chrome branch that didn't have `onParty` wired at all. Fixed and confirmed clean: typecheck, lint, comment-prose, full test suite, and both `apps/web`/`apps/api` builds.

@codex review

---

gtanczyk:
Real finding, fixed in eefb062b: `PartyPage` now takes `partyError` and renders a failed Play Together attempt directly above the rail that triggered it, instead of it only appearing near the footer below the rail, steps, and custom-game section (worst on mobile, where it was off-screen). The App-level fallback paragraph is suppressed on `/party` so it doesn't show twice.

Confirmed clean: typecheck, lint, comment-prose, full test suite (added a test asserting the error renders before the rail in DOM order), and build.

@codex review

---

gtanczyk:
Added: `/party`'s "Build a game" button now biases the actual creation flow toward a party game, not just the copy around the button.

There's no structured "multiplayer" field on submission today — per `docs/multiplayer-plan.md`, the building agent infers multiplayer from the concept text itself. So the button previously handed off to a blank `/create` composer exactly like the header's generic Create link, and the party framing never reached the agent that builds the game.

`handlePartyCreateNav` now pre-loads the composer (via the existing `retryPrompt` mechanism Studio's "try this again" already uses) with `"a party game for 3–6 players, phones as controllers, about "`, cursor at the end, so the creator's own description completes that sentence instead of starting from blank. Localized for `pl` too. Verified in a real browser (screenshot below) and added an integration test asserting the pre-loaded text and cursor position.

Rebased onto current master again (2 more commits landed) — clean, no conflicts. Confirmed: typecheck, lint, comment-prose, full test suite, build.

@codex review

---

gtanczyk:
Real gap, fixed in b4dc947d. `retryPrompt` is shared, unscoped state seeding both `/create` and Home's composer on every mount, and nothing ever cleared it — a pre-existing quirk for Studio's "try this again" that the new party starter made visible (leave `/create` without submitting, and the party text kept reappearing on Home and via the header's plain Create button).

Generic Create and Home now clear `retryPrompt` before navigating, so it survives only within the same in-progress flow. Studio's own retry-then-navigate-home call is a separate code path (`CreatorStudioView` calls `onNavigate` directly, not the header's Home handler) and is unaffected. Added a test exercising the exact leak scenario: Build a game → don't submit → click the logo → composer is blank on Home.

Confirmed clean: typecheck, lint, comment-prose, full test suite, build. No rebase needed this time.

@codex review

---

gtanczyk:
Right — patching each newly-found exit path (`onUp={navigate}`, `handleHomeAnchorNav`) individually doesn't converge, since there's no enumerable list of "ways to leave /create" that's guaranteed complete (browser back would've been next).

Changed approach in aa22a35c: the party seed no longer lives in the shared `retryPrompt` state at all. It's now a ref (`partySeedRef`), read exactly once by the `CreatePage` render that follows Build a game, then unconditionally nulled by a no-deps effect that runs after every commit. No exit path can leak it because nothing after that first render ever reads it again — structural, not enumerated. Reverted the Home/Create `retryPrompt`-clearing from the previous commit since it's no longer needed and was touching unrelated Studio-retry behavior outside this PR's scope.

Added a regression test for your exact repro: Build a game → Up chevron → reopen `/create` via the generic header Create → composer is blank.

Confirmed clean: typecheck, lint, comment-prose (exactly at baseline), full test suite, build.

@codex review

---

gtanczyk:
Real bug, fixed in 1a9ab7bb: `??` checked `retryPrompt` before `partySeedRef.current`, so a stale retained Studio "try this idea again" concept would win over the fresh, one-shot party seed — exactly backwards.

Extracted the precedence into `resolveCreateInitialPrompt` in its own module (App.tsx can't export a non-component without tripping `react-refresh/only-export-components`), flipped the order so the party seed wins, and unit-tested it directly rather than standing up a full Studio-retry integration flow just to exercise a two-line function.

Confirmed clean: typecheck, lint, comment-prose, full test suite, build.

@codex review

---

PR #885, opened 2026-08-18
Fix footer flash during catalog loading

gtanczyk:
## What changed

Keep the home page footer below the fold while the catalog is settling, including the recommendation, featured-pool, auth, and creator-shelf loading gates.

## Why

The footer mounted immediately while the home catalog only contained its small loading state, so it briefly appeared in the viewport before catalog cards arrived and pushed it down.

The catalog now applies a pending layout class and reserves one viewport of height during those states. The reservation is released as soon as the grid is ready.

## Validation

- `npm run type-check`
- `npm run lint`
- `npm run test`
- `npm run build`
- Browser verification at 1440×900 and 390×844, including no horizontal overflow on mobile


gtanczyk:
@codex

PR #814, opened 2026-08-14
Stop showing two controls for the same builder handoff

gtanczyk:
While a platform build is in progress, the composer showed both a
"Switch to your own agent" badge (arm-then-confirm) and a "STOP"
button side by side — both ultimately call the same handoff, but one
looked like a slower, separate action. Hide the badge while STOP is
visible; it reappears once the stop request is in flight to show its
pending state, same as before.

PR #886, opened 2026-08-18
Fix confusing STOP/switch-agent messaging in the build composer

gtanczyk:
## Summary

Reported from a screenshot: while a platform build was running, the Studio composer showed a "Przełącz na swojego agenta" badge and a "STOP" button side by side — both trigger the same builder handoff, but nothing indicated they were the same action.

- Hide the arm-then-confirm "switch to your own agent" badge whenever STOP is already offering the same action; it reappears once the stop request is in flight, showing its pending state.
- Make STOP an icon-only round button (new `stop` PixelIcon) so it matches the icon-only send button instead of being the one text-labeled control in the toolbar. Dropped the now-unused `statusView.feedback.stop` string.
- Stopping/switching an agent is confirmed by the next status poll, not by the request itself, so a creator could be stuck on "waiting for the current agent to acknowledge the stop request" forever if that confirmation never lands. After 20s pending, the badge now shows "taking longer than usual" with a retry button that re-sends the handoff request and restarts the wait.

## Test plan

- [x] `npx tsc --noEmit -p .` in `apps/web`
- [x] `npx vitest run` in `apps/web` — 1338 tests passing, including new coverage for the hidden-duplicate-control behavior and the stale-retry flow

---
_Generated by [Claude Code](https://claude.ai/code/session_01PLWuTjkKrRsVVUnde4ADs3)_


gtanczyk:
$(cat <<'EOF'
Closing as a duplicate — this exact work (same title, same behavior) already shipped to `master` on 2026-08-14 via #822 (a continuation of #814), from a different session working the same report. `master` already has:

- The badge hidden while STOP covers the same handoff
- STOP as an icon-only button with the `stop` PixelIcon
- The 20s stale-timeout retry on the pending handoff badge

Diffing this branch against current `master` shows no functional difference in the touched logic — only comment wording. Rebasing this branch would produce an effectively empty diff on top of unrelated feature work `master` has since gained (attachments, sketches, allow-sending-while-building), so there's nothing to land here.
EOF
)

---

PR #890, opened 2026-08-18
fix(studio): show BYOCA switch on rounds with no committed builder

gtanczyk:
## Summary
- `canOfferSelfHandoff` (`SubmissionStatusView.tsx`) required `status.builder === 'platform'` exactly.
- The API only echoes `builder` when the Firestore record has it set (`submissions.ts:2758-2759`); a fresh queued round (e.g. opened via `code_surface_opened`, before any message is sent) has no `builder` field, so `status.builder` is `undefined`.
- With `canChooseBuilder` also excluding `queued` state, the creator had **no** UI path to pick BYOCA (self) for such a round — reported live on `transport-tycoon-remake` (job 1000081, ownerUid `bot:grok`).
- Fix loosens the check to only exclude an already-`self` round, letting `undefined` through (it resolves to `platform` by default server-side anyway via `builderOf()`).

## Test plan
- [x] `npx vitest run apps/web/src/SubmissionStatusView.test.ts` — 73/73 pass
- [ ] Verify in prod: reopen transport-tycoon-remake Studio, confirm switch-to-self control now renders on the queued round

### DIFF
--- a/apps/api/src/submissions.test.ts
+++ b/apps/api/src/submissions.test.ts
@@ -1441,6 +1441,43 @@ describe('submission routes', () => {
     await app.close();
   });
 
+  it('platform→self handoff on a never-dispatched round resumes immediately, not pending', async () => {
+    // A never-dispatched round has no agent to ack a stop request.
+    const { githubClient } = createGithubClientStub({ issueNumber: 77 });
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #891, opened 2026-08-18
Pass creator reference images to chat agent for context

gtanczyk:
## Summary
Enable the chat agent to receive and process reference images that creators attach to their submissions, allowing the AI model to analyze visual context when providing feedback.

## Key Changes
- **Modified `storeCreatorReferenceImages()`** to return both storage IDs and validated image data (`ChatAgentImage[]`) instead of just IDs, avoiding redundant decoding
- **Updated `ChatAgentRequest` interface** to accept an optional `images` field containing validated PNG images with metadata
- **Added `ChatAgentImage` type** to define the structure of images passed to the chat agent (base64 data + mediaType)
- **Modified `VertexStudioChatAgent.decide()`** to attach reference images only to the current user turn (not history), using the genaicode `image()` and `user()` helpers to construct multimodal prompts
- **Updated submission routes** (two locations) to extract validated images from storage and pass them to the chat agent
- **Added test coverage** verifying images attach to the live turn only and that turns without images don't include an empty images field

## Implementation Details
- Images are validated and stored once, then the base64 data is passed directly to the chat agent without re-decoding
- Only the current user message includes images; historical turns remain text-only
- When no images are present, the user turn is sent as a plain string rather than including an empty images array
- All reference images are validated as PNG format before being passed to the agent

https://claude.ai/code/session_01EcmXFquw9NWNYgE5ub9YEd

### DIFF
--- a/apps/api/src/chat-agent.test.ts
+++ b/apps/api/src/chat-agent.test.ts
@@ -54,6 +54,9 @@ function contextText(request: GenerationRequest): string {
 function liveMessageText(request: GenerationRequest): string {
   return [...request.prompt].reverse().find((item) => item.type === 'user')?.text ?? '';
 }
+function liveMessageImages(request: GenerationRequest) {
+  return [...request.prompt].reverse().find((item) => item.type === 'user')?.images ?? [];
+}
 
 describe('VertexStudioChatAgent', () => {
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #899, opened 2026-08-19
fix(game-seed, source-delivery): stop silent seed crash + silent typecheck skip

gtanczyk:
## Summary

Root-caused from a stuck build (`pinnacle-fairways-championship`, 2026-08-19): the round never landed a first draft, and nothing in the Studio thread said why. Two independent silent failures compounded:

- **`VertexGameSeeder.pickReferences`** (`game-seed.ts`) capped `maxOutputTokens` at 512 while forcing `thinking: 'low'` — both draw from the same budget. A verbose thinking pass could burn the whole 512 and leave an empty answer, and `JSON.parse('')` threw `SyntaxError: Unexpected end of JSON input` uncaught, taking the whole seed down instead of the pick alone failing open (per this module's own "fail-open, always" contract). Fix: bump the pick call's budget to 2048 (still tiny next to the 65,536 generate calls) and catch the parse failure explicitly, logging it distinctly from a generic seed failure.

- **`source-delivery.ts`**: when typecheck preflight is bypassed after hitting the refusal cap, or skipped because its time budget ran out, the only record was a `log.warn` — `roundTypecheckPreflightBypassErrors` is written to the store but nothing ever reads it back, so a delivery could go out unverified with zero signal to the creator. Fix: both paths now also post a `kind: 'blocked'` build-thread event via the existing `appendBuildEvent`/`BuildEvent` pipeline — the web client already renders `'blocked'` with its own icon (`SubmissionStatusView.tsx`), so this needed no client change, just wiring the write.

## Test plan
- [x] `npx vitest run game-seed --dir apps/api` — 45/45, including new regression test for an empty pick response
- [x] `npx vitest run source-delivery --dir apps/api` — 13/13, including new assertion that the bypass path emits a `blocked` thread event
- [x] `npx vitest run --dir apps/api` — full suite, 3285/3285
- [x] `npx tsc --noEmit -p apps/api`, `eslint`, `prettier --check` on all touched files

### DIFF
--- a/apps/api/src/game-seed.test.ts
+++ b/apps/api/src/game-seed.test.ts
@@ -498,6 +498,19 @@ describe('VertexGameSeeder', () => {
     expect(await seeder.seed(request)).toBeNull();
   });
 
+  it('treats an empty pick response as no references, not a crash', async () => {
+    // Regression: seen in prod 2026-08-19 — a verbose 'low' thinking pass spent the whole
+    // maxOutputTokens budget on thinking and left nothing for the answer. The empty text
+    // used to reach JSON.parse('') uncaught, throwing "Unexpected end of JSON input" and
+    // taking the whole seed down instead of the pick alone failing open.
+    const seeder = new VertexGameSeeder({
+      context: stubContext(),
+      client: stubClient([{ text: '' }, { text: GOOD_DRAFT }]),
+    });
+
+    await expect(seeder.seed(request)).resolves.toBeNull();
+  });
+
   it('returns null when the draft is not a usable game', async () => {
     const seeder = new VertexGameSeeder({
       context: stubContext(),
--- a/apps/api/src/game-seed.test.ts
+++ b/apps/api/src/game-seed.test.ts
@@ -499,10 +499,7 @@ describe('VertexGameSeeder', () => {
   });
 
   it('treats an empty pick response as no references, not a crash', async () => {
-    // Regression: seen in prod 2026-08-19 — a verbose 'low' thinking pass spent the whole
-    // maxOutputTokens budget on thinking and left nothing for the answer. The empty text
-    // used to reach JSON.parse('') uncaught, throwing "Unexpected end of JSON input" and
-    // taking the whole seed down instead of the pick alone failing open.
+    // Regression: empty pick text used to throw JSON.parse('') uncaught.
     const seeder = new VertexGameSeeder({
       context: stubContext(),
       client: stubClient([{ text: '' }, { text: GOOD_DRAFT }]),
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex

gtanczyk:
@copilot-pull-request-reviewer re: \`maxItems: String(this.references)\` (game-seed.ts:492) — this is intentional, not a bug. \`@google/genai\`'s own \`Schema\` type (\`node_modules/@google/genai/dist/genai.d.ts\`) declares \`maxItems\`, \`minItems\`, \`maxLength\`, \`minLength\`, \`maxProperties\`, \`minProperties\` all as \`string\`, while true double fields (\`minimum\`/\`maximum\`) are \`number\` — the standard protobuf-JSON int64 convention (int64 bounds serialize as strings to avoid float64 precision loss), not a generic-JSON-Schema field. The SDK passes \`responseSchema\` straight through with no client-side transform, so matching its declared type is what the wire format expects. Keeping it as-is.

The 4 comment-prose findings were correct and are fixed as of 7695ce5c8.

gtanczyk:
@codex review

gtanczyk:
@chatgpt-codex-connector all three P2s confirmed and fixed in 4f05bd4d3:
- Deferred both blocked-event posts until after \`putCandidateSources\` succeeds, so a delivery that fails to store no longer claims it shipped.
- Wired a \`Translator\` through \`reportDeliveryEvent\` (same lazy-default pattern as agent-channel.ts) so the event gets \`textLocalized\`/\`locale\` like every other build-thread event.
- A clean delivery that clears \`roundTypecheckPreflightBypassErrors\` now also posts a \`milestone\` resolution event, so \`events[0]\` doesn't keep the stale warning as the headline.

Added regression tests for all three (localization, ordering-on-store-failure, resolution event).

gtanczyk:
@codex review

gtanczyk:
@chatgpt-codex-connector both confirmed and fixed in 722f87dd4:
- \`check.skipped\` (no_kit/timeout/checker_failed) is now excluded from clearing the bypass state / posting the "now passes" milestone — only a genuine unskipped pass resolves it.
- The deferred event-post loop is now wrapped in try/catch and logs on failure instead of rejecting \`deliver()\` after the version is already stored.

Regression tests added for both.

gtanczyk:
@codex review

gtanczyk:
@chatgpt-codex-connector confirmed and fixed in 7158892ed. Added \`preserveEnded\` to \`appendBuildEvent\` (Store interface + both backends), matching \`touchLastAgentSignalAt\`'s existing pattern:
- Both new system-generated events (blocked/milestone) always pass \`preserveEnded: true\` — a system notice is never proof the agent resumed.
- Went a bit further than the single line you flagged: the pre-existing end-of-\`deliver()\` \`touchLastAgentSignalAt\` call had the exact same bug for *every* creator manual delivery, not just the new notices — now keyed off \`actor === 'creator'\`.

Regression test added (marks the round ended, delivers as \`actor: 'creator'\`, asserts \`agentEndedAt\` survives).

gtanczyk:
@codex review

PR #920, opened 2026-08-20
Issue OAuth challenge at MCP initialize, not tools/call

gtanczyk:
## Summary

Move the MCP OAuth 401 challenge from the first `tools/call` to the `initialize` handshake. This ensures browser-capable clients receive the `WWW-Authenticate` header at the correct point in the protocol flow, allowing them to discover and initiate OAuth sign-in before attempting tool calls.

## Problem

Previously, the OAuth challenge was deferred to the first `tools/call`, which meant credentialed clients worked fine but anonymous visitors received a 200 response at `initialize`. Per RFC 9728, a 200 handshake signals "no auth needed," so clients never probed the protected-resource metadata document and had no way to discover the authorization endpoint. This broke browser-capable clients (Claude connectors, Cursor desktop) that expected to find a sign-in link.

## Changes

- **`mcp-oauth-metadata.ts`**: Modified `shouldIssueMcpOAuthChallenge()` to issue the challenge at `initialize` when no Bearer token is present, while keeping `tools/list` and other read-only methods anonymous
- **`mcp-oauth-metadata.ts`**: Updated `mcpMethodAllowsAnonymousHandshake()` to remove `initialize` from the anonymous methods list
- **`mcp-oauth-metadata.ts`**: Clarified comments explaining that the challenge must happen at `initialize` for OAuth discovery to work
- **New test file `mcp-oauth-handshake.test.ts`**: Added comprehensive tests for the handshake behavior:
  - Verifies `initialize` without credentials returns 401 with `WWW-Authenticate` header
  - Confirms `initialize` with Bearer token returns 200
  - Ensures `tools/list` remains accessible without credentials
- **Updated existing tests**: Added `authorization: 'Bearer handshake'` headers to all `initialize` calls across test files to reflect the new requirement

## Implementation Details

- Any Bearer token shape (including throwaway values) clears the handshake challenge; actual token validation still happens per-call
- All existing credentialed lanes already present a Bearer at `initialize` (creator keys in headers, vaulted tokens from brief, Copilot secrets), so no production code paths are broken
- The change is backward-compatible for all real clients; only the keyless-opener lane (already retired) would have been affected
- `tools/list`, `ping`, and `notifications/*` remain anonymous to keep the tool surface readable as a directory

https://claude.ai/code/session_01SQPm5imykfyW283rvvsegs

### DIFF
--- a/.claude/skills/byoca-mcp/SKILL.md
+++ b/.claude/skills/byoca-mcp/SKILL.md
@@ -849,6 +849,44 @@ cache), so a resume cannot keep showing “finished this round” next to live p
   MCP tools refuse an already-issued key with a reconnect instruction. Do not restore a
   per-game compatibility path in UI, API routes, or MCP tool handling.
 - An opener is checked by `start`; later calls use the returned round-scoped `sessionKey`.
+
+### The handshake carries the OAuth challenge, not the first `tools/call`
+
+`shouldIssueMcpOAuthChallenge` (`apps/api/src/mcp-oauth-metadata.ts`) answers a
+credential-less `initialize` with the 401 + `WWW-Authenticate` challenge. That 401 is the
+**only** in-band signal that starts OAuth discovery (RFC 9728 / the MCP auth flow): a
+client that handshakes to a clean 200 concludes the server needs no auth, never probes the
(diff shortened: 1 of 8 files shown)


gtanczyk:
@codex

PR #933, opened 2026-08-21
Improve GameKit touch UI hiding for hybrid pointer devices

gtanczyk:
## Summary
Updated the CSS media query logic for hiding GameKit touch UI chrome to better handle hybrid mouse+touchscreen devices while preserving the touch interface on actual mobile phones.

## Key Changes
- Changed media query from `@media not all and (any-pointer:coarse)` to `@media (hover:hover),(pointer:fine)` to more accurately detect devices with fine/hoverable primary pointers
- Simplified the selector from `.gamekit-touch:not(:has(.gamekit-touch-pad))` to `.gamekit-touch` since the new media query is more precise
- Updated test expectations to match the new media query and selector logic

## Implementation Details
The previous approach used a negative selector that only hid touch UI on devices without coarse pointers, which incorrectly showed the full phone pad+button overlay on hybrid devices (mouse+touchscreen). These devices report `any-pointer:coarse` due to having both input types, but users primarily interact via mouse/keyboard on desktop-sized playfields.

The new approach uses positive selectors for devices with hover capability or fine pointers, which correctly identifies:
- Desktop computers with mice (hover:hover)
- Hybrid touchscreen devices used with mouse (pointer:fine)
- While preserving the touch UI on actual phones (hover:none + pointer:coarse)

This provides a better user experience on hybrid devices by hiding the touch-optimized overlay when the primary interaction method is mouse/keyboard.

https://claude.ai/code/session_01PVsxn4fjeuRHgvy5FVAw9d

### DIFF
--- a/apps/web/src/gamePlayer.test.ts
+++ b/apps/web/src/gamePlayer.test.ts
@@ -9,9 +9,9 @@ describe('embedGameHtml', () => {
     // Style + script land inside <head> so rAF patches precede game scripts.
     expect(out).toContain('<style id="gdpl-embed">');
     expect(out).toContain('#game-title,#game-desc,.game-controls,.hint{display:none!important}');
-    // Desktop mouse: hide buttons-only GameKit chrome (pointer-native tactics UIs).
-    expect(out).toContain('@media not all and (any-pointer:coarse)');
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #949, opened 2026-08-22
Phase 1: stop hand-mirroring API response shapes in the web

gtanczyk:
Second of three PRs finishing Phase 1. The web declared its own structural copy
of every API response it reads — `SubmissionStatus` alone was a **~70-field
mirror policed only by comments**. This publishes those shapes once in
`packages/contract` and deletes both copies.

**No behaviour changes.** Every edit is a type move; the runtime reads are untouched.

## Impact

**−1028/+773 across 27 files, and every consuming file shrank:**

| File | Before | After |
|---|---:|---:|
| `apps/api/src/submission-status.ts` | 505 | **234** |
| `apps/web/src/submissionApi.ts` | 541 | **350** |
| `apps/api/src/telemetry-health.ts` | 470 | **364** |
| `apps/api/src/github-client.ts` | 2084 | **2003** |
| `apps/api/src/creator-studio.ts` | 535 | **478** |
| `apps/web/src/studioApi.ts` | 516 | **474** |
| `apps/web/src/catalog.ts` | 302 | **255** |
| `apps/web/src/healthApi.ts` | 274 | **244** |

Comment-prose debt drops by **4,015 words** — the duplicated docs went with the
duplicated types. The new contract modules carry their own documentation at
**zero** prose debt, so this is a genuine retirement, not a relocation.

New modules, each with a companion test: `submission-status`, `catalog-entry`,
`studio-status`, `game-health`, plus `GATE_PROGRESS_STAGES` / `GateProgress`
folded into the existing `gate-progress`.

## Naming

`SubmissionStatus` meant two different things — the whole response body on the
web, a bare alias for `SubmissionState` on the api. The response object is
published as **`SubmissionStatusResponse`**; the web keeps `SubmissionStatus` as
a local alias, so no consumer changed.

## Divergences the two copies had drifted into

Consolidating forces a decision at each one. These are the interesting part of
the diff:

| Divergence | Resolution |
|---|---|
| `BuildEvent`/`CreatorRevision` carry `textLocalized`/`locale` on the api but not on the wire (`localizeEvents` strips them) | Modelled honestly as `StoredBuildEvent`/`StoredCreatorRevision` extending the wire types, rather than one type that is wrong on one side |
| `progress.revisions` optional on web, required on api | **Required** — both producers always set it. Web read sites already used `progress?.revisions ?? []`, so nothing changes at runtime. ~27 test fixtures gained `revisions: []` |
| `RecentBuild.failedStage` `string` on web, `GateProgressStage` on api | **Typed on both.** The web renders a bar frozen at that stage, so the loose type admitted values outside the 12-stage vocabulary |
| `gateProgress.stage` `string` on **both** sides, though the api had `GateProgressStage` available and assigns a real one | Now the shared `GateProgress` type — both sides were wrong together |
| `CatalogEntry.touch` required-and-nullable on web, optional-and-non-nullable on api | Contract accepts **both** (`touch?: CatalogTouch \| null`). The api genuinely omits the key on the SPEC-only GraphQL fallback, so a single strict choice would break one side |
| `CreatorStudioGame.lastKnownStatus` `string \| null` on api, `SubmissionState \| null` on web | Web was right; the api emits a `SubmissionState` and its type was merely loose |
| `StudioHealthResponse.gamesTruncated`/`totalGames` optional on web, required on api | **Required** — the api always sets them, so the web stops defaulting away a field it can rely on |

The web also gains `SubmissionPublishedResponse`, so it can finally know a
published build always carries its slug instead of non-null-asserting.

## Findings NOT acted on here

Each is a behaviour question, not a type question, so none is changed in a
consolidation PR — but all three are real:

- **`BuildMediaItem.source: 'branch'` is unreachable.** The only producer
  hardcodes `'channel'`, and `buildMediaUrl` maps `'branch'` to
  `/api/submissions/:token/media/:ref` — **a route that does not exist**. Anything
  emitting it would 404 every ``.
- **`BuildProgress.note` is never populated.** Three web components read it as
  the "agent says" fallback, but `parseProgressNote`'s only non-test reference is
  its own declaration. That fallback has silently never fired.
- **`fetchStudioScorecards` drops `truncated` and `totalGames`**, so a creator
  whose scorecard list was truncated is never told.

## Verification

- `npm run type-check` — clean across all 6 workspaces
- `npm run lint` — clean (eslint + comment-prose + module-size)
- Tests: api 213 files/3412 tests, web 187/1614, contract 37/92 — all passing

---


gtanczyk:
@codex review

---

PR #951, opened 2026-08-22
Phase 1: one declaration per byte budget

gtanczyk:
**The fourth and last exit-criterion item.** Every duplicated limit now has exactly
one declaration, in `packages/contract`, each pinned by a test.

Four new modules, each a cohesive group rather than a bag of numbers:

| Module | Contents |
|---|---|
| `zone-contract` | `TICK_HZ_VALUES`, `MAX_PLAYERS_PER_ZONE`, `MAX_STATE_BYTES`, `RESERVED_EVENT_KINDS`, `ZoneEvent` |
| `socket-limits` | `MAX_SOCKET_FRAME_BYTES`, `MAX_SOCKET_FRAMES_PER_SECOND`, `DEFAULT_MAX_SOCKETS_PER_IP` |
| `game-limits` | `MAX_GAME_SAVE_BYTES`, `MAX_TITLE_LENGTH`, `MAX_MULTIPLAYER_SLOTS`, `MAX_SHOT_BYTES` |
| `world-limits` | `MAX_WORLD_FIELDS`, `MAX_WORLD_KEY_LENGTH`, `MAX_WORLD_ENTRY_BYTES` |

## The zone-core comment was stale — that's why this is safe

`packages/zone-core/src/contract.ts` justified keeping its own copy by analogy:

> Keeping a copy rather than importing one is the same call `mp.ts` and
> `apps/web/src/mp/protocol.ts` already make…

**That precedent no longer exists.** `mp/protocol.ts` imports its vocabulary *and*
its protocol version from the contract today (#948).

The **games-repo half** of that argument is untouched and still true: that repo
deploys independently, cannot be imported from, and is held by `contract.test.ts`
pinning the encoding against vectors copied from its own test. **Those 96 zone-core
tests pass unchanged here** — that's the evidence the deterministic contract did
not move.

Net effect on the cross-repo mirror: **three copies → two checked halves**, exactly
what N2 of the north-star plan asks for.

## Frozen files: `submissions.ts` shrinks

**6029 → 6028.** Its existing contract import had room for `MAX_SHOT_BYTES` under
the 120-char limit. Naming the repeated base64 expansion (`MAX_SHOT_BASE64_CHARS`)
then removed **three copies** of that arithmetic and shortened the lines enough that
the file ends up a line *below* where it started — which also lets its ceiling
ratchet back **down**, undoing the +1 reseal #948 needed for a pre-existing prettier
drift.

## Two couplings kept rather than flattened

- **`MAX_SLOTS` was `SLOT_COLORS.length`.** It now reads the shared seat count, so a
  colour added or removed would no longer change it. A new test asserts `SLOT_COLORS`
  still has exactly `MAX_SLOTS` entries — the property that actually mattered.
- **`telemetry.ts`** validated `slots` against a bare `.max(8)`. Same seat count, now named.

## Build ordering, which this change forces

`zone-core` gained `@gamedevpl/contract` as a dependency, so `build:packages` had to
stop building `zone-core` first. **Verified by deleting both dists and building from
clean — the old order fails outright.** `apps/world/Dockerfile` already built contract
first, so it needed nothing.

## Deliberately NOT merged

Several constants share a value across unrelated domains — 64 KiB in
`editor-contract.ts`, `store.ts` and `example-files.ts`; 32 KiB in `kit-files.ts` and
`knowledge-search.ts`. Same number, different reasons. Merging those would create a
coupling bug rather than remove drift.

## One comment corrected, no behaviour change

`apps/web/src/zone/protocol.ts` said `MAX_INPUTS_PER_SECOND` was *"inputs a client may
send per second before the host disconnects it"*. The host refuses the **frame** at
**40** (`MAX_SOCKET_FRAMES_PER_SECOND`, enforced by `FrameLimiter` in
`apps/world/src/app.ts`), not 30 — and refusing a frame is not disconnecting. The 30 is
this client's own budget, used as `INPUT_BUDGET` in `apps/web/src/zone.ts`. The two are
genuinely different constants and stay separate; only the sentence was wrong.

## Verification

- `npm run type-check` — clean across all workspaces
- `npm run lint` — clean (eslint + comment-prose + module-size)
- `npm run build:packages` from deleted dists — proves the new ordering
- Tests: api 214/3418, web 188/1615, **zone-core 5/96**, contract 42/99, world 2/20

---


gtanczyk:
@codex review

---

gtanczyk:
Both findings confirmed and fixed in `ba7087e` (I'd normally reply on each thread directly, but the review-threads API is rate-limited right now — posting here instead).

**Title length**: `apps/api/src/submissions.ts`'s create-round schema and `apps/api/src/refine.ts`'s refine-request schema each validated `.max(80, ...)` as a bare literal, independent of the shared constant. Both now derive from it — `submissions.ts` gains a `TITLE_TOO_LONG_MSG` constant so the validator line stays under 120 chars rather than reflowing across several lines for one call.

**Save cap**: `apps/api/src/store.ts` independently re-declared `MAX_GAME_SAVE_BYTES = 32 * 1024`, and `apps/api/src/game-saves.ts`'s authoritative save-size check imported that copy. `store.ts` now re-exports the contract's constant instead, so `game-saves.ts`'s existing import keeps resolving with no change needed there.

Both are exactly the shape this PR exists to prevent — I'd wired the web-side display value in each case but left the API's real enforcement independent. Good catches.

Both frozen files (`store.ts`, `submissions.ts`) grew slightly for genuine reasons (an import name, a validator wired to it) and were resealed scoped to those two paths per the module-size policy `b659c9d` made explicit. All suites green after the fix and after a subsequent unrelated master merge: api 214/3418, web 189/1617.

---

PR #959, opened 2026-08-22
Phase 2 Wave 4: grow fakeFirestore's query support

gtanczyk:
## Summary

Shared prerequisite named in the Wave 4 plan, not a slice move. `fakeFirestore()` supported `==` and `!=` only; measuring every `FirestoreStore` method still in `store.ts` against it found the gap splits cleanly:

- **`orderBy`**: social, notifications, submission, buildLog, review (13 call sites)
- **range operators (`<=`, `>=`)**: identity (`listAccountsDueForDeletion`), dispatch (`listSeedOutcomesSince`)
- **`in` + `startAfter` cursor pagination**: contribution (`listSuggestions`, `listProposals`)
- **`select()`**: buildLog's shot/preview listings — worth modeling honestly since it's a real payload-size optimization, not a projection nicety

Adding all of it in one PR, before any of those slices move, means their eventual parity suites don't each rediscover a piece of the gap and grow the fake piecemeal across ten PRs.

## Changes

`fake-firestore.ts`:
- `makeQuery`/`makeCollection`: `orderBy(field, direction)`, `select(...fields)`, `startAfter(cursor)`, and `where()` now accepts `<`, `<=`, `>`, `>=`, `in` alongside `==`/`!=`.
- **No `orderBy` means Map insertion order** (matching `InMemoryStore`'s own Maps), not real Firestore's `__name__` fallback — documented rather than silently diverging, since nothing in this codebase reads meaning into unordered-query row order, only that a `startAfter`-paginated read is stable and exhaustive.
- `startAfter` matches its cursor **by id, not by (path, id)**: every real `startAfter` call in this codebase queries a single flat collection, where an id is already unique, so a collection-scoped match is correct and avoids threading `path` onto returned doc snapshots for a case that never occurs.
- `select()` actually restricts returned fields rather than passing everything through: `BuildShotSummary`/`BuildPreviewSummary` get cast straight from `doc.data()`, so a fake that ignored `select()` would hide a bug where a caller reads a field the real optimization withholds (concretely, a preview's multi-hundred-KB `data` payload).

## Tests

6 new cases in `store-firestore.test.ts`'s "the fake itself" block, each named for the real call site it backs: `orderBy` asc/desc, each range operator, `in`, `select`, `startAfter`'s full-pagination guarantee, and the exact `select().orderBy().limit()` chain `listBuildShots` uses.

## Comment-prose note

Three lines I wrote fresh (the range-operator docstring, the no-`orderBy` ordering note, the `startAfter` id-matching note) were trimmed to compliant one-liners before checking — came back to exactly `fake-firestore.ts`'s existing 328-word baseline, no reseal needed at all.

## What's next

Unblocks: social, notifications, submission, buildLog, review, identity, dispatch, contribution — eight of the remaining ten slices. Full plan: `docs/phase2-store-split-execution-plan.md` in the private ops repo.

## Test plan

- [x] `npx tsc --noEmit` clean (apps/api)
- [x] Full apps/api suite: 217 files, 3440 tests passing (6 new)
- [x] Full apps/web suite: 189 files, 1620 tests passing
- [x] `npm run lint` (eslint + comment-prose + module-size ratchets) clean repo-wide

---

### DIFF
--- a/apps/api/src/store-firestore.test.ts
+++ b/apps/api/src/store-firestore.test.ts
@@ -495,6 +495,86 @@ describe('the fake itself', () => {
         .set({ email: undefined } as never),
     ).rejects.toThrow(/Cannot use "undefined" as a Firestore value/);
   });
+
+  it('orders by a field, ascending and descending', async () => {
+    const { db } = fakeFirestore();
+    const col = db.collection('items');
+    await col.doc('a').set({ n: 3 });
+    await col.doc('b').set({ n: 1 });
+    await col.doc('c').set({ n: 2 });
+
+    const asc = await col.orderBy('n', 'asc').get();
+    expect(asc.docs.map((doc) => doc.id)).toEqual(['b', 'c', 'a']);
+
+    const desc = await col.orderBy('n', 'desc').get();
+    expect(desc.docs.map((doc) => doc.id)).toEqual(['a', 'c', 'b']);
+  });
+
+  it('supports range operators, the way listAccountsDueForDeletion needs', async () => {
+    const { db } = fakeFirestore();
+    const col = db.collection('items');
+    await col.doc('a').set({ n: 1 });
+    await col.doc('b').set({ n: 2 });
+    await col.doc('c').set({ n: 3 });
+
+    expect((await col.where('n', '<', 2).get()).docs.map((doc) => doc.id)).toEqual(['a']);
+    expect((await col.where('n', '<=', 2).get()).docs.map((doc) => doc.id)).toEqual(['a', 'b']);
+    expect((await col.where('n', '>', 2).get()).docs.map((doc) => doc.id)).toEqual(['c']);
+    expect((await col.where('n', '>=', 2).get()).docs.map((doc) => doc.id)).toEqual(['b', 'c']);
+  });
+
+  it('supports "in", the way listSuggestions/listProposals filter by status set', async () => {
+    const { db } = fakeFirestore();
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex please review.

---

gtanczyk:
Both confirmed and fixed in 3669082. (Posting as a top-level comment rather than inline replies — the review-threads API is rate-limited for this account right now.)

**Finding 1 (missing-field range matching):** confirmed — `lessThan(x, undefined)` is `false`, so the negated `<=`/`>=` forms came back `true` for any document missing the field, meaning `listAccountsDueForDeletion` would have matched every user, not just scheduled ones. **Finding 2 (dotted paths):** confirmed — `data['seedOutcome.at']` was a literal top-level lookup against a document that actually stores `{ seedOutcome: { at } }`, so `listSeedOutcomesSince`'s query always saw the field as absent.

Fixed both together since they compound: added `getField()` to resolve Firestore-style dotted paths for `where`/`orderBy`/`select`, and `requireFields` tracking so any field touched by an inequality operator or `orderBy` excludes documents where `getField()` returns `undefined` — matching real Firestore's field-existence requirement. `==`/`in` were left alone since an absent field already fails those naturally.

Reproduced both first: reverted the fix, confirmed three new regression tests fail with exactly the symptoms described (missing-field docs matching, `orderBy` not excluding a missing-field doc, dotted path returning every document unordered), then restored the fix and confirmed all three pass.

---

PR #970, opened 2026-08-23
feat(api): assemble fast preview immediately on source delivery

gtanczyk:
### Summary

When an agent delivers sources (`submit_sources` / `sourceDelivery.deliver`), previously the Studio stage waited for the entire 5–8 minute Cloud Build Acceptance/Publish Gate (typecheck → smoke → build → trace → video/MP4 capture → validate → accept → playtest) to finish before `preview.html` / `bundle.html` was written and available to the creator.

This change:
1. Assembles and stores `preview.html` in-process immediately on candidate delivery (`githubClient.getGameSources` + `assembleGameHtml`) and stores it as a derived artifact on `gamesStore`, alongside updating channel `BuildPreview`.
2. Connects `githubClient`, `engineRef`, and `stagedPreviews` to `createSourceDeliveryService` in `submissions.ts` and `creator-code.ts`.
3. Allows creators in Studio to immediately play and iterate on the new build in ~1–2 seconds while the heavy Publish Gate (video capture, certification) runs asynchronously in the background.

### Verification

- `npm run type-check` passed cleanly across all workspaces.
- `npm run lint` passed with 0 errors and 0 warnings.
- `npm run test` passed (including new unit test in `source-delivery.test.ts`).
- `npm run build` passed.

PR #975, opened 2026-08-23
fix(api): retry candidate preview instead of silently skipping busy slot

gtanczyk:
## Summary
- `publishCandidate()` in `apps/api/src/delivery/staged-preview.ts` answered `'skipped'` — with no retry and no log — whenever a background staged-preview assembly for the same round was already running. This let a *delivered* version's fast preview get dropped outright, leaving the Play button gated behind the full publish gate (~11 min) instead of unlocking as soon as the build compiled.
- `publishCandidate` now waits out a busy slot (bounded, 15s, polling at `busyRetryMs`) before giving up, instead of skipping on the first check.
- Every non-`published` outcome (`skipped`, `incomplete`, `too_large`) is now logged via `options.log.warn` — previously silent, matching the module's intentional silence for the *background* staging path but hiding real problems on the *delivered-candidate* path.

## Why
Reported by the owner: a build stuck on "Sprawdzamy tę wersję" / "SPRAWDZANIE" in Studio, with Play blocked on the full validation pipeline (agency check etc.) despite PR #970 ("assemble fast preview immediately on source delivery") having been merged to fix exactly this. Traced to a race: `submit_sources` calls `publishCandidate` while an earlier `stage_source_file`-triggered background assembly (`schedule`/`arm`/`publishNow`) for the same `issueNumber` is still `running`, so the concurrency guard silently drops the real delivered preview and `status.playable` never gets an entry.

## Test plan
- [x] `npx vitest run src/delivery/staged-preview.test.ts` — 38/38 passing, including a new test (`waits out a busy slot instead of skipping a delivered candidate`) that reproduces the race and asserts the candidate eventually publishes once the busy slot frees.
- [x] `npx tsc --noEmit` in `apps/api` — no new errors (one pre-existing, unrelated error in `build-changelog.ts` confirmed present on a clean checkout of this branch before these changes).
- [x] `npx eslint` on both changed files — clean.


gtanczyk:
@codex

PR #980, opened 2026-08-23
fix(web): keep last build visible and honest during a new round

gtanczyk:
## Summary

Two related Studio bugs, both caused by state scoped to the wrong lifetime:

- **The stage went blank the moment a new self-build (BYOCA) round opened** on an already-published game — the main preview panel fell back to a "Budowa gry trwa." / "PLANOWANIE" placeholder instead of the last known-good build. Root cause: `apps/web/src/useStageSource.ts`'s fallback fetch of the last published build (`fetchPublishedGame`) was gated on `status.status === 'published'`, which is never true for the *current* round while an improvement round is in flight — even though the game itself already has a live published version. The fetch now triggers on `status.slug` alone and is used as the lowest-priority fallback in the display chain (`preview ?? channel ?? published`), so the last known-good build stays on screen until the new round produces its own.

- **The build-progress pill kept reading "Sprawdziany zaliczone" (checks passed)** for a round that had not delivered anything of its own yet. Root cause: `status.recentBuilds` is scoped to the *slug* (spans every round ever), while a self-build round mints its own `issueNumber` — so `recentBuilds[0]` could still be the *previous* round's green verdict. Exposed the round's own `issueNumber` on `SubmissionStatusResponse` (`packages/contract/src/submission-status.ts`, populated in `apps/api/src/submissions.ts`'s `nativeJobStatus`) so `apps/web/src/buildBarModel.ts` can detect the mismatch and show a distinct "Runda w toku" / "Round in progress" reading instead of trusting a stale verdict.

## Why

Reported by the owner from a live screenshot: a BYOCA round had just opened on a game with several successful prior publishes, and the stage showed the "building" placeholder (hiding the still-playable prior build) under a pill that falsely claimed checks had passed for the new round.

## Test plan
- [x] `npx vitest run src/buildBarModel.test.ts src/useStageSource.test.ts` (apps/web) — new tests added for both fixes, all passing.
- [x] Full `apps/web` suite: 1635/1635 passing.
- [x] `apps/api` `submissions.test.ts`: 210/210 passing.
- [x] `npx tsc --noEmit` clean in `packages/contract`, `apps/api`, `apps/web`.
- [x] `npm run lint` clean (0 errors); raised the small, legitimate module-size ceilings for the touched files.


gtanczyk:
@codex

PR #981, opened 2026-08-23
feat(web): improve search-or-create hero prompt and smart intent card usability

gtanczyk:
### Frontend & UX Improvements
- **Smart Intent Match Card**: Added screenshot poster thumbnail (`matchedPoster`), genre and multiplayer badges, line-clamping (`line-clamp: 2`) for game descriptions/controls, and subtle creator guidance (`smartMatchHint`).
- **Creation Card**: Updated "No Match" state from negative search error framing to positive prompt building guidance.
- **Search/Prompt Placeholder**: Clarified dual intent in Polish and English locale placeholders.
- **Client Debounce & Request Cancellation**: 200ms debounce with `AbortController` cancellation to prevent out-of-order responses.
- **Unit Tests**: Full test coverage for hero prompt matching, badge rendering, and cancellation.

### Semantic Search & Metadata Backend
- **Endpoint**: `GET /api/catalog/search` with rate limiting (`120 req/min`) and in-memory LRU query embedding cache (up to 500 queries).
- **Vertex AI Embeddings**: `VertexEmbeddingService` generating normalized vectors (`text-embedding-005`), with clean failure degradation and logger warnings.
- **In-Memory Vector Index**: `CatalogVectorIndex` with cosine similarity ranking and asynchronous background warm-up on boot.
- **Enrichment Pipeline**: `getOrEnrichCatalogGame` with `gemini-3.5-flash-lite` generating Polish/English taglines, controls, and search keywords, backed by Firestore `catalogEnrichment` collection with `contentHash` cache invalidation.
- **Concurrency & Outage Resilience**: Throttled index rebuilds with 1-minute retry backoff on failure and 10-minute TTL on stale entries, decoupled from the critical request path.


gtanczyk:
@codex

gtanczyk:
@codex

apps/api/src/catalog/embedding-service.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  // Generates pseudo-vector for local development when offline.
+  generateDeterministicFallback(text: string, dimensions = 64): number[] {
+    const vec = new Array<number>(dimensions).fill(0);
+    for (let i = 0; i < text.length; i++) {
+      const code = text.charCodeAt(i);
+      const idx = (code + i * 31) % dimensions;
+      vec[idx] += 1;
+    }
```
gtanczyk: **The silent fallback will produce confidently wrong matches in production.**

This `catch` swallows *every* failure — auth, 429, 5xx, malformed response — and returns a 64-dim character-histogram pseudo-vector, with no logging and no way for the caller to tell a degraded vector from a real one.

Two distinct failure modes:

**1. False positives above the threshold.** Character-frequency vectors converge as text gets longer, so unrelated long strings look similar. I ran this fallback against realistic enriched doc text (title + genre + EN/PL tagline + keywords):

```
query: "a relaxing farming simulator where you grow vegetables and raise
        chickens on a small island"
  vs "Mexico '86 Arcade Football. Sports. Relive the golden age of arcade
      football with lightning passes … football, soccer, sports, arcade,
      retro, multiplayer, pixel, tournament"
  cosine = 0.726
```

That clears the `threshold = 0.65` in `CatalogVectorIndex.findBestMatch`, so during a Vertex outage the hero card confidently announces "Found matching game: Mexico '86 Arcade Football" for a farming prompt. Long descriptive prompts are exactly what this feature targets, and they are the worst case here.

**2. Silent total blackout.** `text-embedding-005` returns 768 dims. If the index is built with real vectors and a later query falls back to 64 dims, `cosineSimilarity` returns `0` on the length mismatch — so search silently returns "no match" forever, indistinguishable from a genuine no-match, with nothing in the logs.

Suggested shape: log the error, let `embedText` reject (or return `null`) so `/api/catalog/search` can answer `{ match: null }` honestly, and keep `generateDeterministicFallback` for tests only. If you do keep a runtime fallback, at minimum record which mode produced the index and refuse to compare across modes.

---

apps/api/src/catalog/catalog-enricher.ts
```diff
@@ -0,0 +1,210 @@
...
+  addTokens(entry.slug);
+  addTokens(entry.title);
+  addTokens(entry.genre);
+  return Array.from(words);
+}
+
+// Generates or retrieves cached AI enrichment for one game.
+export async function getOrEnrichCatalogGame(
+  entry: CatalogGameEntry,
+  specMd: string,
```
gtanczyk, line 44: **Nothing ever calls this, so the enrichment half of the feature is inert as merged.**

`getOrEnrichCatalogGame`, `enrichCatalogEntries`, and `createDefaultEnricherClient` have no production callers — grepping `apps/` and `packages/` finds only `catalog-enricher.test.ts` and the definitions themselves. The only thing wired into a route is `attachCatalogEnrichments` (submissions.ts:5812, 5831), which *reads* from the store.

Since nothing ever writes to the `catalogEnrichment` collection:

- `attachCatalogEnrichments` always returns entries unchanged;
- `tagline` / `shortControls` / `searchKeywords` are always `undefined` in prod, so the new hero-card tagline path, the keyword match rule, and the `catalog.controls: shortControls` branch all dead-fall to the old behaviour;
- `ensureCatalogVectorIndex`'s `docText` collapses to `` `${title}. ${genre}.  ` `` — embedding little more than what a substring match already covers.

Was a batch/backfill job or a publish-time hook meant to land in this PR? As it stands the LLM prompt, the Firestore slice, and the store methods are all reachable only from tests.

---

apps/api/src/submissions.ts
```diff
@@ -5816,6 +5820,68 @@ export async function registerSubmissionRoutes(
...
+
+  // Multimodal semantic vector search across catalog games.
+  app.get('/api/catalog/search', async (request, reply) => {
+    const query =
+      typeof request.query === 'object' && request.query !== null && 'q' in request.query
+        ? String((request.query as { q: unknown }).q || '').trim()
+        : '';
+    if (!query || query.length < 2) {
+      return reply.send({ match: null, score: 0 });
+    }
```
gtanczyk: **Unauthenticated endpoint that bills a paid API per request, with no rate limit.**

Every other route registered in this file carries an explicit budget, e.g.

```ts
{ config: { rateLimit: { max: 60, timeWindow: '1 hour' } } }
```

This one has none. Combined with the caller side — `HeroPromptSection.tsx:273` fires on a 200 ms debounce with no minimum quiet period beyond that — a single visitor typing one sentence produces several Vertex embedding calls, and an anonymous client can drive Vertex spend directly.

Two things worth adding:

1. A `rateLimit` config matching the convention above.
2. A small LRU/TTL cache on the query embedding. Hero prompts repeat heavily across visitors ("tetris", "football", "snake"), so caching by normalised query would cut most of the traffic without changing behaviour.

Also: there's no server-side test for this route. The only coverage is `HeroPromptSection.test.tsx` with a mocked `fetch`, which exercises the client contract but not the handler's query parsing, threshold, or error path.

---

apps/api/src/submissions.ts
```diff
@@ -5816,6 +5820,68 @@ export async function registerSubmissionRoutes(
...
+            searchKeywords: entry.searchKeywords,
+            embedding: vec,
+          });
+        }
+      }
+    } catch {
+      // Non-blocking index sync
+    }
+  }
+
```
gtanczyk: **Build-once index: never refreshed, no in-flight dedup, and the cold build runs inside a user request.**

Three separate issues in this guard:

**Staleness.** `if (catalogVectorIndex.size() > 0) return` means the index is built once per process and never updated. A game published after the first `/api/catalog/search` call is invisible to semantic search until the instance restarts — and enrichment updates (new tagline/keywords) never propagate either.

**Thundering herd.** There's no memoised in-flight promise. N concurrent requests arriving at a cold instance each see `size() === 0` and each run the full loop, so a cold start with 4 simultaneous searches embeds the whole catalog 4×. Storing the promise fixes it:

```ts
let indexBuild: Promise<void> | null = null;
function ensureCatalogVectorIndex() {
  return (indexBuild ??= buildIndex().catch((e) => { indexBuild = null; throw e; }));
}
```

**Latency.** The loop `await`s one embedding per entry sequentially, on the request path. The first search on each instance pays `catalogSize × embeddingLatency` before it can answer. Worth batching (Vertex `:predict` accepts multiple `instances`) and moving the build off the request path — a warm-up on boot plus a TTL refresh would address staleness and latency together.

Note this is also per-instance state, so on Cloud Run every instance pays its own cold build.

---

apps/web/src/HeroPromptSection.tsx
```diff
@@ -248,7 +260,50 @@ export function HeroPromptSection({
...
+              });
+              return;
+            }
+          }
+          setVectorMatchedGame(null);
+        })
+        .catch(() => {
+          setVectorMatchedGame(null);
+        });
+    }, 200);
```
gtanczyk, line 300: **No request cancellation — stale responses can overwrite newer ones.**

The cleanup clears the debounce timer, but an already-dispatched `fetch` keeps running. Typing past the 200 ms debounce fires request A, then request B; if A is slower than B (very plausible here, since A may be the cold request that triggers `ensureCatalogVectorIndex` and embeds the entire catalog while B hits a warm index), A resolves last and `setVectorMatchedGame` writes A's stale match over B's correct one. The card then shows a game matching text the user has already edited away.

Standard fix — abort in the cleanup:

```ts
const controller = new AbortController();
const handle = setTimeout(() => {
  fetch(`/api/catalog/search?q=${encodeURIComponent(trimmed)}`, { signal: controller.signal })
    // …
    .catch(() => { if (!controller.signal.aborted) setVectorMatchedGame(null); });
}, 200);
return () => { clearTimeout(handle); controller.abort(); };
```

This also stops the abandoned requests from billing Vertex, which ties into the missing rate limit on the endpoint.

---

apps/web/src/HeroPromptSection.tsx
```diff
@@ -50,7 +50,14 @@ function findMatchingGame(query: string, catalog: CatalogEntry[]): CatalogEntry
       return entry;
     }
 
-    // 2. Special aliases
+    // 2. Keyword or search tag match (e.g. from Flash-Lite enrichment)
+    if (
+      entry.searchKeywords?.some((k) => normalized.includes(k.toLowerCase()) || k.toLowerCase().includes(normalized))
+    ) {
+      return entry;
```
gtanczyk, line 60: **This rule is loose enough to hijack the match card, and it runs before the alias rules.**

`normalized` is the whole lowercased query, so both directions are substring tests against LLM-authored tags:

- `normalized.includes(k)` — any query containing a generic tag (`2d`, `ai`, `retro`, `fun`, `pixel`) matches whichever game happens to carry it first in catalog order.
- `k.includes(normalized)` — a two-character query like `ar` matches the keyword `arcade`.

There's no length floor on this path: `catalog-enricher.ts` builds `searchKeywords` from the model with only `.filter(Boolean)`, while the *fallback* path (`extractFallbackKeywords`) does apply `t.length > 2`. So the LLM-sourced keywords — the ones that actually reach this code — are the less constrained of the two.

Because this sits at step 2, ahead of the "mario → plumber" style aliases, a stray generic tag also shadows the hand-tuned aliases below it.

Suggest matching on whole tokens rather than substrings, and applying the same `length > 2` floor the fallback path uses:

```ts
if (entry.searchKeywords?.some((k) => {
  const kw = k.toLowerCase();
  return kw.length > 2 && tokens.includes(kw);
})) {
  return entry;
}
```

Given the enrichment writer isn't wired up yet, this is latent today — but it becomes live the moment the backfill lands.

---

apps/api/src/catalog/embedding-service.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    const vec1 = await service.embedText('arcade football');
+    const vec2 = await service.embedText('arcade football');
+    const vec3 = await service.embedText('space shooter');
+
+    expect(vec1.length).toBeGreaterThan(0);
+    expect(vec1).toEqual(vec2);
+    expect(cosineSimilarity(vec1, vec2)).toBeCloseTo(1);
+    // Distinct texts should have similarity < 1
+    expect(cosineSimilarity(vec1, vec3)).toBeLessThan(1);
+  });
```
gtanczyk, line 76: The test named "when offline" isn't offline — it constructs a real `VertexEmbeddingService` and calls `embedText`, which attempts real `GoogleAuth` credential resolution and a real `fetch` to `aiplatform.googleapis.com`. It passes today only because that attempt fails and falls into the `catch`.

Two consequences: it's a latency/flake risk in CI (metadata-server probes can hang rather than fail fast), and on a machine that *does* have ADC it silently starts testing the live API instead of the fallback, so the assertions no longer mean what the name says.

Call `generateDeterministicFallback` directly — it's already a public method — or stub `fetch`/`GoogleAuth` so the failure path is deliberate rather than incidental.

---

apps/web/src/styles.css
```diff
@@ -990,32 +990,74 @@ a {
...
 
 .matched-desc {
   margin: 0;
-  font-size: 0.9rem;
+  font-size: 0.85rem;
+  line-height: 1.4;
   color: var(--muted);
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
```
gtanczyk, line 1053: Minor: add the standard `line-clamp: 2` alongside `-webkit-line-clamp` — it's the property browsers are standardising on, and some stylelint configs flag the prefixed form when it appears alone.

Also `text-overflow: ellipsis` is a no-op under `display: -webkit-box`; the ellipsis comes from `-webkit-line-clamp` itself. Safe to drop.

---

gtanczyk (commented):
## Review

The PR description covers the hero card, i18n strings, CSS and web tests — but the diff also adds a public API endpoint (`/api/catalog/search`), a Vertex embedding client, an LLM enrichment pipeline, and a new Firestore collection (`catalogEnrichment`). That's a much larger surface than the summary suggests; worth expanding the description so reviewers know a paid-API endpoint is in scope.

All 8 checks are green. Findings below, most severe first — details are in the inline comments.

**1. The embedding fallback produces confidently wrong matches.** `VertexEmbeddingService.embedText` catches every failure and silently returns a 64-dim character-histogram vector. I measured it against realistic doc text: a farming-simulator query scores **0.726** against a *Mexico '86 Arcade Football* document, clearing the `0.65` threshold in `findBestMatch`. During a Vertex outage the hero card will confidently announce the wrong game. The converse case is just as bad — a 768-dim index with a 64-dim query makes `cosineSimilarity` return `0` on the length mismatch, so search silently dies with nothing in the logs.

**2. The enrichment writer isn't wired up.** `getOrEnrichCatalogGame`, `enrichCatalogEntries` and `createDefaultEnricherClient` have no production callers — only tests. Nothing writes to `catalogEnrichment`, so `attachCatalogEnrichments` always returns entries unchanged and every new field (`tagline`, `shortControls`, `searchKeywords`) is `undefined` in prod. The vector index ends up embedding little more than `"Title. Genre."`. Is a backfill job meant to land with this?

**3. `/api/catalog/search` is unauthenticated with no rate limit.** Every other route in `submissions.ts` carries a `config.rateLimit`; this one bills Vertex per request without one, fed by a 200 ms debounce on the client.

**4. `ensureCatalogVectorIndex` is build-once with no dedup.** The `size() > 0` guard means new games never get indexed until restart; concurrent cold requests each rebuild the whole thing; and the sequential embed loop runs inside a user request.

**5. Missing `AbortController` on the client fetch.** The debounce cleanup doesn't cancel in-flight requests, so a slow earlier response can overwrite a newer match.

**6. The new `searchKeywords` match rule is very loose** and sits ahead of the hand-tuned aliases — bidirectional substring tests against LLM tags with no length floor (the fallback keyword path *does* apply one).

**7. The "offline" embedding test isn't offline** — it makes a real auth + `fetch` attempt.

**8. No server-side test for `/api/catalog/search`** — only the web-side test with a mocked `fetch`.

Minor: `-webkit-line-clamp` without the standard `line-clamp`, and a no-op `text-overflow` under `-webkit-box`; `extractFallbackKeywords` will throw if `entry.genre` is ever null despite the contract type.

The UX work itself (poster thumbnail, badges, positive no-match framing, clearer placeholders) reads well and is the uncontroversial part of the change. If you want to land that soon, splitting the semantic-search backend into its own PR would let items 1–4 be worked through without holding up the hero card.

---

apps/api/src/submissions.ts
```diff
@@ -5816,6 +5824,113 @@ export async function registerSubmissionRoutes(
...
+    const isStale = Date.now() - lastIndexBuildTime > INDEX_TTL_MS;
+    if (catalogVectorIndex.size() > 0 && !isStale) {
+      return Promise.resolve();
+    }
+    if (!indexBuildPromise) {
+      indexBuildPromise = buildCatalogVectorIndex().finally(() => {
+        indexBuildPromise = null;
+      });
+    }
+    return indexBuildPromise;
```
gtanczyk, line 5916: **The TTL guard is bypassed whenever the index is empty, which turns a Vertex outage into a rebuild hot-loop.**

```ts
if (catalogVectorIndex.size() > 0 && !isStale) return Promise.resolve();
```

`isStale` is only consulted when `size() > 0`. But `buildCatalogVectorIndex` only upserts entries where `vec.length > 0`, and `embedText` now returns `[]` on any Vertex failure — so if Vertex is down or unconfigured, **every** embedding is empty, nothing is upserted, and `size()` stays `0`. The guard then returns false on every call regardless of how recently the build ran.

Result: each `/api/catalog/search` request kicks a full rebuild — `getCatalogEntries` (GitHub), N Firestore `getCatalogEnrichment` reads, up to N `SPEC.md` fetches, and up to N Gemini calls. `indexBuildPromise` serialises them, but they run back-to-back continuously for as long as the outage lasts, at up to 120 req/min/IP. The failure mode that should be cheapest is the most expensive one.

Compounding it, `lastIndexBuildTime = Date.now()` sits inside the `try` at line 5875, after the loop. If `getCatalogEntries` throws, the `catch` swallows it and the timestamp stays `0`, so `isStale` is permanently `true` — same hot-loop, this time even if the index later has entries.

Suggested shape: record the attempt regardless of outcome, and gate on time first.

```ts
async function buildCatalogVectorIndex() {
  lastIndexBuildTime = Date.now();   // record the attempt, not the success
  try { /* … */ } catch { /* … */ }
}

function ensureCatalogVectorIndex() {
  if (Date.now() - lastIndexBuildTime <= INDEX_TTL_MS) return Promise.resolve();
  // …
}
```

Worth a shorter retry TTL for the empty-index case if you want faster recovery — but bounded, not per-request.

---

apps/api/src/submissions.ts
```diff
@@ -5816,6 +5824,113 @@ export async function registerSubmissionRoutes(
...
+              }
+            }
+          } catch {
+            // Non-blocking per-game enrichment
+          }
+        }
+      }
+
+      const enriched = await attachCatalogEnrichments(published, store);
+      for (const entry of enriched) {
```
gtanczyk, line 5884: Thanks for wiring the enricher up — that closes the dead-code gap. Two things about *where* it runs:

**1. The backfill is on the request path, serially.** On a cold Firestore this loop does, per game: a `getCatalogEnrichment` read, a `getGameFile('SPEC.md')` GitHub fetch, and a Gemini call with a 10 s timeout — all `await`ed one at a time, then N embedding calls after it. For a few dozen games that's minutes inside a single HTTP request, and Cloud Run will cut the request at its timeout long before it finishes. Because the partial build leaves `size()` at whatever it reached, the next request starts over.

A boot-time warm-up, or letting `/api/catalog/search` answer `{ match: null, score: 0 }` immediately while the build runs detached, would keep the endpoint's latency bounded. The client already treats a null match as "no match", so a cold miss degrades gracefully.

**2. `if (!existing)` makes `contentHash` dead.** `getOrEnrichCatalogGame` already compares `contentHash` against the SPEC.md it's handed and regenerates on mismatch — that's the whole point of the field. But it's only reached when no record exists, so once a game is enriched its tagline/keywords never refresh, however much the spec changes.

If the concern was avoiding an unconditional `SPEC.md` fetch per game per rebuild, gating on `updatedAt` age would keep refresh working without the per-cycle cost:

```ts
const existing = await store.getCatalogEnrichment(entry.slug);
const stale = !existing || Date.now() - Date.parse(existing.updatedAt) > ENRICHMENT_TTL_MS;
if (stale) { /* fetch spec + getOrEnrichCatalogGame */ }
```

---

apps/api/src/catalog/embedding-service.ts
```diff
@@ -0,0 +1,115 @@
...
+      if (this.cache.size >= VertexEmbeddingService.MAX_CACHE_SIZE) {
+        const firstKey = this.cache.keys().next().value;
+        if (firstKey) this.cache.delete(firstKey);
+      }
+      this.cache.set(trimmed, normalized);
+      return normalized;
+    } catch {
+      // Returns empty vector on Vertex AI failure or unconfigured credentials
+      return [];
+    }
```
gtanczyk: Dropping the pseudo-vector fixes the dangerous half of this — no more confident wrong matches. The observability half is still open though: `catch { return []; }` swallows auth failures, 429s, 5xx and malformed responses identically and silently.

An empty vector propagates to `{ match: null, score: 0 }`, which is exactly what a legitimate no-match looks like. So a misconfigured `VERTEX_PROJECT_ID`, a missing IAM binding, or a quota exhaustion leaves semantic search permanently dead with nothing in the logs to point at it — and per the `ensureCatalogVectorIndex` comment, it also puts the rebuild loop into its most expensive state.

`VertexEmbeddingService` takes no logger today, but `CatalogEnricherOptions` already establishes the pattern (`log?: (message: string) => void`). Even a `log?.(\`Vertex embedding failed: ${String(err)}\`)` here would make the difference between "nobody searches for these words" and "the integration is broken".

---

apps/api/src/catalog/embedding-service.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    const vec2 = await service.embedText('arcade football');
+    expect(vec2).toEqual([0.6, 0.8]);
+    expect(fetchSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns empty array cleanly when Vertex AI is offline or fails', async () => {
+    vi.spyOn(globalThis, 'fetch').mockRejectedValue(new Error('Network error'));
+
+    const service = new VertexEmbeddingService();
+    const vec = await service.embedText('arcade football');
```
gtanczyk, line 72: Much better — the caching test is properly isolated now.

One leftover in this second test: only `fetch` is mocked, not `auth.getClient`. `embedText` calls `getClient()` first, so this still triggers real ADC resolution (metadata-server probe, well-known credentials file lookup) before it ever reaches the mocked `fetch`. It passes because that throws into the same `catch`, but it's the original flake/latency risk in a smaller form — and it means the test isn't actually exercising the fetch-rejection path it's named for.

Stubbing `getClient` the same way the test above does would make the failure deliberate:

```ts
vi.spyOn((service as unknown as { auth: { getClient: () => Promise<unknown> } }).auth, 'getClient')
  .mockResolvedValue({ getAccessToken: async () => ({ token: 'mock-token' }) });
```

---

gtanczyk (commented):
## Re-review of `53d2206`

Good pass — 7 of the 8 findings are properly addressed, and several went further than I suggested (the query embedding cache wasn't asked for and does real work here). CI is green and the PR is `clean`.

**Resolved**

| # | Finding | How |
|---|---|---|
| 1 | Fallback embedding produced confident wrong matches | `generateDeterministicFallback` removed entirely; `embedText` returns `[]` and the route answers `{ match: null }` |
| 2 | Enricher had no production caller | `getOrEnrichCatalogGame` + `createDefaultEnricherClient` now run from `buildCatalogVectorIndex` |
| 3 | No rate limit on a paid endpoint | `{ rateLimit: { max: 120, timeWindow: '1 minute' } }`, plus a 500-entry embedding cache |
| 4 | Build-once index, no dedup | `indexBuildPromise` + 10-minute TTL |
| 5 | No request cancellation | `AbortController`, with the `aborted` check in `catch` |
| 6 | Loose `searchKeywords` substring match | `tokens.includes(kw) && kw.length > 2`, and the same floor applied in the enricher |
| 7 | "Offline" test made real network calls | `fetch` and `auth.getClient` mocked |
| 8 | No server-side route test | New `catalog-search.test.ts` |

Minors also picked up: `line-clamp: 2` added, the no-op `text-overflow` dropped, `entry.genre` guarded.

**Still open / newly introduced**

The fixes to #1 and #4 interact in a way that's worth another look before merge:

1. **Rebuild hot-loop when Vertex is unavailable** (`ensureCatalogVectorIndex`). The guard is `size() > 0 && !isStale`, so the TTL is only consulted once the index is non-empty. With Vertex down every embedding is `[]`, nothing gets upserted, `size()` stays `0`, and each request kicks a full rebuild — GitHub catalog fetch, N Firestore reads, N `SPEC.md` fetches, N Gemini calls — continuously. `lastIndexBuildTime` also only being set on the success path (inside `try`, after the loop) produces the same effect if `getCatalogEntries` throws. Gating on time first, and recording the *attempt* rather than the success, fixes both.

2. **The enrichment backfill runs inline on the request path**, serially, with a 10 s LLM timeout per game. On a cold Firestore the first search can't return within Cloud Run's request timeout. A boot warm-up or a detached build (answering `{ match: null }` meanwhile) would bound it.

3. **`if (!existing)` makes `contentHash` dead.** `getOrEnrichCatalogGame` compares the hash and regenerates on mismatch, but it's only reached when no record exists — so enrichment never refreshes after a `SPEC.md` change.

4. **No logging on Vertex failure.** An empty vector is indistinguishable from a genuine no-match, so a broken IAM binding or exhausted quota leaves search silently dead. `CatalogEnricherOptions` already has the `log?:` pattern to copy.

5. Minor: the second embedding test only mocks `fetch`, so it still does real ADC resolution before reaching it.

(1) is the one I'd want fixed before merge — it converts an outage into sustained GitHub + Gemini + Firestore load. The rest are follow-up-able.

---

PR #979, opened 2026-08-23
Phase 3 Wave A: move creation files into N1 buckets (final Wave A batch)

gtanczyk:
## Summary

Sixth and final batch of Phase 3 Wave A (following #971-#976): mechanical `git mv` of the creation N1 bucket, no logic changes. `apps/api/src/creation/` is now a real directory holding the job/round/seed/remix/refine machinery (`builder`, `chat-agent`, `code-lane`, `creation-limits`, `creator-code`, `creator-profile*`, `creator-studio`, `dispatch-reaper`, `editor-*`, `game-seed`, `job-*`, `quota-gate`, `refine`, `remix*`, `seed-*`, `session-crash`, `source-*`, `tab-complete`, `type-check*`, `version-verdict`, and friends — 97 files). Every relative import repo-wide was rewritten to match.

**With all nine N1 buckets (platform, creation, agent-surface, delivery, catalog, community, realtime, telemetry, notifications) now real directories, Phase 3 Wave A is complete.**

Reference classes the import-rewriter can't reach by construction, all caught by the full verification pass (which now always includes a full root `npm run build`, not just apps/api's own tsc/eslint/vitest — a lesson from #976's near-miss):
- `agent-build-examples.ts` imports its own `agent-build-examples.json` sibling data file via `with { type: 'json' }` — the rewriter intentionally never resolves non-`.js`/`.jsx` specifiers, so it never moves the file the specifier points at either. Moved the JSON file alongside its only consumer into `creation/`, keeping the import specifier unchanged.
- `tab-complete.test.ts`'s `vi.mock('./agent-surface/genai.js', ...)` was correct while both files sat flat in `apps/api/src`, but is a string literal the rewriter never touches — now that `tab-complete.test.ts` itself moved into `creation/`, `agent-surface/` is a sibling, not a child, and the mock needed `'../agent-surface/genai.js'`.
- `remix.test.ts` and `typecheck-preflight.test.ts` both moved themselves this batch, so their own already-correct hardcoded relative paths (`readFileSync(new URL('./platform/app.ts', ...))` from #976, and a `'../../../../www.gamedev.pl-games/...'` fixture lookup) needed one more `../` each.
- `moderation-metrics.test.ts`'s hardcoded call-site list needed `editor-drafts.ts`/`refine.ts`/`remix.ts` retargeted to `creation/`.

[...]

PR #982, opened 2026-08-23
fix(catalog): retarget games-repo-contract-check's hardcoded creation/ paths

gtanczyk:
## Summary

master's "Games-repo contract" CI check went red right after #979 merged, with `ENOENT: no such file or directory, open '.../apps/api/src/editor-contract.ts'`.

`LOCAL_EDITOR_CONTRACT_PATH` and `LOCAL_TS_ANY_SCAN_PATH` in `games-repo-contract-check.ts` are hardcoded string paths (not static imports), so the Phase 3 Wave A move script's import-rewriter never touched them — the exact same blind spot already fixed once for the catalog move in #973, now recurring because `editor-contract.ts` and `ts-any-scan.ts` moved again, this time into `creation/`, in #979.

The check reads the file directly off disk via `readFileSync` to diff it against the games-repo checkout's mirror copy, not via an import, so `tsc`/`eslint`/`vitest` all stayed green while this broke — only the live cross-repo CI job (which needs `GAMES_REPO_TOKEN`, unavailable locally) exercises this path.

Retargeted both path constants and their two error-message strings to `creation/`. Confirmed no other hardcoded string reference to either file remains anywhere in the repo, and that both new paths resolve to real files on disk.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx eslint apps/api/src apps/api/scripts` clean
- [x] `npx vitest run` in `apps/api` — 219 files / 3485 tests passed, including the 69 `games-repo-contract-check`/`games-repo-contract` unit tests
- [x] `npm run lint` at repo root — exits 0, back to the same 233 pre-existing cross-domain warnings, non-blocking
- [x] Verified `LOCAL_EDITOR_CONTRACT_PATH`/`LOCAL_TS_ANY_SCAN_PATH` resolve to real files on disk at the new location

---

### DIFF
--- a/apps/api/src/catalog/games-repo-contract-check.ts
+++ b/apps/api/src/catalog/games-repo-contract-check.ts
@@ -55,8 +55,11 @@ import {
 } from './games-repo-contract.js';
 import { isRateLimitResponse } from './github-rate-limit.js';
 
-const LOCAL_EDITOR_CONTRACT_PATH = path.join(path.dirname(fileURLToPath(import.meta.url)), '../editor-contract.ts');
-const LOCAL_TS_ANY_SCAN_PATH = path.join(path.dirname(fileURLToPath(import.meta.url)), '../ts-any-scan.ts');
+const LOCAL_EDITOR_CONTRACT_PATH = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex review

---

PR #985, opened 2026-08-23
fix(e2e): block service workers in signedInContext

gtanczyk:
## Summary

Third try at unblocking the deploy gate. #983 added a `page.route` mock for `GET /api/games/e2e-studio-shell` — but the deploy that built #983's own merge commit failed the exact same way, still 404ing against the real candidate server.

## Root cause

The app registers a service worker for offline game caching, and it intercepts `/api/games/:slug` requests. `page.route()` only sees requests initiated by the page itself — a service worker's own `fetch()` (from inside its `fetch` event handler) is a separate execution context that Playwright's page-level routing does not see, so the mock added in #983 silently never fired and the real request went straight through to the network, which correctly 404'd (the fixture slug isn't actually published).

## Fix

`serviceWorkers: 'block'` on the `signedInContext()` browser context (`apps/e2e/src/browser.ts`), used by `studio-shell.test.ts`, `games-playable.test.ts`, `resilience.test.ts`, and `walkthrough.test.ts`. None of these four suites test service-worker/PWA/offline behavior itself (checked), so disabling it removes the interception ambiguity without losing coverage — and for `games-playable.test.ts` (which plays real games over the real network, no mocking) it's a strict improvement: no risk of a stale SW-cached response making the test non-deterministic.

## Test plan
- [x] `npx tsc --noEmit` clean in `apps/e2e`.
- [ ] Can't reproduce the service-worker interception locally (needs a live candidate + registered SW) — verification is the next deploy's browser-gate run.

### DIFF
--- a/apps/e2e/src/browser.ts
+++ b/apps/e2e/src/browser.ts
@@ -161,9 +161,13 @@ export async function signedInApiContext(): Promise<APIRequestContext> {
   return request.newContext({ baseURL: BASE_URL, storageState, ...proxyOptions() });
 }
 
-/** A browser context carrying the bot's session cookie. */
+// Service workers blocked: the app's SW intercepts /api/games/:slug ahead of page.route().
 export async function signedInContext(browser: Browser, api: APIRequestContext): Promise<BrowserContext> {
-  return browser.newContext({ storageState: await api.storageState(), viewport: { width: 1280, height: 900 } });
+  return browser.newContext({
+    storageState: await api.storageState(),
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

gtanczyk:
Superseded by #986, merged with the same root-cause fix (independently diagnosed in a parallel session) — more surgical too, opt-in per-suite instead of blocking service workers on all four `signedInContext` call sites. Closing.

PR #986, opened 2026-08-23
fix(e2e): block service workers in studio-shell so route stubs work

gtanczyk:
## Summary

#983 already stubbed `GET /api/games/<slug>` for this suite, but the deploy pipeline's "Browser gate against the candidate" step still failed the same way: the candidate server still saw the real request and 404'd.

Root cause is documented directly in Playwright's own type definitions: `page.route()` does not intercept requests once a Service Worker controls the page — even one that, like this app's `sw.js`, passes API requests straight to the network untouched (`isApiRequest()` returns early without `respondWith()`). The real production worker registers during every `/studio` visit this suite makes, so from the second navigation on, every stubbed API route silently stopped mattering.

Added an optional `options` param to `signedInContext()` (backward compatible — the other three call sites are untouched) so a suite can opt into `serviceWorkers: 'block'`, which Playwright's docs recommend for exactly this case. `studio-shell.test.ts` uses it: the suite tests layout, not PWA/offline behavior, so blocking registration changes nothing it actually verifies while making every stubbed route reliable.

## Test plan

- [x] `npm run type-check -w @gamedevpl/e2e` clean
- [x] `npx eslint apps/e2e/src/browser.ts apps/e2e/src/studio-shell.test.ts` clean
- [x] `npx prettier --check` clean on both files
- [x] `npm run lint` at repo root — exits 0, back to the same 233 pre-existing cross-domain warnings, non-blocking
- [x] Confirmed via Playwright's own bundled type definitions (`node_modules/playwright-core/types/types.d.ts`) that this is the officially documented behavior and recommended fix, not a guess
- [x] Could not run the e2e suite itself locally (needs a live `GAMEDEV_ACCESS_TOKEN` and real deployed backend) — the next "Deploy to Cloud Run" run's browser gate is the actual verification

---


gtanczyk:
@codex review

---

PR #987, opened 2026-08-24
fix(web): guard synchronous SecurityError on navigator.serviceWorker access

gtanczyk:
## Summary
- PR #986 made the e2e deploy-gate use `serviceWorkers: 'block'` so Playwright's `page.route()` stubs are reliable against `studio-shell.test.ts`.
- That block mode makes Chromium treat the page as sandboxed without `allow-same-origin`, so the `navigator.serviceWorker` getter throws a synchronous `SecurityError` on access instead of the property simply being absent.
- `main.tsx` and `shellUpdate.ts` only guarded with `'serviceWorker' in navigator` (a safe existence check that doesn't invoke the getter), then accessed `navigator.serviceWorker` directly — so the throw became an uncaught page error, which failed `studio-shell.test.ts`'s zero-console-error assertion on every viewport (10 failing tests), and with it the "Browser gate against the candidate" step of `Deploy to Cloud Run` for commit `458eabd`.
- This is also a real robustness gap outside of tests: the same `SecurityError` can occur in production if the app is ever embedded in a sandboxed iframe without `allow-same-origin`.
- Fix: wrap the actual property access in `try/catch` in both files, treating the throw the same as "unsupported".

## Test plan
- [x] `npm run build` (root, full monorepo: lint + typecheck + unit tests + build) passes
- [x] `apps/web/src/shellUpdate.test.ts` unit tests pass
- [x] Resealed `eslint-rules/module-size-baseline.json` for the two files' legitimate line growth (prose-comment baseline untouched — comments were trimmed to stay within it)
- [ ] CI / deploy-gate against the candidate revision (this is the actual reproduction environment — local runs hit `https://www.gamedev.pl` by default, not a locally-built candidate)

---


gtanczyk:
@codex review

---

gtanczyk:
Addressed Copilot's review: `AppUpdateBanner.tsx` mounts on every signed-in page and had the same unguarded `navigator.serviceWorker` access — fixed with the same try/catch pattern. `pushApi.ts`'s `getRegistration`/`register`/`ready` call sites are now routed through a shared `serviceWorkerContainer()` helper so `isPushSupported()` correctly reports `false` when the getter itself throws, not just when it's absent.

@codex review

---

gtanczyk:
Superseded by #988 — an independent fix for the identical bug opened moments after this one (looks like a parallel session picked up the same deploy-gate failure). #988's approach is cleaner: a single shared `hasServiceWorkerSupport()` helper reused across all four call sites, plus a genuine regression test that reproduces the throwing-getter scenario directly (`serviceWorkerSupport.test.ts`) — something I could not verify locally here since `E2E_BASE_URL` defaults to production.

Closing this in favor of #988.

---

PR #988, opened 2026-08-24
fix(web): guard navigator.serviceWorker reads against a throwing getter

gtanczyk:
## Summary

Deploy attempt #4. #986 blocked service workers for `studio-shell.test.ts` to make its route stubs reliable (root-causing #983's earlier 404), which fixed the gate's original failure — but the deploy still failed, this time on:

```
[pageerror] SecurityError: Failed to read the 'serviceWorker' property from 'Navigator': Service worker is disabled because the context is sandboxed and lacks the 'allow-same-origin' flag.
```

## Root cause

Every service-worker call site in this app guards with `if ('serviceWorker' in navigator)` (or the negation). That guard is not actually safe: `in` checks property *existence* via `[[HasProperty]]` and never invokes the accessor's getter — verified directly:

```js
Object.defineProperty(Navigator.prototype, 'serviceWorker', { get() { throw new Error('boom'); } });
'serviceWorker' in navigator   // → true, does not throw
navigator.serviceWorker        // → throws
```

Chromium's `serviceWorkers: 'block'` mode (what #986 turned on) implements the block by making the property *exist but throw on read* rather than simply omitting it — so every `in`-guard in the app passes, and the very next line (a real read, e.g. `navigator.serviceWorker.addEventListener(...)`) throws for real. Same failure mode a real sandboxed iframe missing `allow-same-origin` would hit in production, not just this Playwright mode.

## Fix

Added `hasServiceWorkerSupport()` (`apps/web/src/serviceWorkerSupport.ts`) — reads the property inside a `try`/`catch` instead of only `in`-checking it — and swapped it into the four call sites that gate service-worker use: `main.tsx`, `shellUpdate.ts`, `AppUpdateBanner.tsx`, `pushApi.ts`'s `isPushSupported()`.

## Test plan
- [x] Verified the `in`-vs-read distinction directly in Node against a throwing getter (see above).
- [x] New unit test (`serviceWorkerSupport.test.ts`) reproduces the throwing-getter case and asserts `hasServiceWorkerSupport()` returns `false` without throwing.
- [x] Full `apps/web` suite: 1648/1648 passing.
- [x] `npx tsc --noEmit` clean.
- [ ] Can't reproduce the sandboxed-context throw itself locally (needs Playwright's `serviceWorkers: 'block'` against a live candidate) — verification is the next deploy's browser-gate run.


gtanczyk:
@codex review

(Note: this supersedes #987, an independent fix for the identical bug opened moments earlier by a parallel session — closing that one in favor of this one, since this has a dedicated `hasServiceWorkerSupport()` helper with a real regression test reproducing the throwing-getter case, which is cleaner than #987's per-call-site try/catch.)

---

gtanczyk:
Fixed: `afterEach` now deletes the injected `serviceWorker` property when there was no original descriptor to restore, instead of silently no-op'ing.

@codex review

---

PR #994, opened 2026-08-24
feat(lint): exempt type-only imports from the module-boundary rule

gtanczyk:
## Summary

Phase 3 of `north-star-architecture.md` (private ops repo) asks to flip the `gamedev/module-boundary` rule from warn to error, module by module, now that Wave A physically moved files into their N1 buckets.

Before starting that flip, this fixes a gap in the rule itself: a whole-statement `import type { X } from '...'` (or `export type { X } from`) is erased at compile time — it carries no runtime coupling. Flagging it the same as a real value import conflates "a domain depends on another's published type" (exactly the shape `packages/contract` — N2 — is meant to formalize) with actual behavioral coupling that the boundary rule should catch.

A mixed statement importing at least one real value (`import { a, type B } from './x.js'`) is still checked — the exemption is whole-statement only.

## Effect

- `apps/api/src` warnings: 237 → 179 total; `crossBucket` specifically: 186 → 132.
- Zero behavior change to the app — this only changes what the (still warn-mode) `npm run module-boundary` reports.
- The 132 remaining `crossBucket` warnings are genuine value-level cross-domain calls — the real work for the module-by-module error flip, coming in follow-up PRs.

## Test plan
- [x] Added `module-boundary.test.mjs` cases for the exemption (`import type`, `export type`) and confirmed a mixed import (`import { a, type B }`) still reports `crossBucket`
- [x] `npx vitest run eslint-rules/module-boundary.test.mjs` — 19/19 passing
- [x] Re-ran `npm run module-boundary` against real `apps/api/src` and confirmed the count drop matches expectations
- [x] Full root `npm run build` (lint + typecheck + unit tests + build) passes

---

### DIFF
--- a/eslint-rules/module-boundary.mjs
+++ b/eslint-rules/module-boundary.mjs
@@ -98,7 +98,12 @@ export const moduleBoundary = {
 
     const directory = path.dirname(filename);
 
-    function check(source) {
+    function check(node, source) {
+      // Erased at compile time -- a type-only edge is the contract N2 plans to give its own
+      // package, not runtime coupling. Only the whole-statement form (`import type { X }
+      // from`) is exempt; a mixed statement (`import { type X, y }`) still imports a real
+      // value and is checked like any other.
+      if (node.importKind === 'type' || node.exportKind === 'type') return;
       if (!source || source.type !== 'Literal' || typeof source.value !== 'string') return;
       const specifier = source.value;
       if (!specifier.startsWith('./') && !specifier.startsWith('../')) return;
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex review

---
_Generated by [Claude Code](https://claude.ai/code)

---

gtanczyk:
Good catch — fixed in #998: dropped the private-plan reference from the comment, keeping only the local implementation-level rationale.

---
_Generated by [Claude Code](https://claude.ai/code)

---

PR #1008, opened 2026-08-25
fix(api): raise build-preview byte cap from 320KB to 700KB

gtanczyk:
## Summary

Raises `MAX_BUILD_PREVIEW_BYTES` from 320KB to 700KB.

## Why

The fast in-process preview (#970, retry-hardened in #975) is still failing silently for real games — Cloud Run logs show `"candidate preview too large"` at 22:57 today for issue #1000135, version `v20260824T225657255Z-7aa754`, assembled at **698,997 bytes** (~683KB), well over the 320KB cap.

That cap was never actually about a Firestore constraint despite the old comment ("clears the games-repo bundle cap"). Each preview is its own Firestore document (`apps/api/src/store/slices/build-media.ts`'s `previewsCollection(issueNumber).doc(id).set(...)`), stored as base64 (`Buffer.from(html, 'utf8').toString('base64')` in `staged-preview.ts`) — base64 inflates by ~4/3. The real ceiling is Firestore's 1MiB-per-document hard limit, which in raw-HTML terms is closer to ~768KB before the encoding overhead eats the budget — the app-level cap was sitting at less than half of what the store can actually hold, for no documented reason tied to that constraint.

700KB raw → ~956KB base64, leaving ~90KB headroom under the 1,048,576-byte Firestore limit (small field overhead — `slug`, `label`, `id`, `createdAt` — is a few hundred bytes, negligible against that margin). This comfortably covers the failing case above.

## Test plan
- [x] `npx tsc --noEmit` clean in `apps/api`.
- [x] `npx vitest run src/delivery/staged-preview.test.ts` — 39/39 passing (the `too_large` test overrides `maxBytes` explicitly, so it's unaffected by the default change).
- [x] Checked for other tests hardcoding the old 320KB value — none (the one hit for `700 * 1024` in `agent-channel.test.ts` is an unrelated screenshot-upload size cap, confirmed by reading its context).

### DIFF
--- a/apps/api/src/delivery/build-preview-limits.ts
+++ b/apps/api/src/delivery/build-preview-limits.ts
@@ -1,2 +1,2 @@
-// 320KB decoded — clears the games-repo bundle cap; never duplicate this number.
-export const MAX_BUILD_PREVIEW_BYTES = 320 * 1024;
+// base64-encoded, own doc per preview — stays under Firestore's 1MiB cap.
+export const MAX_BUILD_PREVIEW_BYTES = 700 * 1024;


gtanczyk:
@codex

PR #1049, opened 2026-08-26
Extract the feedback and improve routes out of submissions.ts

gtanczyk:
`POST /api/submissions/:token/feedback` and `.../improve` are the same act at two points in a game's life: a creator asking for a change. Feedback steers a draft still being built; improve opens a new round on a game that has already shipped, because publishing is terminal. They are neighbours, not one routine — so they land as two files, not one. At 276 and 203 lines of route body, a combined file would also have cleared the 500-line cap.

| new file | registrar | deps |
| --- | --- | --- |
| `creation/feedback-routes.ts` | `registerFeedbackRoutes` | 14 |
| `creation/improve-routes.ts` | `registerImproveRoutes` | 14 |

**submissions.ts: 3171 → 2665 lines.** Since #1046 it has lost 858 lines and five route clusters.

## The shared schema

`FeedbackRequestSchema` and `REFERENCE_IMAGES_BODY_LIMIT_BYTES` were private to `submissions.ts` and are read by both new files *and* by the create route that stays. Exporting them from `submissions.ts` would have been a runtime import cycle, so they move to `creation/feedback-request.ts` instead. `ReferenceImagesSchema` is named there too, so the create route no longer reaches for a bare `z.array(referenceImageSchema)`.

`ResumeOutcome` and the improvement-round return type are declared structurally in the new files, for the same cycle reason as #1048.

## Verification

Each route body was diffed against `master` with comments and blank lines stripped: **byte-identical**, 200 and 174 code lines respectively. The seam lists were not guessed from a regex scan — that gave false negatives on this cluster — but enumerated by `tsc` against a first-pass deps interface.

Comments carried over from `submissions.ts` were rewritten as `//` one-liners; both new files start at zero comment-prose debt. `npm run lint` exits 0, `tsc` clean, 3550 tests pass.

## The guard test

`moderation-metrics.test.ts` pins the list of every module calling `logModerationRejection`, so the suite cannot pass vacuously. Two call sites moved, and it correctly failed until it named the new files. Its line ceiling is raised by two for those entries: a pinned inventory grows when the modules it inventories split, and there is nothing in it to shrink.

---


gtanczyk:
@codex

PR #1050, opened 2026-08-26
Extract the job reconcilers out of submissions.ts

gtanczyk:
`reconcileNativeJob` and `reconcileGateVerdict` are the two places the platform finds out what actually happened to a round, from the only two authorities there are: the agent's own session, and our gate's verdict off the version manifest. Neither writes a reply, neither is a route, and both are **already** passed as deps to code that lives elsewhere — `notify-sweep-routes` takes both, and so does the status refresh. They move to `creation/job-reconciler.ts` behind `createJobReconciler`, so those callers stop reaching into this file for them.

Unlike the last five batches this is a **factory, not a route registrar**: it returns the two functions rather than mounting anything on the app.

**submissions.ts: 2653 → 2331 lines.**

## The two textual changes

Both mechanical, and both called out because they mean this is not a byte-for-byte move:

- **16 `app.log` → `log`**, a dep bound to `app.log` where the factory is constructed. A narrowed `ReconcilerLog` names the three levels the reconcilers actually use, rather than dragging Fastify's logger type across the seam.
- **`const gamesStore = options.agentChannel?.gamesStore` becomes a dep**, read once at construction instead of on every gate poll. Same object, same reasoning as #1048 and #1049.

## Verification

Both bodies diffed against `master` with comments and blank lines stripped. The raw diff shows exactly the 16 logger call sites and nothing else; after normalising `app.log` → `log` on the master side, the only remaining line is that `gamesStore` const. Nothing else moved.

The `reconcileNativeJob` doc comment travelled with the implementation rather than being left behind in `submissions.ts` above a factory call it no longer describes. All prose was rewritten as `//` one-liners — the new file starts at zero comment-prose debt.

`npm run lint` exits 0, `tsc` clean, 3550 tests pass.

---

### DIFF
--- a/apps/api/src/creation/job-reconciler.ts
+++ b/apps/api/src/creation/job-reconciler.ts
@@ -1,0 +1,382 @@
+import { deriveGateStatusString, derivePreviewGateStatus } from '@gamedevpl/contract';
+import type { AgentBackend } from '../agent-surface/agent-backend.js';
+import { postGateScreenshotToThread } from '../delivery/gate-screenshot.js';
+import { probeGateCrash } from '../delivery/gate-crash.js';
+import type { GamesStore } from '../delivery/games-store.js';
+import {
+  builderLabelFromRecord,
+  failedStageFromProgress,
+  logDeliveryGateVerdict,
+  type DeliveryGateStatus,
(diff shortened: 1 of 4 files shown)


gtanczyk:
@codex

PR #1053, opened 2026-08-26
Extract dispatchBuild and redispatchQueuedJob out of submissions.ts

gtanczyk:
`dispatchBuild` starts a round on a job that has none: it mints the per-job channel credential, decides whether and how a seed reaches the agent, opens the round with the selected backend, and records what that cost. `redispatchQueuedJob` is the reaper's one retry when a dispatch died before producing a ref, and calls straight into it. Both move to `creation/dispatch-build.ts` behind `createDispatcher`, together with `reconstructDispatchSpec`, which only `redispatchQueuedJob` uses.

**submissions.ts: 1917 → 1693 lines.** Since #1046 the file has lost **1830 lines** and its last long-lived helper. What remains is route registration, the status cache, and the wiring that hands these modules to each other.

## The one structural change

`dispatchBuild` was a hoisted **function declaration**, so it could reference `seedDeliveryFor`, `seedBuild` and `publishSeedPreview` from a `const` destructured 350 lines further down. A factory call is evaluated where it sits, so `createDispatcher` is now called *after* that destructuring.

Both consumers — `startImprovementRound` and the `createGameCreator` wiring — run later still, so nothing observes the move. `tsc` catches this class of mistake either way, and did: the first attempt put the call at the old position and got four "used before its declaration" errors.

## Stale comment, again

`resumeBuild`'s doc comment stayed in `submissions.ts` when #1052 moved the function, sitting above a factory call and describing something no longer there. It moves to `resume-build.ts` here. That is the third such find in four PRs — #1051 caught `buildDispatchIssueBody` claiming a consumer that did not exist, #1052 caught two comments explaining a cycle that no longer existed.

## Verification

Both bodies diffed against `master` with comments and blank lines stripped: **byte-identical**, 167 code lines. `reconstructDispatchSpec` identical too. Seams enumerated by `tsc`, which also corrected three guesses: `isActiveBuildRound` lives in `builder.ts` not `job-state.ts`, `seedOutcomeFor` was a missed import, and `managedAvailabilityGate` is not used by this family at all.

`npm run lint` exits 0, `tsc` clean, 3550 tests pass.

---

### DIFF
--- a/apps/api/src/creation/dispatch-build.ts
+++ b/apps/api/src/creation/dispatch-build.ts
@@ -1,0 +1,280 @@
+import type { AgentBackend, SeedFiles } from '../agent-surface/agent-backend.js';
+import { mintAgentToken, mintManagedMcpOpener } from '../agent-surface/agent-token.js';
+import type { GameSeeder } from './game-seed.js';
+import type { BuilderKind } from './builder.js';
+import type { Store, SubmissionRecord } from '../platform/store.js';
+import { buildDispatchIssueBody } from './create-game.js';
+import { seedOutcomeFor } from '../agent-surface/seed-status.js';
+import { isActiveBuildRound } from './builder.js';
+import { canTransition } from './job-state.js';
+import type { SeedPipeline } from './seed-pipeline.js';
+
+// Rebuilds a dispatch spec from stored spec and qa; null if none.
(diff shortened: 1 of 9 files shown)


gtanczyk:
@codex

PR #1054, opened 2026-08-27
Mount the agent channel and the MCP server from buildApp

gtanczyk:
`registerSubmissionRoutes` was awaiting two other modules' registrars as its last act, so the app's route table was decided partly inside a 1,700-line leaf. Both mounts move to `buildApp`, which already owns every other registration. This is the Phase 3 line "the agent channel and MCP server mount from it, not from `submissions.ts`".

## The seam

`registerSubmissionRoutes` returns `agentSurface` — the half of each option object only it can build. `channel` is a `Pick` over `AgentChannelOptions` of: `agentTokenSecret`, `now`, `sourceDelivery`, `onEvent`, `onBuilderHandoffAcknowledged`, `onSourcesStaged`, `onRegenerateSeed`. `mcp` is a `Pick` over `McpServerOptions` of: `agentTokenSecret`, `now`, `startImprovementRound`, `continueDraftRound`, `createGame`, `contentChecker`, `dailyImprovementQuota`, `dailyFeedbackQuota`.

These are closures over the status cache, the staged-preview publisher, the quotas and the round verbs — values with no other way out of that function. Typing them as a `Pick` over the registrars' own option types means the seam list cannot drift from what they accept.

`buildApp` supplies the half it owns (store, buckets, gate trigger, flags) and calls both registrars itself:

    await registerAgentChannelRoutes(app, { ...agentChannelOptions, ...agentSurface.channel, store });
    await registerMcpServerRoutes(app, { ...agentSurface.mcp, store, platformConnectorSecret, privateBeta, ... });

`submissions.ts` now imports both modules type-only.

## Three options that only tunnelled through

`resolveProposalBase`, `privateBeta` and `platformConnectorSecret` were declared on `SubmissionRoutesOptions`, read at exactly one place each — the MCP mount — and never used by the submission routes themselves. All three leave. `platformConnectorSecret` moves to `BuildAppOptions`, which is where its one test seam injects it now; that also removes the `delete agentChannel.platformConnectorSecret` dance the `mcp-server.test.ts` helper needed to keep it off the channel options.

## Behaviour

Both option objects carry the same keys with the same values, and no key appears in both spread groups, so spread order cannot change a value. `store` moves from third position to last in the channel options — nothing after it writes that key.

Nothing runs between the registrar returning and the two mounts, and both were previously the last statements inside it, so route registration order relative to everything else in `buildApp` (the mp relay, the beta wall hook) is unchanged.

`registerSubmissionRoutes` has one other caller, `creator-code.test.ts`, and it goes through `buildApp`.

Full API suite: 226 files, 3550 tests, green.

## Comments corrected rather than carried over

Moving the call sites exposed three comments that were false at the new location:

* The catalog-lane block said the MCP proposal tools "are registered by `registerSubmissionRoutes` below" — after this change they are registered in that same file, twenty lines down.
* A note said the module "keeps no games-repo or snapshot dependency". True of `submissions.ts`; `app.ts` builds `gamesRepoClient` immediately above.
* The `MCP_UI` note described `mcp-server.ts`'s own env read and sat above two unrelated values. Dropped — the flag stays covered by `infra/env-manifest.json` and its own tests.

## Ratchets

`app.ts` grows 20 lines and its module-size ceiling is resealed 1067 → 1087. The composition root gaining the route table is the direction the plan asks for, and `submissions.ts` — the file the freeze protects — did not grow. Its comment-prose baseline reseals 8072 → 3908, locking in the slack the extraction wave left behind.

`npm run lint` exits 0.

### DIFF
--- a/apps/api/src/agent-surface/mcp-server.test.ts
+++ b/apps/api/src/agent-surface/mcp-server.test.ts
@@ -189,11 +189,11 @@ async function createApp(
   return await buildApp({
     store,
     sessionSecret: 'dev-session-secret-change-me',
+    ...(platformConnectorSecret ? { platformConnectorSecret } : {}),
     submissionRoutes: {
       githubClient: stubGitHub(),
       githubToken: 'gh-token',
       submissionTokenSecret: secret,
-      ...(platformConnectorSecret ? { platformConnectorSecret } : {}),
       agentChannel: {
         ...(gamesStore ? { gamesStore } : {}),
         ...(objectStore ? { objectStore } : {}),
--- a/apps/api/src/platform/app.ts
+++ b/apps/api/src/platform/app.ts
@@ -63,6 +63,8 @@ import { registerScorecardRoutes, type ScorecardRoutesOptions } from '../creatio
 import { createInternalAuthVerifierFromEnv } from './internal-auth.js';
 import { registerRefineRoute, type SpecRefiner } from '../creation/refine.js';
 import { BOT_UID_PREFIX, InMemoryStore, type Store } from './store.js';
+import { registerAgentChannelRoutes } from '../agent-surface/agent-channel.js';
+import { registerMcpServerRoutes } from '../agent-surface/mcp-server.js';
 import { registerSubmissionRoutes, type SubmissionRoutesOptions } from '../submissions.js';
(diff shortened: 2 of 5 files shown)


gtanczyk:
@codex review

---

PR #1055, opened 2026-08-27
Stop 5xx responses echoing the error message to the client

gtanczyk:
`setErrorHandler` appears nowhere in this repo. Every uncaught throw in every route has been falling through to Fastify's default 500, which puts `err.message` in the response body verbatim. Whatever the failing call put in that message — a bucket name, a filesystem path, a signed URL, a token fragment — went out to the caller with it.

This is the Phase 3 line "add the app-wide error handler mapping typed domain errors to HTTP shapes".

## What throws

21 non-test `.parse(` sites throw rather than `safeParse`. 15 are in route handlers reading `request.params`, spread across `suggestion-inbox.ts` (4), `draft-lifecycle-routes.ts` (3), `handoff-seal-routes.ts` (2) and one each in `game-play-route.ts`, `draft-preview-routes.ts`, `submissions.ts`, `feedback-routes.ts`, `improve-routes.ts`. The other 6 parse model replies inside the LLM call helpers, and a `ZodError` message quotes the input it rejected. Beyond zod, any throw from the GCS client, the GitHub client or the Firestore driver lands the same way.

## The handler

    app.setErrorHandler((error: FastifyError, request, reply) => {
      // Fastify reads both; statusCode wins when an error carries each.
      const statusCode = error.statusCode ?? (error as { status?: number }).status ?? 500;
      if (statusCode >= 400 && statusCode < 500) {
        void reply.send(error);
        return;
      }
      request.log.error({ err: error, method: request.method, url: request.url }, 'unhandled route error');
      void reply.code(500).send({ error: 'internal' });
    });

`{ error: '<code>' }` is what the API already speaks — 81 distinct codes across the routes, and nothing sends a `message` key.

Only a real 4xx passes through. An error carrying a status outside 400–499 (including one carrying none) is treated as a failure and flattened, so a thrown `statusCode: 302` cannot smuggle the message out through the pass-through branch.

Three things I checked against Fastify 5.10 directly rather than assumed:

* `reply.send(error)` inside the handler does **not** re-enter it — four requests, four handler invocations — and reproduces the default body byte for byte.
* The default handler reads `status` as well as `statusCode`, preferring `statusCode` when an error carries both. This repo has four error classes that use `status` — `GitHubRequestError`, `AgentTasksError`, `ManagedAgentError`, `MailerError` — so classifying on `statusCode` alone would have turned a 404 from one of those into a 500. That is a regression, not a missed improvement, and it is why the handler reads both. `tsc` could not have caught it: `status` is not on the `FastifyError` type.
* The default handler logs 5xx itself, so a custom handler that does not log would silently remove every 500 from Cloud Logging. Hence the explicit `request.log.error`, with a test asserting the message reaches the log even though it no longer reaches the client.

No plugin sets its own error handler, so nothing was overridden. `setNotFoundHandler` is untouched.

## The web rendered raw server strings

`throwResponseError` puts the response body's `error` field into `Error.message`. Two call sites then displayed that message to the user directly, so a 503 rendered the literal text `store_unavailable` on screen and a 500 rendered the English `Internal Server Error` to a Polish creator. Pre-existing, but this PR changes what lands there, so both are fixed.

`SubmissionStatusView.tsx` now shows `t('errors.generic')`: its non-400 failures are transient, the poller retries on its own cadence, and there is no recovery action to name.

`App.tsx`'s chain needed more care. An expired session (401 `authentication required`, from `checkUserAccess` in `platform/auth.ts`) and the coarse IP limiter (429 `too many submissions, please try again later`, from `create-game.ts`) match none of the message branches, so folding them into a generic fallback would have replaced recoverable guidance with "Something went wrong" — and for the 401, retrying cannot resolve it. Two new keys in both locales, classified by `status`.

That chain moved to `submissionErrors.ts` as a pure `submissionErrorKey({ status, message, category })`, because the ordering is subtle and nothing covered it: **three** different refusals answer 429 — the creator's daily quota, the two site-wide limits, and the IP limiter — and only the last should say "you went too fast".

## Verification

* 8 tests in `apps/api/src/platform/error-handler.test.ts`. Reverting `app.ts` alone fails the leak assertions and passes the unchanged-behaviour ones, so they discriminate in both directions.
* 7 tests in `apps/web/src/submissionErrors.test.ts`, one of which asserts every key the mapper can return resolves in both `en.json` and `pl.json`.
* API suite 227 files / 3558 tests green. Web suite 199 files / 1749 tests green.
* No test anywhere asserted a 500 status before this PR, and `grep` finds no consumer of the string `Internal Server Error`.
* `npm run lint` exits 0. `app.ts` grows 18 lines; its module-size ceiling reseals 1087 → 1105.

## Deliberately not in this PR

Mapping `ZodError` to a 400 `{ error: 'invalid_request' }`. It is the more correct status for those 15 routes, but it changes a status code rather than removing a leak, and it deserves its own diff.

### DIFF
--- a/apps/api/src/platform/app.ts
+++ b/apps/api/src/platform/app.ts
@@ -223,7 +223,8 @@ export async function buildApp(options: BuildAppOptions = {}): Promise<FastifyIn
 
   // Fastify's default 500 echoes err.message; 4xx replies pass through.
   app.setErrorHandler((error: FastifyError, request, reply) => {
-    const statusCode = error.statusCode ?? 500;
+    // Fastify reads both; statusCode wins when an error carries each.
+    const statusCode = error.statusCode ?? (error as { status?: number }).status ?? 500;
     if (statusCode >= 400 && statusCode < 500) {
       void reply.send(error);
       return;
--- a/apps/api/src/platform/app.ts
+++ b/apps/api/src/platform/app.ts
@@ -3,7 +3,12 @@ import fastifyStatic from '@fastify/static';
 import fastifyWebsocket from '@fastify/websocket';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
-import Fastify, { type FastifyInstance, type FastifyRequest, type FastifyServerOptions } from 'fastify';
+import Fastify, {
+  type FastifyError,
+  type FastifyInstance,
+  type FastifyRequest,
+  type FastifyServerOptions,
+} from 'fastify';
 import { registerAccessTokenRoutes } from './access-token-routes.js';
 import { registerJobAdminRoutes } from '../creation/job-admin-routes.js';
 import { createGameSeederFromEnv } from '../agent-surface/agent-backend-env.js';
@@ -215,6 +220,18 @@ export async function buildApp(options: BuildAppOptions = {}): Promise<FastifyIn
     trustProxy: 1,
     routerOptions: { maxParamLength: MAX_REMIX_ID_LENGTH },
   });
+
+  // Fastify's default 500 echoes err.message; 4xx replies pass through.
+  app.setErrorHandler((error: FastifyError, request, reply) => {
+    const statusCode = error.statusCode ?? 500;
+    if (statusCode >= 400 && statusCode < 500) {
+      void reply.send(error);
+      return;
+    }
+    request.log.error({ err: error, method: request.method, url: request.url }, 'unhandled route error');
+    void reply.code(500).send({ error: 'internal' });
+  });
+
   const store = options.store ?? new InMemoryStore();
 
   const isProd = process.env.NODE_ENV === 'production';
--- a/apps/api/src/platform/error-handler.test.ts
+++ b/apps/api/src/platform/error-handler.test.ts
@@ -35,6 +35,19 @@ async function appWithThrowingRoute(logLines?: string[]) {
     throw error;
   });
 
+  // GitHubRequestError, MailerError and friends carry their code in `status`.
+  app.get('/api/test-throw-status-field', async () => {
+    const error = new Error(LEAK) as Error & { status?: number };
+    error.status = 404;
+    throw error;
+  });
+
+  app.get('/api/test-throw-status-field-5xx', async () => {
+    const error = new Error(LEAK) as Error & { status?: number };
+    error.status = 503;
+    throw error;
+  });
+
   app.get('/api/test-throw-redirect', async () => {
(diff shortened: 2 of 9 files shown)


gtanczyk:
@codex review

---

gtanczyk:
Fixed in `e2f23f3`, and the finding was right in a way worth spelling out: it was not a missed improvement but a regression this PR would have introduced.

I checked Fastify 5.10 directly rather than reasoning about it. Its default handler reads both fields and prefers `statusCode`:

    .status = 404              -> 404
    .status = 404, .statusCode = 422 -> 422

So a `GitHubRequestError` carrying `status: 404` out of a route produced a 404 before this PR, and my handler — reading only `statusCode` — would have classified it as "no status" and answered 500. The classification now mirrors Fastify's own precedence:

    const statusCode = error.statusCode ?? (error as { status?: number }).status ?? 500;

Two tests added: a `status: 404` passes through with the default body, a `status: 503` is flattened to `{ error: 'internal' }` with no message. 8 tests in the file, API suite 227 files / 3558 tests green, `npm run lint` exits 0.

@codex review

---

gtanczyk:
Confirmed and fixed in `494ce50`. I traced both responses rather than assuming:

* `checkUserAccess` (`platform/auth.ts:978`) sends 401 `authentication required`, and `POST /api/submissions` calls it — so the expired-session case is real.
* `create-game.ts:159` sends 429 `too many submissions, please try again later` for the coarse IP limiter.

Neither matches a branch, so both were landing in the fallback. Worth noting what the old code actually did with them: it rendered `err.message` verbatim, so a Polish creator saw the English string `authentication required`. The finding stands either way — an untranslated but actionable sentence still beats "Something went wrong", and for the 401 no amount of retrying helps.

Two new keys in both locales (`errors.signInRequired`, `errors.tooManyAttempts`), classified by `status`.

The ordering here is subtle enough to be worth guarding, and nothing covered it: **three** different refusals answer 429 — the daily quota, the two site-wide limits, and the IP limiter — and only the last should say "you went too fast". So the chain moved to `submissionErrors.ts` as a pure `submissionErrorKey({ status, message, category })`, with 7 tests pinning the precedence and one asserting every key it can return resolves in both `en.json` and `pl.json`.

`SubmissionStatusView.tsx` deliberately stays generic: its non-400 failures are transient, the poller retries on its own cadence, and there is no recovery action to name.

Web suite 199 files / 1749 tests green, API 227 / 3558 green, `npm run lint` exits 0.

@codex review

---

PR #1056, opened 2026-08-27
Make platform/store.ts a façade over the two store implementations

gtanczyk:
`platform/store.ts` was 2812 lines with a clean seam already inside it, left over from the Phase 2 slice work:

| Lines | What | Count |
| --- | --- | --- |
| 1–335 | re-export prelude + the `Store` aggregate interface | 335 |
| 337–1542 | `class InMemoryStore implements Store` | 1206 |
| 1544–2812 | `class FirestoreStore implements Store` | 1269 |

Two facts made the split mechanical rather than a judgement call: the file has **zero** module-private top-level declarations (the helpers left in Phase 2 wave 4), and neither class references the other. So both come out whole.

`store.ts` is now 302 lines and re-exports both classes, so all **229** files that import from it are untouched by this PR.

## Behaviour

Both bodies moved unchanged. Comment- and blank-stripped and whitespace-trimmed, the new files are identical to the corresponding ranges of the old one — 954 code lines for `InMemoryStore`, 1010 for `FirestoreStore`. The diff is a move plus import blocks.

Each new file's imports were derived from `store.ts`'s own import map by script and then confirmed by `tsc`, rather than assembled by eye — both files compiled clean on the first pass. `tsc` then named the 34 now-dead imports in `store.ts`, and only those were removed.

The one import neither file could inherit is `Store` itself, which is declared in `store.ts`. Both take it as `import type`, which is erased at compile time, so there is no runtime cycle and the boundary rule (type-only imports exempt) is unaffected.

## Ratchets

`store.ts` shrinks 2812 → 302 lines and 817 → 625 prose words. Both of its baselines reseal **down**; leaving the old ceilings in place would have let the file regrow to its former size unnoticed, which is the opposite of what the freeze is for.

The two new files are sealed at their moved sizes (1286 and 1350 lines, 40 and 145 prose words) rather than being rewritten to fit the 500-line new-file cap. They are moved code, and the byte-identical proof above is worth more than the cap here — rewriting the comments or resplitting the bodies would forfeit it. Both are under the 1500-line Phase 3 exit criterion. Splitting what remains of each class is the continuation of the Phase 2 slice work and belongs in its own diff.

The second commit exists because the first one sealed those two baselines *before* the pre-commit prettier pass rewrapped the generated import blocks, which put both files 27 lines over their own freshly-written ceiling and failed `module-size` in CI. Only the import blocks changed; the code-line comparison above was re-run against the committed files and still holds.

`npm run lint` exits 0.

## Verification

API suite 227 files / 3558 tests green, including `store-parity.test.ts`, which runs the shared contract suite against both implementations.


gtanczyk:
@codex review

---

gtanczyk:
Correct finding, and already fixed — the review ran against `9e4a548`, and `e4692a7` landed a few minutes before it with exactly this change. Recording it here since the diagnosis is worth keeping:

The cause was ordering, not judgement. I sealed both baselines, then `git commit` ran the repo's `lint-staged` prettier pass, which rewrapped the generated import blocks and added 27 lines to each file — putting both over the ceiling I had just written. The baselines were right for the text I measured and wrong for the text that got committed.

On the current head the two now agree:

    apps/api/src/store/in-memory.ts    1286 lines, baseline 1286
    apps/api/src/store/firestore.ts    1350 lines, baseline 1350

`Lint, Test & Build` is green on `e4692a7`, along with the other seven checks.

The exemption is retained rather than split, for the reason in the PR body: these are moved bodies, and the code-line proof (954 and 1010 lines, identical to the old ranges) is worth more here than the 500-line new-file cap. I re-ran that comparison against the committed files after the prettier pass — it still holds, because prettier only touched the import blocks, which is also why the raw line count moved while the stripped comparison did not.

@codex review

---

PR #1058, opened 2026-08-27
Give the store publication lane one published predicate

gtanczyk:
First of the "one meaning for publish" batches, and the mechanical one.

## What the scan found

The plan's justification for this item points at `submissions.ts:5575`, which no longer exists — the catalog merge moved into `catalog/catalog-routes.ts` during Wave B. The underlying complaint holds, and is larger than the plan describes: **three** vocabularies for "published", not two, across **32** direct checks in 24 files.

| Lane | Spelling | Sites |
| --- | --- | --- |
| store | `publication.state === 'published'` | 19 |
| repo catalog | `entry.status === 'published'` | 13 |
| submission record | `record.publishedAt` OR'd into a predicate (`creation/job-costs.ts`), and `status.status === 'published'` (`notifications/notify.ts`) | 2 |

Only the *membership* question is unified today, behind `PublishedSlugGate` / `createCombinedPublishedSlugGate` in `catalog/published-slugs.ts`, built once in `app.ts`. That covers telemetry, votes and recommendations. Everything else asks directly.

## This PR

The store lane. Seventeen sites across twelve files each spelled the same question in one of four shapes:

    publication?.state === 'published'
    publication?.state !== 'published'
    !publication || publication.state !== 'published'
    publication?.state === 'published' ? publication.currentVersion : null

`platform/publication-state.ts` now holds `isPublished` and `publishedVersion`, and all seventeen call them. `isPublished` is a type guard, so the sites that relied on narrowing to keep the record typed — `game-page-routes.ts`'s `? publication : null`, and every `currentVersion` read after a check — still narrow. `tsc` passes with no added assertions, which is the evidence that none of them lost it.

`platform/` is where it belongs by the boundary map's own rule: "every real (value-level) edge out of the bucket stays inside `platform/` or its own bucket", and four of the twelve importers are in enforced buckets. It sits next to `round-base-version.ts`, whose one publication check was the `publishedVersion` shape verbatim.

## Not in this PR

* The 13 repo-catalog sites.
* The third spelling in `job-costs.ts`.
* Making the repo lane an importer that writes the same registry entry — the plan's actual end state. That one carries a correctness trap worth stating before anyone starts it: today a game removed from `catalog.json` drops out of the gate on the next TTL refresh, but an importer would leave it published forever without reconciliation or tombstones. `published-slugs.ts` also records that the catalog fan-out has already caused one GitHub rate-limit outage, so turning it into a per-refresh Firestore writer is a load-profile change, not just a plumbing one. I'll bring a design for that before writing any of it.

## Verification

Behaviour-preserving by construction: every replacement is the same comparison behind a call, and `!publication || publication.state !== 'published'` collapses to `!isPublished(publication)` because the guard already answers false for `null`.

API suite 227 files / 3557 tests green. `npm run lint` exits 0.

Seven files gain exactly one line each — the shared import — and their ceilings reseal by one. `publication-state.ts` is registered as a platform primitive in the boundary map; without that entry the rule correctly refuses to trust the new edge rather than silently allowing it.

---


gtanczyk:
@codex review

---

PR #1059, opened 2026-08-27
Give the repo catalog lane one published predicate

gtanczyk:
Batch 2 of "one meaning for publish". [#1058](https://github.com/gamedevpl/www.gamedev.pl/pull/1058) did the store lane; this does the repo catalog lane.

Ten sites across eight files each compared `entry.status` against the literal `'published'`. `CATALOG_PUBLISHED_STATUS` and `isPublishedEntry` now live in `packages/contract/src/catalog-entry.ts` beside `CatalogEntry`, following `isBuilderKind`'s precedent in the same package, and all ten call the predicate.

The contract package rather than `platform/` because that is where the type lives, and `apps/web` reads the same entries — which turned out to matter, see below.

## What the unification exposed

The two lanes are not symmetric, and the predicate makes that legible rather than fixing it:

| Lane | Field | Type |
| --- | --- | --- |
| store | `PublicationRecord.state` | `'published' \| 'archived' \| 'disabled'` |
| repo catalog | `CatalogEntry.status` | `string` |

The repo lane's status is a bare string parsed out of agent-authored SPEC frontmatter — any value at all typechecks, and only an exact `'published'` match counts as live. This PR does not change that; it gives the comparison one home so it can be tightened in one place later.

## The second commit, and why it was needed

My first pass missed two sites because the scan itself was wrong in two ways: it matched only `=== 'published'`, making every negation invisible, and it covered `apps/api` only. Each gap hid exactly one real site, both spelled `status !== 'published'`:

* `apps/web/src/catalog.ts` — `normalizeCatalogEntry`'s reject list
* `apps/api/src/catalog/game-page-routes.ts` — the repo-entry guard

That combination was the one worth catching: with `catalog.ts` left behind, a later tightening of `isPublishedEntry` would have made web's normalizer drop entries the API still served — two contradictory definitions of a live game, which is the exact failure this item exists to prevent. Re-running the corrected scan (both operators, all workspaces) finds no other `CatalogEntry` liveness check.

## Matches deliberately left alone

`grep` finds fourteen textual matches. Six are not this vocabulary, and folding them in would have changed behaviour:

* `creation/seed-context.ts` — reads `entry.status === undefined || entry.status === 'published'`. An absent status counts as published there, which is a different rule; `isPublishedEntry` would silently drop that case.
* `catalog/github-client.ts` — the parser that *produces* the field, not a consumer of it.
* `notifications/notify.ts`, `apps/web/src/creatorGames.ts`, `apps/web/src/EditorPanel.tsx` — `status.status` / `CreatorGameItem.status` / the `getSubmissionStatus` response. Same words, submission vocabulary, not a catalog entry. Batch 3.
* `community/suggestion-outcomes.ts` — a comment.
* `platform/dev-seed-studio.ts` — a dev-seed fixture shape.

## Verification

Behaviour-preserving: every replacement is the same comparison behind a call, and the `.filter((entry) => entry.status === 'published')` cases become `.filter(isPublishedEntry)` with the same arity.

`tsc` clean on both workspaces. API suite 227 files / 3557 tests green, web suite 199 files / 1749 tests green. `npm run lint` exits 0.

Five ceilings reseal: four files gain the import line, and `catalog-entry.ts` gains the predicate itself.

## Still to come in item 6

Batch 3 is the third spelling (`record.publishedAt` OR'd into a predicate in `creation/job-costs.ts`, plus the submission-status reads listed above). Batch 4 is the plan's actual end state — the repo lane as an importer writing the same registry entry — which needs a reconciliation design first, since a game removed from `catalog.json` would otherwise stay published forever.


gtanczyk:
@codex review

---

gtanczyk:
Correct, and it found a real gap in how I scanned rather than a judgement call. Fixed in `047f660`.

My scan for this batch had two blind spots: it matched only `=== 'published'`, so every negation was invisible, and it covered `apps/api` only. Both of the surviving checks are spelled `status !== 'published'`, and one of them is in `apps/web` — so each gap hid exactly one site.

    apps/web/src/catalog.ts:185          normalizeCatalogEntry's reject list
    apps/api/src/catalog/game-page-routes.ts:114   repo-entry guard

Both now call `isPublishedEntry`. Re-running the corrected scan (both operators, all workspaces) finds no other `CatalogEntry` liveness check.

Two nearby matches are deliberately untouched, and they are the reason the corrected scan still needs reading rather than a blanket replace: `creatorGames.ts` reads `CreatorGameItem.status` and `EditorPanel.tsx` reads the `getSubmissionStatus` response. Both spell the same words over the submission vocabulary, not a catalog entry — folding them in would merge two lanes that genuinely differ. They belong to batch 3.

The point about a later tightening creating contradictory definitions is the right frame for why this mattered: with `catalog.ts` left behind, a stricter `isPublishedEntry` would have made web's normalizer drop entries the API still served.

API 227 files / 3557 tests green, web 199 / 1749 green, `npm run lint` exits 0.

@codex review

---

PR #1061, opened 2026-08-27
Remove the dead 'gating' job state

gtanczyk:
Item 4 of the north-star plan.

## Why it was safe to remove

`gating` was `JobState`'s one unreachable value, and the codebase already said so before this PR — a comment on the union itself: *"Our gate runs against delivered sources; never actually observed here."* Another, on the transition table, records the incident that taught this: when `gating` was `submitted`'s only exit, a delivered job's reconciler computed the right target (`ready_for_review` / `needs_changes`), `canTransition` refused it because nothing had ever transitioned *into* `gating`, and the job sat showing "delivered, gate never started" for as long as the creator kept the page open.

The mechanism: the gate reports by writing its verdict straight to the version manifest, which is read on a poll — so a delivered job goes from `submitted` to its outcome without ever being seen mid-gate. `submitted → gating` was the only entry in `ALLOWED_TRANSITIONS`, and nothing wrote it.

## What moved

Removed from `JOB_STATES` (contract package) and every switch, comparison, and transition list that named it — 8 sites in `apps/api`, 5 in `apps/web`. Each was found by `tsc`, not by eye: removing the union member first and fixing exactly the errors it named.

Two matches for the word are deliberately untouched, because they are different vocabularies:

* `community/proposal-state.ts` — `ProposalState` has its own separate `gating`, for proposal review, unrelated to job state.
* `agent-surface/mcp-ui-round-status.html` — a display label the round-status card computes from `gate.deliveryId` ("a delivery in flight reads as gating"), not a read of the job record.

## The one real risk, and how it's covered

`toSubmissionStatus` is an exhaustive switch with no default, and nothing validates a stored record's `state` on read. So a Firestore document written while `gating` still existed — however implausible, given the above — would fall through the switch and answer `undefined` on the status route.

`fromStoredSubmission` (`store/records/submission.ts`) normalizes a legacy `'gating'` to `'submitted'` at every materialization point in the submission store slices — 3 in `submission.ts`, 8 in `submission-queries.ts`, all Firestore reads, confirmed with `grep` before and after. Six tests pin it now: the rewrite itself, that every other state passes through untouched, that a record with no `state` at all is untouched, that the input object isn't mutated, and — added after review — the same rewrite exercised against a real `FirestoreStore` via `fakeFirestore()`, plus a regression test for a document gone by read time (the account-deletion race described below).

## Review round

Codex found three real issues on the first pass, all fixed:

* `packages/contract` has its own test pinning `JOB_STATES`' full membership, which I hadn't run (only `apps/api` and `apps/web`). Updated to the eleven-state list.
* The web locale key `statusView.phases.gating` has a second, live reader I'd missed: `SubmissionStatusView.tsx` and `StudioStageCard.tsx` both use it to label a *running* gate, keyed off `gateProgress`/`gate.stage` — unrelated to the dead `phase === 'gating'` comparisons. Restored in both locales.
* `fromStoredSubmission` crashed on an `undefined` snapshot — a document gone by read time, e.g. an account-deletion race. Rather than widen the function's return type (which `tsc` showed would force every safe `QueryDocumentSnapshot` call site in `submission-queries.ts` to handle a case that can't occur there), the fix guards the one call site that actually lacked the `snap.exists` check its two siblings already had.

## Verification

`tsc` clean on both workspaces, and on the `packages/contract` build. Eight existing tests asserted the old vocabulary directly — including the one written to document the original incident — and are updated to assert the current transitions rather than deleted; none is weaker than before. API suite 229 files / 3563 tests green (6 new). Web suite 200 files / 1752 tests green. Contract package 42 files / 100 tests green. `npm run lint` exits 0.

### DIFF
--- a/apps/api/src/creation/builder.test.ts
+++ b/apps/api/src/creation/builder.test.ts
@@ -89,7 +89,6 @@ describe('builder helpers', () => {
     expect(shouldSteerFeedbackViaInbox({ state: 'dispatched', ...withRef })).toBe(true);
     expect(shouldSteerFeedbackViaInbox({ state: 'building', ...withRef })).toBe(true);
     expect(shouldSteerFeedbackViaInbox({ state: 'submitted', ...withRef })).toBe(true);
-    expect(shouldSteerFeedbackViaInbox({ state: 'gating', ...withRef })).toBe(true);
     expect(
       shouldSteerFeedbackViaInbox({
         state: 'needs_changes',
--- a/apps/api/src/creation/builder.ts
+++ b/apps/api/src/creation/builder.ts
@@ -18,7 +18,6 @@ export function isActiveBuildRound(record: { state?: JobState; transitions?: Job
     case 'dispatched':
     case 'building':
     case 'submitted':
-    case 'gating':
     case 'publishing':
       return true;
     case 'needs_changes': {
--- a/apps/api/src/creation/job-admin-routes.test.ts
+++ b/apps/api/src/creation/job-admin-routes.test.ts
@@ -86,9 +86,9 @@ describe('buildJobQueue', () => {
   });
 
   it('shows both the internal state and what the creator is being told', () => {
-    const queue = buildJobQueue([record({ issueNumber: 1, state: 'gating', stateSince: ago(MINUTE) })], NOW);
+    const queue = buildJobQueue([record({ issueNumber: 1, state: 'submitted', stateSince: ago(MINUTE) })], NOW);
 
-    expect(queue.jobs[0].state).toBe('gating');
+    expect(queue.jobs[0].state).toBe('submitted');
     expect(queue.jobs[0].creatorStatus).toBe('building');
   });
 
@@ -427,4 +427,3 @@ describe('GET /api/admin/jobs/:issueNumber/preview', () => {
     await app.close();
   });
 });
-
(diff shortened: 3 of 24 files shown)


gtanczyk:
@codex review

---

gtanczyk:
All three confirmed real, not judgement calls. Fixed in `8bad4cc`.

**P1 (contract test).** Right — I verified `apps/api` and `apps/web` and never touched `packages/contract`'s own suite. Updated `job-state.test.ts` to the eleven-state list.

**P2 (locale key).** Also right, and it exposed a gap in my read of the code: I deleted `statusView.phases.gating` on the premise that its only reader was `status.phase === 'gating'`, which is genuinely dead. But `SubmissionStatusView.tsx` and `StudioStageCard.tsx` both also read that same key as the label for a running gate, keyed off `gateProgress`/`gate.stage` — a live, unrelated concept wearing the same translation key. Restored in both locales.

**P2 (undefined snapshot).** Right, and worth explaining the fix shape: widening `fromStoredSubmission`'s return type to admit `undefined` was the wrong repair — it would have forced every `QueryDocumentSnapshot` call site in `submission-queries.ts` to handle a case that provably can't occur there (`tsc` confirmed this by turning 8 new errors on when I tried it). The real gap was narrower: `setSubmissionPublishedAt` was the one call site missing the `snap.exists` guard its two siblings (`getSubmission`, `recordDeliveryNudge`) already had. Added it there.

New `gating-normalize.test.ts` uses `fakeFirestore()` against the real `FirestoreStore`, not a mock — I reverted the fix locally and confirmed the crash test fails without the guard and passes with it, plus a second test for the normalization itself.

API 229 files / 3563 tests green, web 200 / 1752 green, contract package 42/100 green (the suite I'd missed). `npm run lint` exits 0.

@codex review

---

PR #1062, opened 2026-08-27
Rename issueNumber to jobId across contract, store, API, and web

gtanczyk:
Item 5 of the north-star plan.

## Why

The field was already called `jobId` in practice everywhere that mattered: `store.allocateJobId()`, `nextJobId`, and the local variable at nearly every call site (`const jobId = await store.allocateJobId(); await store.createSubmission(jobId, ...)`). Only the stored/typed name lagged behind — `SubmissionRecord.issueNumber`, the `Store` interface, two admin route params, and the wire fields in `packages/contract`.

## Scope check (same collision-check as items 4 and 6)

`catalog/github-client.ts` and `catalog/local-games-repo.ts` also use the identifier `issueNumber` — but for actual GitHub issue numbers on the separate games-repo, a genuinely different concept (currently uncalled by any live code path). Confirmed by tracing `allocateJobId()`'s counter (`nextJobId`, `JOB_ID_FLOOR`) back to its own origin: an independent internal sequence, unrelated to GitHub. Both files, and their tests, are deliberately excluded from the rename.

Everything else — `SubmissionRecord.issueNumber`, the `Store` interface, `/api/admin/jobs/:issueNumber/{publish,preview}`, and `RecentBuild.issueNumber` / `SubmissionStatusResponseBase.issueNumber` in `packages/contract` — is one field, renamed everywhere it appears: 162 files across `packages/contract`, `apps/api`, and `apps/web`.

## Why one PR, not a batch split like items 4/6

This isn't independently migratable. `'gating'` removal and the "publish" predicates could land in checkpointed batches because each call site could move on its own schedule. A field rename on `SubmissionRecord`/`Store` is atomic — `tsc` won't allow a partially-renamed state, since every caller of `store.createSubmission(issueNumber, …)` or every reader of `record.issueNumber` breaks the moment the interface changes, in both `apps/api` and `apps/web` at once.

## What moved

Mechanical identifier rename (`\bissueNumber\b` → `jobId`), covering bare identifiers, destructured properties, `Pick<...>` string-literal members, and the two route path params — followed by a manual pass collapsing the resulting `jobId: jobId` object-literal noise into shorthand (`{ jobId, ... }`). No logic changes.

## Verification

`tsc` clean on `packages/contract`, `apps/api`, `apps/web`. Full production build (`npm run build`) succeeds. All three test suites green at their existing baselines: contract 42 files / 100 tests, API 229 files / 3563 tests, web 200 files / 1752 tests. `npm run lint` exits 0. Two module-size baselines resealed after prettier reformatted pre-existing overlong lines it touched along the way (`build-prompt.test.ts`, `source-delivery.ts`) — resealed after formatting, per the usual ratchet rule.

This is item 5 of the authorized 1–7 program; item 8 (Phase 4 web restructure) remains out of scope.

---


gtanczyk:
@codex review

---

gtanczyk:
All three findings confirmed and fixed in 7a4cd95:

- **P1** (`store/records/submission.ts`): `fromStoredSubmission()` now backfills `jobId` from a stored `issueNumber` when the new field is absent, same pattern as the existing `gating`→`submitted` normalization. Firestore is schemaless, so every already-persisted submission needed this.
- **P2** (`delivery/games-store.ts`): centralized all nine raw-JSON-to-`VersionManifest` parse sites behind one `parseVersionManifest()` helper that does the same backfill for GCS-stored manifests.
- **P2** (`scripts/managed-probe.ts`): updated, plus a real bug this same sweep gap caused in `scripts/managed-probe.ts`'s sibling `scripts/slug-backfill.ts` — both live outside the API tsconfig, so `tsc` never caught them. `slug-backfill.ts`'s `existing.issueNumber` was always `undefined` post-rename, which broke its collision check.

Six new tests cover the normalization (both the Firestore and GCS paths) and the "already-migrated record passes through untouched" case. Full verification unchanged: `tsc` clean on all three workspaces, production build succeeds, API suite 229 files / 3567 tests (+4), lint exits 0.

@codex review

---

gtanczyk:
Confirmed and fixed in c534961: `docs/deployment.md`'s dry-run output shape updated to `{jobId, title, slug}`. Also swept the rest of `docs/*.md` for the same drift, since none of it was in the original rename's file list (markdown was never in scope for the `apps/api/src apps/web/src packages/contract/src` grep) — six more architecture-plan docs described Firestore paths, token minting, or log correlation keys under the old field name; updated all of them to match what the code already does.

@codex review

---

gtanczyk:
Verified this one against the code, and I'm not adding the alias — the coexistence window is real, but it's not novel to this PR and the repo already has a general answer to it.

The service worker precaches the shell per build revision and hands control to `AppUpdateBanner.tsx` on the next deploy (`sw.js`'s own header: "a stale shell now lasts exactly as long as someone ignores a banner" — mounted app-wide in `App.tsx`, not scoped away from `/admin`). That's the deliberate, already-shipped mitigation for exactly this class of problem: an open tab running an old bundle against a newer API after any breaking wire change, admin panel or otherwise. It isn't specific to this rename, and the repo's answer to it isn't "version every wire field" — it's "tell the stale tab to reload," which is what happens here too.

Given that, and that this item was scoped as one clean rename specifically to avoid a field living under two names during a rollout window, I'd rather rely on the existing mechanism than reintroduce `issueNumber` as a permanent-looking wire alias for a low-severity, self-resolving, admin-only edge case (broken action buttons until the banner's prompted reload, not data loss or a security issue).

---

gtanczyk:
@codex review

---

gtanczyk:
Both verified against the code.

**P1 (rollback dual-write) — confirmed, fixed in 0e84456.** `fromStoredSubmission()`/`parseVersionManifest()` cover old data read by new code, but not the reverse: a Cloud Run rollback (`docs/runbooks/rollback-deploy.md` — traffic reassignment, seconds, no rebuild) swaps to a previous revision whose code only reads `issueNumber`. `createSubmission` and `putCandidateSources` — the two places that create a brand-new record/manifest — now dual-write `issueNumber` alongside `jobId`. Every other write path only merges specific fields into an already-created document, so it doesn't need this. Two new tests assert the raw stored bytes carry both ids while the returned/typed object stays clean of the legacy field. Drop the dual-write once the pre-rename revision ages out as a rollback target.

**P2 (stale-client status/build-bar aliases) — declined, same reasoning as the earlier alias suggestion, traced through to be sure.** This is a stale-tab-not-a-rollback case, and it's cosmetic here: I followed `buildBarModel`'s guard failure through to what it actually gates, and the one concrete consequence (a Seal button rendering on the wrong preview row) doesn't matter, because `POST /api/submissions/:token/seal` derives `jobId` from the auth token server-side (`verifyToken(token, ...)`) and reads the current record via `store.claimSeal`, never from anything the client sends. A stale render can't make it act on the wrong round — worst case is a confusing button until the existing `AppUpdateBanner` prompts a reload, same as the admin-panel case.

@codex review

---

PR #1096, opened 2026-08-30
Phase 4: core/persistence.ts — one localStorage wrapper for the web

gtanczyk:
Phase 4 (item 8) of the north-star plan — web restructure. `core/`'s data-layer bullet: "localStorage access goes through one persistence module." This is that module — first batch of the post-Wave-A `core/` work, ahead of the fetch-cache data layer and the studio-poller consolidation (both scoped as separate follow-up PRs).

## What changed

`apps/web/src/core/persistence.ts` wraps `localStorage`/`sessionStorage` access (`readStorageItem`/`writeStorageItem`/`removeStorageItem` + JSON convenience wrappers), matching the fail-silent try/catch convention every one of the 12 modules that touched storage directly had already independently reinvented. Migrated: `builderKind.ts`, `connectCollapse.ts`, `i18n/index.ts`, `mySpecs.ts`, `pendingQa.ts`, `priorRoundDismiss.ts`, `pwa.ts`, `recentPlays.ts`, `surfaces/catalog/catalogSort.ts`, `surfaces/studio/StudioConnectCard.tsx`, `surfaces/studio/SubmissionStatusView.tsx`, `surfaces/studio/studioWelcome.ts`. Each call site is a mechanical swap — same keys, same fail-silent semantics.

**A real bug caught along the way**: a naive `storage: Storage = localStorage` parameter default evaluates *before* the function body's `try` runs, so it doesn't catch the `ReferenceError` thrown by referencing the bare `localStorage` identifier outside a browser (81 test files run in vitest's default Node environment, not jsdom, where `localStorage` isn't a global at all). The full test suite caught this immediately (81 failing files) — fixed by resolving the storage object *inside* the `try`, matching what every migrated call site already did.

## Verification

- `tsc --noEmit` clean
- `apps/web` suite: 217/217 test files, 1830 tests
- Repo `npm run lint`: 0 errors (comment-prose and module-size baselines resealed)
- Production build succeeds

## Scope note

The fetch-cache data layer (`core/dataLayer.ts`, request dedup + invalidation) and the five-Studio-pollers-into-one consolidation are separate follow-up PRs — this is just the persistence piece.

---

### DIFF
--- a/apps/web/src/builderKind.ts
+++ b/apps/web/src/builderKind.ts
@@ -2,6 +2,7 @@
 // without waiting on a status field the API may not yet echo.
 
 import { BUILDERS, isBuilderKind, type BuilderKind } from '@gamedevpl/contract';
+import { readStorageItem, writeStorageItem } from './core/persistence.js';
 
 export { BUILDERS, isBuilderKind, type BuilderKind };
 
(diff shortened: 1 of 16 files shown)


gtanczyk:
@codex review

---

gtanczyk:
Both addressed in b28a9fe:

- **P3** (`writeStorageJSON` circular/bigint crash): fixed — `JSON.stringify` now runs inside its own `try`, degrading to a no-op like the rest of this API.
- **P2** (`window.localStorage`/`window.sessionStorage` access outside the wrapper): the `pwa.ts` resolver itself is fixed — it now delegates to a new `resolveWebStorage()` exported from `core/persistence.ts`, so there's one guarded lookup instead of two. The other five `sessionStorage`-only files you found (`recommendationsApi.ts`, `remixSessionPersist.ts`, `shellUpdate.ts`, `StudioShotToasts.tsx`, `webmcp.ts`) are left out of this PR on purpose: the north-star plan's data-layer bullet specifically says "*localStorage* access goes through one persistence module," this PR's own title and scope are about that, and none of those five were touched by the 12-file localStorage migration this PR already tested. Migrating them is the same mechanical pattern as this batch and is a natural, low-risk follow-up — just a separate one, so this PR's diff stays what was actually verified.

---

PR #1097, opened 2026-08-30
Phase 4: core/dataLayer.ts — fetch-cache with dedup + invalidation

gtanczyk:
Phase 4 (item 8) of the north-star plan — web restructure. `core/`'s data-layer bullet: "one small fetch-cache module (hand-rolled, ~200 lines, no dependency) with request dedup and invalidation." This is that module — second batch of the post-Wave-A `core/` work (after `core/persistence.ts`, #1096), ahead of the poller consolidation itself.

## What changed

`apps/web/src/core/dataLayer.ts` exports `fetchCached`, `peekCached`, `invalidateCached`, and `invalidateCachedPrefix`. Wired into `submissionApi.ts`'s `getSubmissionStatus()` — the endpoint all five Studio pollers independently hit on their own timers — with the default `ttlMs: 0`, so only genuinely concurrent callers for the same key share one request; a call that starts after the previous one resolved always refetches. This changes nothing about freshness today; it just collapses the near-simultaneous requests those five pollers already produce.

**A real correctness bug caught while writing the invalidation tests**: invalidating a key mid-flight left its `inflight` promise sitting in the cache, so a *new* `fetchCached` call for that key would join the doomed in-flight request (and get its stale result) instead of starting its own fetch. Fixed with a generation counter — a resolving fetch only writes its result to the cache if it's still that entry's current generation, so a stale in-flight response never clobbers a newer one and a new caller after invalidation always starts fresh.

## Verification

- `tsc --noEmit` clean
- `apps/web` suite: 218/218 test files, 1843 tests, including `core/dataLayer.test.ts` (11 tests covering dedup, TTL reuse, rejection non-caching, and every invalidation race) and the real poller consumers (`SubmissionStatusView`, `useStudioStatusPoll`, `StudioDetailsBuildProgress`, `StudioDetailsMedia`, `StudioShotToasts`, `EditorPanel`) — all pass unchanged, including the fake-timer polling-cadence tests
- Repo `npm run lint`: 0 errors (comment-prose and module-size baselines resealed)
- Production build succeeds

## Scope note

This wires the data layer into one endpoint only — the one the poller consolidation (next PR) actually needs. Migrating the other ~23 `*Api.ts` fetch wrappers onto this cache is a separate, much larger effort not attempted here.

---


gtanczyk:
@codex review

---

gtanczyk:
Both addressed in 97b6d8b:

- **P2** (generation counter lost on delete): confirmed real — traced the exact repro you described. `invalidateCached`'s non-in-flight branch called `cache.delete(key)`, so a settled-then-invalidated entry's generation counter disappeared; the next fetch restarted at 1, colliding with a still-settling older fetch's captured generation. Fixed: `invalidateCached` now always replaces the entry with a tombstone that keeps bumping the generation, never deletes. Added a regression test (`dataLayer.test.ts`) that reproduces your scenario exactly (invalidate an in-flight fetch → let a second fetch settle → invalidate that too → start a third fetch → resolve the first, stale one last) and asserts it never overwrites the newer entry.
- **P2** (mutations not invalidating): fixed — `abandonSubmission`, `deleteGame`, `handoffToPlatform`, `handoffToSelf`, `sealPreview`, and `submitFeedback` now all call a new `invalidateSubmissionStatus(token)` on success, which drops every locale variant for that token via `invalidateCachedPrefix`. A status read issued right after one of these always starts its own fresh fetch now, rather than possibly joining an in-flight request that predates the mutation.

---

PR #1098, opened 2026-08-30
Phase 4: studioStatusStore.ts — shared poll registry for Studio status

gtanczyk:
Phase 4 (item 8) of the north-star plan — web restructure. "Replace the five studio pollers with one status store" — this is the store itself, ahead of migrating any of the five consumers onto it (each is a separate follow-up PR, in risk order).

## What changed

`apps/web/src/surfaces/studio/studioStatusStore.ts` exports `subscribeStudioStatus(token, locale, subscriber)` and `pokeStudioStatus(token, locale)`. One shared poll runs per `(token, locale)` key, regardless of how many components subscribe to it — replacing what today are five independent `setInterval`/`setTimeout` loops (`StudioDetailsBuildProgress`, `StudioDetailsMedia`, `StudioShotToasts`, `useStudioStatusPoll`, `SubmissionStatusView`) each hitting `GET /api/submissions/:token` on their own timer.

Each subscriber supplies its own cadence policy via `intervalMs(latest, error) => delay | null` — a fixed-cadence consumer (the three Details/Toast surfaces) just returns a constant; an adaptive one (`useStudioStatusPoll`'s `delayFor`, `SubmissionStatusView`'s `pollDelayMs`) inspects the latest status. The shared loop ticks at the **fastest** cadence any current subscriber still wants (so no subscriber ever polls slower than it asked to), and stops once every subscriber opts out (returns `null`) or unsubscribes. `pokeStudioStatus` forces an immediate re-poll ahead of the schedule — for visibility/focus/pageshow wake-ups and "I just mutated this, don't wait out the cadence" callers.

Built on `core/dataLayer.ts`'s `fetchCached` (via `getSubmissionStatus`, unchanged) for request dedup at the network layer; this store consolidates the *schedules* themselves.

## Verification

- `tsc --noEmit` clean
- `apps/web` suite: 219/219 test files, 1856 tests, including 12 new tests for the store covering: first-subscriber fetch, concurrent-subscriber sharing, late-subscriber-gets-cached-value, min-cadence-across-subscribers, stop-when-all-opt-out-then-resume, timer-cleared-on-last-unsubscribe, reschedule-on-partial-unsubscribe, error delivery + per-subscriber opt-out-on-error, and `pokeStudioStatus` (immediate refetch, no duplicate scheduled tick, no-op on an unknown key)
- Repo `npm run lint`: 0 errors (comment-prose baseline resealed)
- Production build succeeds (bundle unchanged — nothing imports this module yet)

## Scope note

Not wired into any component in this PR. The five consumers migrate in separate follow-up PRs, lowest-risk first: `StudioDetailsBuildProgress`/`StudioDetailsMedia`/`StudioShotToasts` (simple fixed-cadence readers), then `useStudioStatusPoll` (visibility/focus wake-ups, feeds `StudioStrip`/`StudioVersionRibbon`/`useStageSource`), then `SubmissionStatusView` last (owns status-transition side effects — telemetry, pending-revision reconciliation).

---


gtanczyk:
@codex review

---

gtanczyk:
Both fixed in 2549b5e:

- **P2** (deadline drift on unsubscribe): confirmed real. `schedule()` now tracks `lastTickAt` and schedules the *remaining* time (desired interval minus elapsed since that tick, clamped to zero) instead of a fresh interval from whenever it happens to run. Added a regression test for your exact scenario. Side effect worth flagging: this also means a subscriber joining a poll that's gone dormant well past its own cadence now gets an immediate catch-up tick instead of waiting out a fresh interval — updated two existing tests whose expectations assumed the old (buggy) behavior.
- **P2** (double delivery on mid-flight join): confirmed real, traced the exact race. Added a `ticking` flag: a subscribe while a tick is in flight no longer schedules its own timer — the in-flight tick's own completion reschedules against the updated subscriber set when it's done. Added a regression test that reproduces the race (a subscriber joins, a cadence-length delay passes while the fetch is still unresolved, then the fetch resolves) and asserts every subscriber gets exactly one delivery.

---

gtanczyk:
@codex review

---

gtanczyk:
Confirmed real, fixed in 4d8c461. Traced it: `subscriber.onUpdate?.(status)` runs inside the outer `try`, so a throw there was actually redirected into the `catch` block and treated as a fetch error (calling `onError` on every subscriber, including the ones that already got their `onUpdate`) — and if that `onError` also threw, the whole thing escaped uncaught, skipping `state.ticking = false` and `schedule()`. Once `ticking` got stuck `true`, the shared poll was dead for every subscriber, old or new.

Fixed by isolating each subscriber callback in its own try/catch and moving the `lastTickAt`/`ticking`/`schedule()` cleanup into a `finally` block, so one subscriber's bug can no longer affect delivery to the others or the tick's own bookkeeping. Added a regression test (`keeps ticking and notifies the other subscribers when one onUpdate throws`) that fails against the pre-fix code and passes now.

@codex review

---

gtanczyk:
All three confirmed real, fixed in 4f418f3:

- **Cached-delivery throw orphans the subscriber**: confirmed — a throwing `onUpdate`/`onError` during the synchronous cached-value delivery escaped before the `isDormant` gate ran, so `tick()`/`schedule()` never started for that key, but the subscriber (and its unsubscribe closure, never returned to the caller) stayed registered forever. Isolated with the same per-callback try/catch pattern as the tick delivery path.
- **`intervalMs` throw kills scheduling for everyone**: confirmed — `nextDelay()` propagates a throw out of `schedule()`, called from `tick()`'s `finally`; nothing catches it there, so it becomes an unhandled rejection with no replacement timer set, silently stopping the poll for every subscriber, broken or not. Fixed by isolating each `intervalMs()` call in `nextDelay()` — a broken cadence policy is now just skipped, not treated as a vote.
- **Stale success beats a fresher error on join**: confirmed. Traced the invariant: a successful tick always clears `lastError`, but a failed tick never clears `latest` — so `lastError` being non-null is *always* more recent than `latest`. Swapped the check order (error first) rather than adding a new field, since that invariant already gives the right answer.

Added one regression test per finding, each reproducing the exact scenario described and failing against the pre-fix code.

@codex review

---

gtanczyk:
All confirmed real, fixed in d0da32d:

- **Snapshot subscribers before delivering each tick**: confirmed. `for (const subscriber of state.subscribers)` iterates the live `Set`, and per spec a `Set` entry added during iteration *is* visited if unvisited — so a subscriber added synchronously from another subscriber's `onUpdate`/`onError` (itself running inside that same loop) got delivered to twice: once via `subscribeStudioStatus`'s own cached-value delivery, once by the still-running loop revisiting it. Fixed by snapshotting `[...state.subscribers]` before each loop.
- **Recheck generation after subscriber callbacks**: confirmed, traced the exact race. `tick()`'s `finally` mutated `ticking`/`lastTickAt` and called `schedule()` unconditionally. If a callback synchronously called `pokeStudioStatus` for the same key, that poke bumps the generation and starts its own tick — but the *original* tick's `finally` still ran after it, scheduling a stray extra timer at the now-current generation. If the poke's re-fetch outlasted the subscriber's own cadence, that stray timer fired first and started a third, redundant tick. Fixed by having the `finally` check `superseded()` (already used after the fetch) before touching shared state — a superseded tick now leaves cleanup entirely to whichever generation currently owns the key.
- **Remove prose-debt allowances for new files**: confirmed — I was resealing the baseline with `--force` instead of following `docs/contributing-for-agents.md`'s "new files start at 0" rule and just writing one-liners from the start. Rewrote every comment in both files as compliant single `//` lines (no `/** */` blocks, no stacked paragraphs, ≤12 words each) and resealed; both files now carry **zero** baseline debt.

Added one regression test for each of the two concurrency findings (verified each fails against the pre-fix code, passes against the fix, by temporarily reverting just those two changes and re-running).

@codex review

---

PR #1099, opened 2026-08-30
Make personal access token expiry calculation testable with clock injection

gtanczyk:
## Summary
This change makes the personal access token expiry calculation deterministic and testable by allowing injection of a custom clock function. Previously, the token info endpoint used `Date.now()` directly, making tests flaky when they ran across day boundaries.

## Key Changes
- Added `now?: () => number` option to `AuthPluginOptions` to allow clock injection in the auth plugin
- Added `accessTokenRoutes?: Partial<AccessTokenRoutesOptions>` to `BuildAppOptions` to pass clock configuration through the app builder
- Updated the token info endpoint to use the injected clock instead of calling `Date.now()` directly
- Modified test helper `appWith()` to accept and forward a `now` function to both the auth plugin and access token routes
- Updated test to use a fixed clock timestamp for both token minting and info retrieval, eliminating day-boundary flakiness

## Implementation Details
- The same clock instance is passed to both the minting path and the `/api/auth/token-info` endpoint to ensure consistency
- The remaining lifetime calculation uses `Math.floor()` to convert milliseconds to whole days, so using different clock readings can cause off-by-one errors
- The test now captures `Date.now()` once and reuses it for both operations, ensuring the expiry calculation is deterministic
- Renamed a local variable from `now` to `nowSeconds` in the session renewal logic to avoid shadowing the new `now` function parameter

https://claude.ai/code/session_01XzonyP8KAt66ZxFXR6qMAo


gtanczyk:
@codex

PR #1100, opened 2026-08-30
Phase 4: migrate three fixed-cadence Studio pollers onto studioStatusStore

gtanczyk:
Phase 4 (item 8) of the north-star plan — web restructure. Follow-up to #1098, which added `studioStatusStore.ts` but wired it into nothing. This PR migrates the three lowest-risk, fixed-cadence consumers onto it — the two remaining (`useStudioStatusPoll.ts`, `SubmissionStatusView.tsx`) are separate follow-up PRs, in risk order.

## What changed

`StudioDetailsBuildProgress`, `StudioDetailsMedia`, and `StudioShotToasts` each ran their own `setInterval` loop calling `getSubmissionStatus` directly (at 10s, 10s, and 12s respectively). All three now subscribe via `subscribeStudioStatus(token, locale, subscriber)` instead — a token/locale with several of these mounted at once (a common case: Details media + the near-Play toast stack + the checklist, all showing the same round) now shares one fetch and one schedule instead of three independent ones.

- `StudioDetailsBuildProgress`'s `refreshNonce` hack (bumped by `onSealed` to force an immediate re-poll instead of waiting out the full interval) is replaced by `pokeStudioStatus(token, locale)` — the store's built-in mechanism for exactly this.
- Per-consumer error handling is preserved exactly as it was: `StudioDetailsMedia` keeps the last-known media on a failed poll; `StudioShotToasts` clears it. Only the polling mechanism changed, not the error-handling semantics.
- `StudioDetailsBuildProgress`'s `status`/`loaded` persistence-across-token-change quirk (a component switching to a new token doesn't reset to a loading state until the new fetch resolves) is preserved unchanged — out of scope for this migration.

## Test changes

The three affected tests' fixtures previously reused the literal token `'tok'` across multiple tests within the same file. Since `studioStatusStore` is a module-level singleton keyed by `(token, locale)`, reusing a token across tests in the same file risks one test's cached poll state leaking into the next (the store doesn't get reset between tests). Gave each test its own unique token to keep them independent — no behavioral assertions changed.

## Verification

- `tsc --noEmit` clean
- `apps/web` suite: 219/219 test files, 1864 tests (all pre-existing tests for the three components pass unchanged in substance)
- Repo `npm run lint`: 0 errors
- Production build succeeds

---
_Generated by [Claude Code](https://claude.ai/code/session_01VaPzGSyNfZ7nvGcJgYNSdL)_

PR #739, opened 2026-08-10
Invert Creator Studio: full-bleed game stage, chat as a floating rail

gtanczyk:
Implements the game-first layout from studio-game-first-implementation-plan.md
as one build. The game is now an always-mounted, full-bleed stage; the thread,
Details, and Edit are layers over it rather than tab swaps or a 316px preview.

- useStageSource: single fetch of preview/channel/published HTML, feeding both
  the stage and the ribbon so they can never disagree about what is on screen.
- StudioStage: watch posture (muted, inert, frozen when hidden) and play
  posture (absorbs the old StudioPlaytestPanel theater), a swap policy that
  never yanks a build mid-run, 10-minute idle throttle, and crash/drew-nothing
  detection via the existing gamePlayer error bridge (no new bridge message).
- StudioStrip replaces the old detail head; StudioVersionRibbon is the
  non-dismissible honesty organ (identity / worst exception / check depth);
  StudioChatRail is the floating thread rail on desktop and a 3-detent bottom
  sheet on phone, keeping SubmissionStatusView mounted (not unmounted) when
  collapsed so its poll and any unsent composer text survive.
- StudioStageCard covers the no-build-yet state; StudioFullBleed is the named
  way back whenever Details/Edit/the shelf covers the stage (the ground-state
  rule), asserted directly in CreatorStudioView.test.tsx.
- router.ts: StudioTab drops 'playtest' (play is a stage posture now); the old
  /playtest URL still resolves, carrying its meaning via a new `posture` field
  on the studio route.
- Deletes StudioLivePreview.tsx and StudioPlaytestPanel.tsx, fully absorbed.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_012nTCrJ3ZA2SnJSJkayHh8f

PR #1101, opened 2026-08-30
Phase 4: migrate useStudioStatusPoll onto studioStatusStore

gtanczyk:
Phase 4 (item 8) of the north-star plan — web restructure. Follow-up to #1100 (three fixed-cadence pollers). This migrates `useStudioStatusPoll` — the fourth of five consumers, and the first with an adaptive cadence. `SubmissionStatusView.tsx` remains the final, highest-risk follow-up (owns status-transition side effects).

## What changed

`useStudioStatusPoll`'s own `setTimeout`-based poll loop is replaced by `subscribeStudioStatus`/`pokeStudioStatus`:

- Its adaptive cadence (`delayFor`, based on the latest status) is now the subscriber's `intervalMs`.
- Its "stop polling entirely on a 400 (invalid token)" special case is now `intervalMs` returning `null` when the latest error's status is 400 — the thread already surfaces that error, so nothing else should keep retrying it.
- Its visibility/focus/pageshow wake-up listeners (throttled-tab timers can stall for minutes) now call `pokeStudioStatus` instead of re-fetching directly — same effect (forces an immediate re-poll ahead of schedule, even past a stopped 400 state), routed through the shared store.
- The render-phase token-change bailout (`if (token !== statusToken) { ... }`, from the Codex review on #739) is untouched — unrelated to the polling mechanism.

## A real regression this surfaced, and its fix

Migrating this hook made the Stage's status poll share a `(token, locale)` key with the Details panes migrated in #1100 for the first time. `subscribeStudioStatus`'s "a later subscriber gets the cached value, no forced extra fetch" behavior is correct and already tested for the store itself — but it silently broke these consumers' original "always fetch fresh on mount" semantics once a second consumer could be sharing their key: switching to Details → Media while the Stage's own poll was already active on that token would show stale cached media instead of fetching current data. Caught by `CreatorStudioView.handoff.test.tsx`, which exercises exactly this handoff-then-navigate sequence.

Fixed by adding an explicit `forceFreshOnMount` option to `subscribeStudioStatus` (only forces a tick when joining an *idle* poll — a still-in-flight fetch is still shared, so the "concurrent subscribers share one fetch" behavior is untouched) and opting all four fixed-cadence/adaptive consumers into it, restoring their original semantics while keeping the schedule/fetch consolidation.

## Verification

- `tsc --noEmit` clean
- `apps/web` suite: 219/219 test files, 1866 tests, including 2 new regression tests for `forceFreshOnMount` (a later subscriber to an idle poll gets a fresh fetch; the very first subscriber doesn't get a redundant double-fetch)
- Repo `npm run lint`: 0 errors
- Production build succeeds

---


gtanczyk:
@codex review

---

PR #1107, opened 2026-08-30
Phase 4: styles.css split batch 3 — catalog rail surface

gtanczyk:
Phase 4 of the north-star plan — continuing the `styles.css` split, batch 3 of the "per surface" item. Batch 1 (#1105) extracted tokens/party, batch 2 (#1106) extracted the review desk; this batch extracts `CatalogRail.tsx`'s CSS.

## What changed

- **`surfaces/catalog/catalog-rail.css`** (236 lines): `.catalog-rail-*`, `.rail-card*` and its `pointer: coarse`/`max-width: 768px` media overrides. Imported only from `CatalogRail.tsx` — picked up transitively by every consumer (`ArcadeCatalog.tsx`, `PartyPage.tsx`, `CreatePage.tsx`) via the module graph, no extra wiring needed since they all render `<CatalogRail>` rather than duplicating its markup.
- Verified every selector in the candidate range individually before extracting (lesson from batches 1–2: don't trust range boundaries). This caught a real near-miss: an `#hero-prompt` ID selector — owned by `App.tsx`/`CreatePage.tsx`, nothing to do with the catalog rail — sandwiched between `.catalog-shelf` and `.catalog-rail-track`. A naive `grep '^\.'` scan (matching only class selectors) would have missed it entirely and swept it into the new file by mistake. Left it in `styles.css`.
- Deliberately **out of scope**: `CatalogRail.tsx` also owns `.featured-game-*` (a separate, non-contiguous block earlier in `styles.css`) and `ArcadeCatalog.tsx`'s much larger, more entangled CSS footprint (`.arcade-*`, `.catalog-toolbar`/`-sort-*`/`-grid`/`-card`/`-moments`, plus several non-obviously-named classes like `.card-title`/`.genre-pill` that turned out to be `ArcadeCatalog`-owned despite generic names) — that surface alone is already ~650 lines before any further splitting and needs its own multi-file batch.
- Moved `styles.catalogRail.test.ts` → `surfaces/catalog/catalog-rail.css.test.ts`, now reading `catalog-rail.css`.
- Raised `CatalogRail.tsx`'s module-size ceiling by exactly 1 line (407→408) for the new `import './catalog-rail.css'` line — scoped `--write --force`.

## Verification

- Content move verified exact: non-blank-line multiset of `styles.css` (pre-split) matches `styles.css` + `catalog-rail.css` (post-split) with zero missing/extra lines, accounting for three multi-line `/* */` comments deliberately condensed to one-liners during the move.
[...]

PR #1108, opened 2026-08-30
Phase 4: styles.css split batch 4 — arcade catalog surface

gtanczyk:
Phase 4 of the north-star plan — continuing the `styles.css` split, batch 4 of the "per surface" item. Batches 1–3 (#1105, #1106, #1107) extracted tokens/party, review desk, and the catalog rail; this batch extracts `ArcadeCatalog.tsx`'s CSS — the largest and most entangled of the remaining catalog-adjacent surfaces.

## What changed

- **`surfaces/catalog/catalog-toolbar.css`** (336 lines): header, jumpbar, pager, toolbar, sort menu — `.arcade-*`, `#browse-everything`/`#play-anchor`/`#party-rail`, `.catalog-jumpbar`, `.jump-chip*`, `.catalog-pager*`, `.catalog-toolbar`, `.catalog-sort-*`.
- **`surfaces/catalog/catalog.css`** (302 lines): grid, cards, media, previews, moments — `.catalog-grid`/`-card`/`-media`/`-preview*`/`-moments`/`-moment`/`-overlay`, `.preview-toggle`, `.touch-warning-pill`, `.genre-pill`, `.card-title`/`-copy`/`-author`/`-actions`. Split into two files rather than one because the combined content is ~650 lines, well over the 500-line hard cap for new files (same pattern as batch 1's `party.css`→`party.css`+`controller.css` split).
- Both imported only from `ArcadeCatalog.tsx`.
- Verified every selector individually before extracting — including a **full-file** grep for the class prefix, not just the initially-found range, per the lesson from batch 2's non-contiguous mobile-override bug. Found and fixed the same bug class again: a stray `@media (max-width: 768px)` block overriding `.catalog-grid`/`.catalog-toolbar`/`.catalog-filter-trigger`/`.catalog-sort-trigger` lived far from the main chunk, inside a large media block shared with many unrelated surfaces. Moved those three rules into their own file-local `@media` blocks (appended after the base rules) so they always win the cascade regardless of load order elsewhere — with a regression test.
- Also caught (and correctly *kept in scope*, unlike batch 3's `#hero-prompt`) an `#browse-everything`/`#play-anchor`/`#party-rail` ID-selector group: verified against the actual JSX that all three are rendered by `ArcadeCatalog.tsx` itself.
- Deliberately **out of scope**: `.catalog-refresh-error*`/`.catalog-state`, which are shared with `PartyPage.tsx`/`SubmissionStatusView.tsx` — same pattern as batch 2's `propose-*`/`proposal-*`.
- Moved `styles.catalogGrid.test.ts` → `surfaces/catalog/catalog.css.test.ts`, now reading `catalog.css`/`catalog-toolbar.css`. It carried 44 words of pre-existing comment-prose debt in the baseline; since the file was being touched anyway, shrunk every comment to a compliant one-liner rather than carrying the debt forward, per the repo's convention.

## Verification

- Content move verified exact: non-blank-line multiset of `styles.css` (pre-split) matches `styles.css` + both new files (post-split) with zero missing lines. The only "extra" is the intentional structural change — the shared media block became two smaller, file-local ones (2 new open/close brace pairs), accounted for explicitly.
- Cross-checked every class in both new files against the whole `apps/web/src` tree; ruled out several substring false positives (`rail-card-title`/`review-card-title`/`studio-stage-card-title`/`status-play-card-title` vs. bare `.card-title`, etc.) before trusting the "ArcadeCatalog-exclusive" conclusion.
- `tsc --noEmit` clean
- `apps/web` suite: 221/221 test files, 1878 tests
- `npm run comment-prose`: clean, no baseline debt (existing debt eliminated, not carried)
- `npm run module-size`: clean, both new files well under the 500-line hard cap
- Repo `npm run lint`: 0 errors
- Production build succeeds

---


gtanczyk:
@codex review

---

PR #1124, opened 2026-09-01
Phase 4: styles.css split — batch E (shared EditorKit grammar)

gtanczyk:
The last cluster the surface-scoped batches could not take, and the reason it waited: the `editor-*` grammar is genuinely dual-owned, so neither `surfaces/studio/` nor the remix side could claim it. Batches A, B and D each deferred it for that reason.

`styles.css`: 10,014 → **9,405 lines**.

## What moved

| File | Lines | Contents |
|---|---|---|
| `apps/web/src/editor-kit.css` | 394 | panel chrome, board, layers, cells, the path painter, tiles and palette |
| `apps/web/src/editor-kit-side.css` | 224 | collection picker, item list, the assist composer, property sheet, live checks |

Split in two **only** to stay under the 500-line new-file cap (the grammar is 608 lines as one block); both files are imported by both roots.

## Ownership

The grammar has exactly two roots — `surfaces/studio/EditorPanel.tsx` and `RemixPainter.tsx` — and the three surfaces beneath them (`EditorSurface.tsx`, `PathPainter.tsx`, `LayeredEditorSurface.tsx`) are each imported by **both**. So both roots import the kit and the three children need no import of their own: every render path reaches them through one of the two. That is why the file sits at the top level rather than under either surface.

## Cascade correctness

- **No ordering hazard.** Every `editor-*` rule lived in one contiguous block and moved together — verified by grepping for `editor-*` rules outside the range (the only hit, `.creator-profile-editor-quiet`, is an unrelated class that stays put).
- The only external references to this grammar live in the already-extracted `remix-editor-stage.css` and `remix-painter.css`, and all are **higher-specificity descendants** (`.remix-editor-stage.is-focus-play .editor-palette`), which win regardless of load order. `editor-controller-*` in `surfaces/studio/editor-controller.css` is a separate namespace with no selector overlap.
- The grammar's trailing `@media (max-width: 800px)` block straddled the split point (`.editor-body`/`.editor-layered-remix-grid` on the canvas side, `.editor-side` on the rail side). It was divided with **each override kept in the same file as its base, after it** — preserving the base-then-override order the cascade depends on.
- Co-applied classes were checked too: `game-page-editor-badge` and `creator-profile-editor-quiet` matched a naive `editor-` grep but are not part of this grammar and do not move.

## Test updates

Three CSS-content suites assert on these rules and now read the kit alongside their existing sources: `styles.editorSurface.test.ts`, `styles.studioEditor.test.ts` (both already concatenated `styles.css` + `studio-stage.css`), and `styles.gamePageRemix.test.ts` (its path-painter assertions now read the kit; its unrelated phone-slice logic was checked and is byte-identical before and after).

## Verification

- **Content-preservation diff** (`styles.css` before vs `styles.css` + both kit files after): **zero CSS lost**; extras are the two file headers and the one duplicated `@media` wrapper from splitting that block.
- Brace balance 0 in all three files; no doubled blank lines; both new files ≤ 500 lines.
- `prettier --check`, `tsc --noEmit`, `module-size`, `comment-prose`: clean.
- Full `apps/web` vitest suite: 221 files / **1885 tests passing**.
- `npm run lint`: 0 errors (131 pre-existing warnings). `npm run build --workspace apps/web`: succeeds.

## Remaining

A dead-CSS sweep for the clusters confirmed to have zero renderers, now collected in `styles.css`: `.studio-detail-head`/`.studio-status-pill`, `.studio-slug`, `.studio-tab-panel`, `.studio-improve`, `.studio-build .status-panel.is-embedded`, `.my-games-*`, `.studio-chat-rail-pill`, and the ~270-line playtest theater run. That is queued as the follow-up.

---


gtanczyk:
@codex review

---

PR #1126, opened 2026-09-01
docs: rewrite architecture.md against the shipped system

gtanczyk:
`architecture.md` described a local-preview app built around a `MockGameGenerator` package and a `POST /api/generate-game` route, under a heading marked **"Current Local-Preview Architecture ✅"**. Neither exists:

- `POST /api/generate-game` has no definition anywhere in the repo.
- `packages/game-generator` is **not tracked in git at all** — `git ls-files` returns nothing for it. It survives on some working copies only as stale `dist/` output, which is what makes it look present.

The second half described the games-repo model as a 🚧 *target*, when it has been the shipped system for months.

## What the new document covers

- **Workspaces as they are** — `apps/{api,web,world,e2e}`, `packages/{contract,zone-core}`, which workspaces actually consume contract, and the three bundled fixture games that make local dev work with no keys.
- **The sandboxed-iframe execution model** — unchanged, still the load-bearing safety decision, citing the two tests that assert the attribute.
- **`apps/api`'s nine buckets** — `store/**` as `platform`, the two unmapped files, and what the boundary rule does and does not enforce, including platform's exemption as an importer and the type-only exemption.
- **`apps/web`'s `core/` + `surfaces/` split** — the four lazy route chunks and the `cssCodeSplit` ordering hazard that follows from them.
- **Deployment** — three Cloud Run services, and why the app's `--max-instances 1` is conditional on the relay being in-process.
- **The repo's four checks**, correctly classified: two baseline ratchets, one exact invariant, one visibility-only.

## Two corrections of substance

**Delivery is not a merged PR.** The docs contradicted each other — `README.md` said "games-repo issue → coding agent PR → human review", `agent-adapters.md` said delivery never travels as a merged PR. The code settles it: `createIssue` has **zero production callers** and `managed-provider-copilot.ts` passes `createPullRequest: false`. A round is dispatched to an agent backend and delivered via `submit_sources`.

**A green gate does not publish.** `delivery/gate-runner.ts` says so in its header: *"This never publishes. It records a verdict; a human still approves."* The registry write is the admin-only `POST /api/admin/jobs/:jobId/publish`, recorded `by: 'operator'`. Neither lane lets a creator publish.

## The fact no document had: two catalog lanes

| | Store lane | Repo lane |
|---|---|---|
| System of record | Publication registry + version sources in GCS | `games/<slug>/` in the games repo |
| Population | Games built through the platform | The ~98 older git games |
| Sealed by | The gate, as `bundle.html` | The snapshot bake |
| An accepted change | Adopted onto an owner-owned job | Landed as a branch + PR by the apply-bot |
| Going live | Operator approval, both lanes | Operator approval, both lanes |

A slug with **no publication record is a repo-lane game** — the literal test in `community/proposal-base.ts`, which resolves the difference *for a proposal's source base*; the serving paths decide independently. The two endpoints order the lanes **differently**: `/api/games/:slug` is store-first; `/api/catalog` is repo-first and appends only untaken slugs, which is the mechanism behind "the repo lane wins catalog ties".

`AGENTS.md`'s "Current architecture" section still described the repo lane as the whole story, so it is updated here too — it is the file every agent reads first.

## Review history — worth reading before trusting a doc rewrite

**Eight codex passes. 19 findings. All 19 valid, all fixed.** Findings per round: 4, 4, 3, 2, 1, 1, 1, 0. Two were regressions I introduced while fixing earlier ones.

Nearly all of them were one mistake: **stating the general case and silently dropping the scope it depends on** — the lane, the env var, the workspace. At architecture altitude that feels like appropriate abstraction. But this document exists so agents don't re-derive the system from source, and an unscoped rule sends a reader to a code path that doesn't exist in their configuration. That is worse than no document, because it is confidently wrong.

The last finding was a different and sharper one: I had quoted `ensureOpenPullRequest`'s doc comment as if it described behaviour. It describes a PR cleanup that never happens — `closePullRequest` has no production caller. **A comment is evidence of intent, not of behaviour.** Every claim in the document is now backed by traced code rather than by prose about code, and the stale comment is flagged in-place so the next reader doesn't repeat it.

All 19 review threads are resolved.

## Verification

- `prettier --check` clean on every touched file. Three files under `docs/runbooks/` also fail it, identically on `master`, and are untouched here.
- Every relative link resolves; every code path named exists.
- A self-audit of every absolute quantifier in the file against the code, which caught two findings codex had not flagged.
- `npm run lint`: 0 errors (131 pre-existing warnings). Docs-only, so no code gate is affected.
- One `Lint, Test & Build` failure on `191c742` was a flake — every suite passed, then a silent exit 1 in `apps/world`'s slot. Demonstrated rather than assumed: `6591f68` passed the identical check with no code change, and `master` was green throughout.

## Noticed, not fixed here — both are code changes

- **`GitHubClient.createIssue` has no production caller.** Dead since the issue-driven build flow was abandoned.
- **`GitHubClient.closePullRequest` has no production caller**, and `ensureOpenPullRequest`'s doc comment describes a lifecycle the code never implements.


gtanczyk:
@codex review

---

gtanczyk:
All four findings verified against the code and fixed in `f1e7bc1`. Every one was correct.

**P1 — publish requires a human (`docs/README.md`).** Right, and this was the important one. `delivery/gate-runner.ts:19` says it outright: *"This never publishes. It records a verdict; a human still approves."* The registry write is `creation/job-admin-routes.ts:161` — `POST /api/admin/jobs/:jobId/publish`, gated by `isAdminSession`, recorded `by: 'operator', reason: 'approved'`. My "gate run → publish" deleted the moderation boundary from the plan of record. Restored it in the README row and the Create diagram, and added a callout stating the gate never publishes and the operator step must not be automated past.

**P1 — store lane vs. `AGENTS.md`.** You were right that the two contracts conflict. On which one is stale: the store lane is real — `catalog/catalog-routes.ts:151` `storePublishedGame`, the publication registry, and `community/proposal-base.ts:1-18`, which exists specifically to resolve the two lanes and states that a slug with no publication record is a repo-lane game. So the stale text is `AGENTS.md`'s "Current architecture" section, which still says production games *will* live in the games repo, full stop — pre-store-lane framing. Took your first option and rewrote it to describe both lanes, pointing at `architecture.md`. That file is what every agent reads first, so leaving it contradicting the architecture doc would have been precisely the failure this PR exists to fix. Thanks for catching it — it was outside the diff and I would have missed it.

**P2 — snapshot-only claim.** Confirmed. `catalog/game-play-route.ts:66` checks `storePublishedGame(slug)` **first** ("delivered games are never committed to the repo") and only falls through to the snapshot; `catalog-routes.ts:340` states store games skip the repo catalog and the bake. Scoped the claim to the repo lane and spelled out how each lane is served.

**P2 — `remix-view` also unmapped.** Confirmed by classifying every non-test `apps/api/src` source through `module-boundary-map.mjs`: `submissions` and `creation/remix-view`. Now names both, keeping the distinction that `submissions.ts` is unmapped deliberately while `remix-view.ts` is merely unclassified.

`prettier --check` clean, `npm run lint` 0 errors.

I could not reply on the individual threads or resolve them — the GitHub API is returning `rate limit already exceeded` for review-thread reads on this account right now. I'll resolve them on my next pass once the limit clears.

---

gtanczyk:
@codex review

Re-requesting on `f1e7bc1` — all four findings from the `1b57f75` pass are addressed, and the diff grew a file: `AGENTS.md`'s "Current architecture" section is rewritten to describe both catalog lanes, which is the fix for the P1 you raised about the two contracts conflicting. Worth a fresh look at that section in particular.

---

gtanczyk:
@codex review

All four from the `f1e7bc1` pass fixed in `8e7468b`. Verified each; all four were correct.

Three shared one root cause worth naming: I wrote the Workspaces, buckets and gates sections **before** discovering the two-lane split, and never re-read them once I had. So the doc asserted the repo lane as universal in three separate places while describing both lanes correctly in a fourth. That's the failure mode of rewriting a doc section by section, and your passes have caught every instance of it.

- **Workspaces / "games are not in this repo"** — scoped to the repo lane, with a pointer to the lanes section. You're right that as written it sends an agent hunting for a `games/<slug>/` that doesn't exist for platform-built games.
- **"Every file belongs to exactly one bucket"** — contradicted this same document eight lines later. Now "every *mapped* file", and the unmapped pair stays named.
- **`delivery/` "turning a merged game into something playable"** — reworded to staged sources → candidate versions → previews → gate, and it now says merge-back lives in `community/proposal-apply-bot.ts`.
- **All four gates called ratchets** — I ran it rather than reasoning about it: `npm run module-boundary` exits **0** with 131 warnings. Three gates block; that one is visibility-only except for `ENFORCED_BUCKETS`, where the main eslint pass runs it at `error`. Corrected rather than left as a guarantee the repo doesn't make.

CI was 8/8 green on `f1e7bc1`; this commit is docs-only on top of it.

Still unable to reply on or resolve the individual threads — the GraphQL review-thread endpoint has been returning `rate limit already exceeded` for this account for ~30 minutes, while REST reads and comments work fine. I'll resolve all eight threads once it clears.

---

gtanczyk:
@codex review

All three from the `8e7468b` pass fixed in `51d0c25`, all verified, all correct. That's 11 for 11 across three rounds.

Rather than hand you a fourth round of the same mistake, I audited the document myself first — grepped every absolute quantifier (`every`, `only`, `never`, `always`, `exactly`, `sole`, `the one`) and checked each against the code. That found **two more**, both mine:

- **`--max-instances 1` is conditional.** I wrote that the app service "is pinned to a single instance". It's `MAX_INSTANCES=1` only when `MP_RELAY_URL` is unset; with the relay split out, the same deploy step sets 4. The pin belongs to the in-process room registry, not to the service.
- **The assembler claim was understated.** `assembleGameHtml` is called by the repo-lane bake, the store-lane play path *and* `gate-runner.ts` — so "exactly one definition of what a served game is" is a stronger property than I'd written, and covers both lanes plus the gate checking the same artefact the player receives.

Your three:

- **"the owner publishes" survived in the lanes table** after I'd fixed that exact error in the README row and the Create diagram. `store.setPublication` has one production caller — the admin-only route, `by: 'operator'`. Corrected, plus an explicit sentence that neither lane lets a creator publish.
- **Play diagram still snapshot-only**, contradicting the prose directly beneath it. Added the store-lane edge in the order the route checks them.
- **Platform's boundary exemption.** Confirmed at `module-boundary.mjs:96` — a mapped `platform` importer returns immediately, necessarily, since the composition root reaches every registrar. Scoped to domain modules, and named the type-only exemption too.

Worth stating the root cause plainly: I rewrote this document section by section and discovered the two-lane split partway through, then never re-read the earlier sections against it. **Six of the eleven findings are that one mistake in different paragraphs.** A doc rewrite has no test suite, and this is what that costs.

---

gtanczyk:
@codex review

All four fixed in `191c742`. 15 findings, 15 valid.

**Two of those four were regressions I introduced in round 3**, so I treated this as root cause rather than another patch pass.

The root cause: I was writing caller-level claims from grep hits instead of tracing the path. Each round I patched the flagged sentence and the patch was wrong one level deeper. So this commit rewrites the affected sections at the precision I can actually defend, and I traced every claim end-to-end before writing it — including re-checking my own new text, which caught that `gate-runner.ts:441`'s "never bundle.html" is about the crash artefact and doesn't contradict `games-store.ts:13`'s *"bundle.html … produced by our gate, never uploaded"*.

- **Store-lane assembly attributed to the play route** — you're right. `storePublishedGame` fetches a pre-derived `bundle.html` and returns those bytes as-is; the `assembleGameHtml` call is reached only on a store miss, i.e. the repo-lane fallback. The doc now names the two *sealing* paths rather than the serving one.
- **One lane order claimed for two endpoints that disagree** — `/api/games/:slug` is store-first, `/api/catalog` is repo-first and appends only untaken slugs (`catalog-routes.ts:260`). That's the actual mechanism behind "repo wins catalog ties", which I'd asserted elsewhere without connecting it to this. Diagram scoped to the play endpoint; both orders stated; explicit warning not to generalise.
- **"Three blocking ratchets"** — correct, `env-manifest` is an exact invariant with no baseline and no `--write --force`. Regrouped as two ratchets, one exact invariant, one visibility-only.
- **Fixtures** — `apps/api/fixtures/games-repo/games/` holds three complete games that `local-games-repo.ts` serves when no checkout is configured, which is what makes local dev work with no keys. Now called out as the deliberate exception.

Standing assessment of this PR: a document rewrite has no test suite, and the claims most worth making are the ones most easily wrong at one level deeper than you checked. Your passes have been the test suite. Thank you — genuinely.

---

gtanczyk:
**CI red on `191c742`: `Lint, Test & Build` failed. I do not believe it is this PR's, and here is the evidence before I re-run it once.**

What the log actually shows: every suite passes — `apps/web` 221 files / 1885 tests, `apps/world` 2 files / 20 tests — and then the process exits 1 **96 ms after world's summary**, with no error output and without `npm run test:rules` ever printing. So the non-zero came out of `npm run test --workspaces`, after all reported tests had gone green. `apps/world` is the last workspace in the run order.

Why it is not this change:

- **The diff is five markdown files.** It touches no code, no config, no test. It cannot affect `apps/world`'s websocket tests.
- **Four earlier heads of this same PR passed this exact check** — `1b57f75`, `f1e7bc1`, `8e7468b`, `51d0c25`, all equally docs-only.
- **`master` is green**, including this PR's base `0b0d6e5`.
- **`apps/world` passes locally on this branch**, exit 0, 20/20.
- `npm run test:rules` also passes locally, 60/60 — and in CI it never ran, so it is not the failure either.

The shape of it — a clean pass followed by a silent non-zero, in a suite whose slowest tests are a 1,031 ms "checksum about once a second" and a 2,335 ms websocket file — points at a teardown/unhandled-rejection race in `apps/world` under CI timing, not at anything in this diff.

Re-running the failed job now. That is my one re-run on this PR; if it fails a second time I will treat it as real and root-cause it rather than re-running again.

All 15 review threads from the four codex passes are now resolved.

---

gtanczyk:
@codex review

Both fixed in `e03f9d7`. **17 findings, 17 valid.** Both of these were the same failure mode a fifth time: a claim true of the main configuration, stated as if true of all of them.

- **"Neither lane assembles a game on the request path"** — you're right, and the contradiction was two lines from itself. `game-play-route.ts:84` gates on `snapshotReader`; with `GAMES_SNAPSHOT_BUCKET` unset there is no reader, so it falls through to `getGameSources` and assembles per request from the fixtures — the very local/dev configuration the next sentence describes. Scoped to snapshot-backed serving, with the local path named explicitly.
- **"`proposal-base.ts` is the one place the difference is resolved"** — overstated, including against its own source comment, which scopes that to a proposal's *source base*. The serving paths decide independently. Fixed in `architecture.md` and in `AGENTS.md`, where I had copied the same overstatement.

Re-reading my own edit before pushing also caught a "the serving paths **above**" that should read "below" — the lanes section precedes the flows section. Small, but it's the class of thing that would have cost another round.

On the failure mode itself, since it has now produced findings in five consecutive passes: my instinct when writing an architecture doc is to state the rule and omit the configuration it depends on, because the exceptions feel like noise at that altitude. In a document whose job is to be the thing agents trust without re-deriving, the configuration *is* the content — an unscoped rule sends someone to a code path that does not exist for them. I have stopped treating "scope it to the lane/config" as a nit.

CI note: the `Lint, Test & Build` red on `191c742` was a re-run I already justified and spent (all suites passed, then a silent exit 1 in `apps/world`'s slot; master green, four earlier heads of this PR green, passes locally). `e03f9d7` will get a fresh run regardless.

---

gtanczyk:
@codex review

Fixed in `6591f68`. **18 findings, 18 valid.**

"Shared by **every app**" is false for `apps/e2e` — no `@gamedevpl/contract` dependency, no imports from it, because it drives the deployed site over HTTP and needs no shared types. Verifying that also caught what the sentence got wrong in the other direction: `packages/zone-core` *does* depend on contract, which the line never said. It now names the four real consumers, exempts e2e with the reason, and keeps the narrower claim that survives — contract is the only workspace package `apps/web` imports.

Before pushing I swept the remaining absolute quantifiers in the file against the code rather than waiting for a seventh pass. Four checked out as written: platform being importable by every other bucket, all three Cloud Run services in `europe-west1`, both deploy paths threading the manifest, and the three `assembleGameHtml` callers.

Six passes, six times the same shape. Naming it precisely, since it is the actual lesson of this PR: **I was writing the general case and silently dropping its scope** — the lane, the env var, the workspace. At architecture altitude that reads like appropriate abstraction, but this document exists so agents do not re-derive the system from source, and an unscoped rule sends them to a path that does not exist in their configuration. That is worse than no document. Every one of your eighteen findings is an instance of it or of its cousin, claiming a caller I had grepped rather than traced.

---

gtanczyk:
@codex review

Fixed in `2914bfb`. **19 findings, 19 valid.**

This one is a *different* mistake from the previous six, and a worse one: **I quoted a source comment as if it described behaviour, without checking the code does it.**

I wrote that a branch or PR in the agent path is resumption mechanics "and the adapter closes it when the job finishes". That sentence came straight from `ensureOpenPullRequest`'s own doc comment. The shipped code does none of it, exactly as you say:

- `managed-provider-copilot.ts` passes `createPullRequest: false` — the agent path opens no PR at all.
- The single production caller of `ensureOpenPullRequest` is `community/proposal-apply-bot.ts`, the repo-lane merge-back.
- **`closePullRequest` has no production caller.** Nothing closes anything.
- The branch goes via `deleteWorkspace`, from `managed-backend.ts`'s `cleanup`, when a *previous* dispatch is released — resume, cancellation or abandonment — not at completion.

So the comment records an intent that was never implemented, and I laundered it into the plan of record. The doc now states the real lifecycle and carries a note that the comment is stale, so the next reader doesn't repeat my mistake by trusting it. Fixing the comment itself is a code change and stays out of this PR.

I then swept the three other source comments this document quotes and confirmed each against the code — `games-store.ts`'s "bundle.html produced by our gate" (stored at `gate-runner.ts:636/:640`), `gate-runner.ts`'s "This never publishes" (sole `setPublication` caller is the admin route), and `game-play-route.ts`'s "delivered games are never committed" (sole `createBranchWithFiles` caller is the apply-bot). Those three hold.

Worth stating for the record, because it generalises past this PR: **a comment is evidence of intent, not of behaviour.** I had been treating in-repo comments as a trusted source precisely because they are written by people who know the system — which is exactly why a stale one is dangerous. Every claim in this document is now backed by the code, not by prose about the code.

---

PR #1127, opened 2026-09-02
Phase 5: one frame limiter, one reconnect curve

gtanczyk:
Phase 5 in the north-star plan asks for a `realtime-core` package holding the frame limiter, caps, tickets, reconnect and a version registry. Measured against the code, **most of that is already single-source**:

| Phase 5 asks for | Actual state |
|---|---|
| Protocol versions declared once | ✅ Already — `MP_PROTOCOL_VERSION` and `ZONE_PROTOCOL_VERSION` each live once in `packages/contract` |
| Caps declared once | ✅ Already — `socket-limits.ts` holds all three |
| One ticket mint/verify | ⚠️ Zone tickets already shared via `zone-core/ticket.ts`; `mp.ts` keeps a separate room-ticket HMAC |
| One frame limiter | ❌ Two copies |
| One reconnect | ❌ Two, but only partly duplicated |

A new package would mostly have moved code that isn't duplicated, so this PR does the part that is. Scoped with the owner beforehand.

## Frame limiter — two copies, now one

`apps/api/src/realtime/mp.ts` and `apps/world/src/app.ts` held `FrameLimiter` classes that were **character-identical except for one parameter name** (`timestamp` vs `at`).

Replaced with `createFrameLimiter()` in `packages/contract/src/frame-limiter.ts` — beside the `MAX_SOCKET_FRAMES_PER_SECOND` cap it enforces. Both apps already read that cap from there, so the limiter had no business living anywhere else. A factory rather than a class because this package ships runtime functions and zero classes; a stateful class would cut against its grain.

## Reconnect — the curve, not the loop

`zoneClient.ts` and `roomClient.ts` are 54% identical by line, but that number is misleading and merging them would have been a false friend. Their timers, disposal rules and final-reason sets differ **deliberately**, and `zoneClient.ts`'s own header explains why: a party guest reconnecting takes a free slot, while a zone player rejoins a world that kept running without them — which is why `zone_full` is a queue to wait in rather than a closed door.

What *was* identical and could have drifted unnoticed is the retry budget and delay curve — `Math.min(500 * 2 ** (n - 1), 8000)`, ceiling 8. Only that is extracted, as `createReconnectBackoff()` in `apps/web/src/core/`.

Each client keeps everything else, **including the one real behavioural difference**: a spent budget disposes the zone client and does *not* dispose the room client. That is preserved exactly and now carries a comment at the line where it lives, rather than being quietly normalised away.

## No behaviour change

Both new units are tested — the curve, the default cap, the sliding window, budget exhaustion, and per-connection isolation.

## Verification

- `npm run lint`: **exit 0**.
- `tsc --noEmit` clean for `apps/api`, `apps/web`, `apps/world`.
- `npm run test`: **exit 0** across every workspace — contract 104 (+4), zone-core 96, api 3610, web 1889 (+4), world 20, rules 60.
- `npm run build`: succeeds.
- Two comment-prose baselines added **by exact delta** for the two new files; the tool refused an unscoped `--write` and nothing else was resealed.
- `packages/contract/src/gate-status.ts` fails `prettier --check` identically on `master` and is untouched here.

## What this leaves

The ratchet line "frame-limiter / HMAC-ticket implementations: 3 → 1" is now: frame limiter **1**, realtime ticket **2** (`zone-core` and `mp.ts`'s room ticket), reconnect curve **1**. Merging the room-ticket HMAC into the zone ticket primitive was considered and skipped — they carry different claims and scopes, and unifying them is a security-shaped change that deserves its own PR rather than riding along with a dedup.

---


gtanczyk:
@codex review

Two things worth your attention specifically, given how #1126 went:

1. **The reconnect extraction is deliberately partial.** `zoneClient.ts` and `roomClient.ts` are 54% identical, but I extracted only the backoff curve and attempt count — not the loop. Their disposal semantics genuinely differ (a spent budget disposes the zone client, not the room client) and I preserved that rather than normalising it. If you think the difference is a latent bug rather than intent, say so; I read it as intent, based on `zoneClient.ts`'s header about `zone_full` being a queue rather than a closed door.

2. **Claims in the description are traced, not grepped.** After #1126 I am not asserting caller relationships I have not followed end-to-end. If any claim above about what is or is not single-source doesn't hold, that's the thing to flag.

---

PR #1137, opened 2026-09-02
Phase 4: decompose CreatorStudioView into eight modules

gtanczyk:
`CreatorStudioView.tsx` was 2,247 lines — the orchestrator plus every panel it renders, seven sub-components deep. This splits the whole tail of the file out in one pass rather than a component at a time, leaving the orchestrator alone in the file.

## What moved

Every extracted symbol was in-file only, with zero consumers outside `CreatorStudioView.tsx` (verified by grep before the split).

| New module | Lines | Contents |
| --- | --- | --- |
| `studioTabs.ts` | 28 | `defaultTabFor`, `tabAvailable`, `resolveTab`, `studioAddress` |
| `studioHealth.ts` | 17 | `formatSeconds`, `healthFor` |
| `StudioShelf.tsx` | 131 | `StudioShelfControls`, `StudioShelfList` |
| `DraftShareControl.tsx` | 111 | `DraftShareControl` |
| `StudioStatsSection.tsx` | 168 | `StatsSection`, `PlayerReactions`, `WINDOWS`, `percent` |
| `StudioSuggestions.tsx` | 176 | `SuggestedImprovements`, `SuggestionContext`, `classLabel`, `DISMISS_REASON_KEYS` |
| `AutonomySetting.tsx` | 81 | `AutonomySetting`, `AUTONOMY_CHOICES` |
| `StudioDetailsPanel.tsx` | 319 | `DetailsPanel`, `DetailsPaneId`, `DetailsPaneDef` |

`CreatorStudioView.tsx` drops **2,247 → 1,192** lines, and its import block shrinks to what the orchestrator itself uses (16 previously-shared imports moved out with their code).

## Behaviour

Same components, same props, same render order, same CSS — with one deliberate exception, a defect that predates this PR and surfaced in review.

**`formatSeconds` rounded the remainder independently of the minutes**, so a non-integer duration could render an impossible seconds component:

| input | before | after |
| --- | --- | --- |
| `59.6` | `60s` | `1m` |
| `119.6` | `1m 60s` | `2m` |
| `3599.7` | `59m 60s` | `60m` |
| `90.4` | `1m 30s` | `1m 30s` |

The total is now rounded first, then split. This is user-visible on the studio overview and stats panes. `studioHealth.test.ts` covers it and `healthFor` — the extraction is what made them directly testable.

Two smaller type-level changes, no runtime effect: the contributions block reads `showContributions && game.slug` so the slug stays narrowed for `ContributionsSetting`, and the workspace pane guards on `game.slug` instead of asserting `game.slug as string`.

`apps/web/src/GameHealthView.tsx` holds an independent copy of `formatSeconds` carrying the same rounding bug. It's outside this diff, left for a follow-up that de-duplicates onto one shared helper.

## Ratchets

Prose comments on the moved code are rewritten as one-line `//` notes so each new file starts at zero comment-prose debt, and the `CreatorStudioView.tsx` baselines are ratcheted **down** to the new figures rather than left at the old ceiling:

- module-size: 2247 → 1192
- comment-prose: 2931 → 1806

No `--force`, no new-file budget granted.

## Verification

- `npm run lint` — 0 errors, 130 warnings (unchanged baseline); comment-prose, module-size, module-boundary, env-manifest all pass
- `npm test -w @gamedevpl/web` — 1,894 tests / 223 files pass, including all five `CreatorStudioView.*.test.tsx` suites
- `npm run build -w @gamedevpl/web` — clean; the lazy `CreatorStudioView` chunk still bundles the extracted modules together (no chunk-split regression)


gtanczyk:
@codex review

---

gtanczyk:
Both Copilot findings were real defects that predate this PR and moved with the code. Fixed in `5d7d108`.

**`formatSeconds` rounding** — confirmed by reproduction:

| input | before | after |
| --- | --- | --- |
| `59.6` | `60s` | `1m` |
| `119.6` | `1m 60s` | `2m` |
| `3599.7` | `59m 60s` | `60m` |
| `90.4` | `1m 30s` | `1m 30s` |

Rounding the remainder independently of the minutes is what produced the impossible seconds component. Now the total is rounded first, then split. Added `studioHealth.test.ts` covering this and `healthFor` — the extraction is what makes them directly testable, so the test is worth having beyond the fix itself.

**`game.slug as string`** — the workspace pane is only offered when `game.slug` is truthy, so the assertion was sound at runtime, but agreed it shouldn't be the compiler's problem to take on trust. Now `activePane === 'workspace' && game.slug ? … : null`, same as the sibling panes.

Note that this is a real behaviour change, small and in the correct direction: durations that previously rendered `1m 60s` now render `2m`. My PR description claimed no behaviour change; that now holds for everything except this fix.

`apps/web/src/GameHealthView.tsx` carries an independent copy of `formatSeconds` with the identical bug. It's a file this PR doesn't touch, so I've left it out rather than widen the diff — worth a follow-up to de-duplicate onto `studioHealth.ts`.

Verification on the new head: type-check clean, lint 0 errors / 130 warnings (unchanged), 1,894 web tests pass (223 files).

---

PR #1136, opened 2026-09-02
fix(web): remove obsolete editable badge from catalog and game details

gtanczyk:
## Summary
- Removes the "Edytowalna" / "Editable" badge and its tooltip (`t('catalog.editorBadge')`, `t('catalog.editorTooltip')`) across catalog surfaces:
  - `CatalogRail.tsx`: removes editor capability badge from game cards while retaining multiplayer badges.
  - `ArcadeCatalog.tsx`: removes editor capability badge from game cards in the catalog view.
  - `GameDetailPage.tsx` and `GamePage.tsx`: removes redundant editable badge in the kicker/breadcrumb next to the genre (the explicit action button to edit content remains intact).
  - `en.json` and `pl.json`: removes unused `editorBadge` and `editorTooltip` translation keys.
- Updates unit tests in `CatalogRail.test.tsx` and `ArcadeCatalog.test.tsx` to verify the badge is no longer rendered and keeps file line ceilings intact.

## Test plan
- [x] `npm run type-check` — passed
- [x] `npm run lint` — passed (0 errors, within module size & comment prose baselines)
- [x] `npm run test` — passed (457 test files, 5513 tests)
- [x] `npm run build` — passed

PR #1139, opened 2026-09-02
Phase 4: split CodeMirrorEditor into eight modules

gtanczyk:
`CodeMirrorEditor.tsx` was 993 lines, of which only the last ~140 were the React component. Everything above it was CodeMirror wiring — widgets, state fields, view plugins, extension factories — that has nothing to do with React and was unreachable from a test while it sat inside the component's module.

## Split along what each piece configures

| New module | Lines | Contents |
| --- | --- | --- |
| `codeMirrorTypes.ts` | 12 | `CodeMirrorDiagnostic`, `CodeMirrorLanguageService`, `GotoDefinitionHandler`, `FetchGhostText` |
| `codeMirrorLanguage.ts` | 24 | `languageExtension` |
| `codeMirrorDiagnostics.ts` | 21 | `toCmDiagnostics`, `tsAdvisoryLintSource` |
| `codeMirrorColorPicker.ts` | 80 | `ColorSwatchWidget`, `colorPickerExtension` |
| `codeMirrorCompletion.ts` | 107 | `measuredTsAutocomplete`, `completionVisibilityExtension`, `CompletionTracker` |
| `codeMirrorTheme.ts` | 166 | `darkHighlight`, `darkChrome` |
| `codeMirrorHover.ts` | 177 | modifier-aware hover, `modifierHoverState`, `makeGotoHandler` |
| `codeMirrorGhostText.ts` | 218 | TA-02 ghost text end to end — field, widgets, fetch plugin, keymap |

`CodeMirrorEditor.tsx` drops **993 → 221** lines and keeps exactly three things: the props type, `languageServiceExtensions` (the one place those pieces are composed), and the component.

This follows the naming already established by the file's siblings — `codeMirrorSearchPanel.ts`, `codeMirrorColors.ts`, `codeSurfaceEditorState.ts`.

## Types

`CodeMirrorDiagnostic` and `CodeMirrorLanguageService` move to `codeMirrorTypes.ts` and are re-exported from `CodeMirrorEditor.tsx`, so `CodeSurface.tsx`'s `import type { CodeMirrorDiagnostic }` needs no change.

`FetchGhostText` and `GotoDefinitionHandler` were each written out twice — once structurally in the props type, once as a named type further down the file. Both now name the shared type, so the props and the implementation can't drift.

## Behaviour

No change. Same extensions, composed in the same order by the same function.

## Verification

- `npm run lint` — 0 errors; comment-prose, module-size, module-boundary and env-manifest all pass. All eight new modules start at zero comment-prose debt, and the `CodeMirrorEditor.tsx` module-size baseline is ratcheted **down** 993 → 221 (no `--force`)
- `npm test -w @gamedevpl/web` — 1,894 tests / 223 files pass, including `CodeMirrorEditor.test.tsx` and the five `CodeSurface.*` suites that mock it
- `npm run build -w @gamedevpl/web` — clean; `CodeMirrorEditor` remains one lazy chunk

The 132 `module-boundary` warnings are master's current baseline (they arrived with #1134–#1136); this diff touches no `apps/api` file, and I confirmed the same count on a clean master tree.

---


gtanczyk:
@codex review

---

gtanczyk:
Valid finding, fixed in `37daa1a` — with one deliberate departure from the suggested shape, which I want to be explicit about rather than bury.

**Taken:** the affordance is now `<button type="button">`. A `<span role="button">` claims interactive semantics it doesn't have and borrows them from ARIA; the native element carries them, and `aria-label` is the only attribute left doing work.

**Departed from:** it keeps `tabIndex = -1`, so it is still not in the tab order.

The reasoning: this widget renders **inside the editor's `contenteditable`**, and it exists only as the pointer fallback — the comment above it records that Tab was unreachable on iOS. `Tab` is already bound to `acceptGhostText` in the same extension's keymap (`codeMirrorGhostText.ts`, the `Tab` binding in `makeGhostTextExtension`), so the keyboard path to this action exists and isn't this element's responsibility. Making it a tab stop would put a focusable control inside the document that appears and disappears with the suggestion, competing with the very Tab binding that already accepts it.

`tabIndex = -1` doesn't hide it from assistive tech — VoiceOver and TalkBack navigate the accessibility tree, not the tab order, so the touch users this exists for still reach and activate it. It only keeps it out of sequential keyboard navigation, where the equivalent command is already bound.

If you'd still prefer it focusable, say so and I'll make the change — but I didn't want to introduce a tab stop into the editor without flagging it.

Two supporting changes:

- **CSS:** added `font-family: inherit` to `.cm-ghost-text-accept`. A `<button>` takes the UA control font where the `<span>` inherited the editor's; without this the glyph would have shifted.
- **Test:** added `codeMirrorGhostText.test.tsx`, which drives the real path — render, doc change, 300ms debounce, resolved proposal — then asserts the rendered element is a `BUTTON` with `type="button"`, `tabIndex === -1`, and no `role`. I verified it's a genuine guard: it fails against the old `<span role="button">` markup and passes against the new one.

It's a new file rather than an addition to `CodeMirrorEditor.test.tsx` because that file is at its module-size ceiling (136 lines) and the ratchet only allows files to shrink. Raising it with `--force` for a test I could just as well put beside the module it covers seemed like the wrong trade.

Verification on the new head: type-check clean, lint 0 errors, 1,895 web tests pass, build clean.

---

PR #1140, opened 2026-09-03
Phase 4: decompose the three studio mega-components

gtanczyk:
`CodeSurface`, `RemixPanel` and `EditorPanel` are each one component holding a whole state machine. Unlike `CreatorStudioView` (#1137) and `CodeMirrorEditor` (#1139), they have no detachable sub-components sitting at the bottom of the file — what makes them large is interleaved state. So this pulls out cohesive slices of that state as hooks, alongside the panels and helpers that were genuinely separable.

## CodeSurface — 1,608 → 1,355

| New module | Lines | What it owns |
| --- | --- | --- |
| `useCodeSurfaceLanguageService.ts` | 73 | The TS worker's whole lifecycle: boot, queued updates while it boots, kit declaration, teardown on slug change |
| `CodeSurfaceAgentMode.tsx` | 134 | The WebMCP dialog, owning its own console input/history/busy state |
| `CodeSurfaceKitViewer.tsx` | 70 | The kit `.d.ts` overlay, its scroll-to-target and Escape effects |
| `codeSurfaceHelpers.ts` | 67 | Timing constants, the four state unions, and the four pure functions |

`agentModeEnabled` deliberately **stays in the parent**. Moving it into the dialog would have scoped the WebMCP bridge registration to the dialog's lifetime, so the bridge would stop when the dialog closed. It's the dialog's *console* state that's local, not the opt-in.

## RemixPanel — 1,799 → 1,690

| New module | Lines | What it owns |
| --- | --- | --- |
| `useRemixGrip.ts` | 62 | The sheet handle's drag/tap gesture, both refs and all four pointer handlers |
| `RemixChatParts.tsx` | 175 | `RemixTranscript`, `RemixActionRow`, `RemixKeepOffer` — components taking explicit props instead of closing over a dozen locals |

## EditorPanel — 1,124 → 1,084

`editorPanelHelpers.ts` (56): `useLabel`, `mergeDraft`, and the four spec narrowers.

## Behaviour

**One behaviour change, and it is a fix for a regression this PR introduced and then corrected in review** (`c86f558`).

The first cut of `CodeSurfaceAgentMode` moved the console input, history and busy flag into a component rendered under `{agentModeOpen ? … : null}`. Those states previously lived in the always-mounted `CodeSurface`, so the conditional mount discarded them on every close: a prepared command was lost, the history was lost — contradicting `AGENT_GUIDE`'s promise that *"every past command and result stays visible there too"* — and `agentConsoleBusy` reset to `false`, so reopening during an in-flight run re-enabled Run and allowed the same mutating tool call to be submitted twice.

The dialog now mounts unconditionally and returns `null` while closed, keeping ownership of its state without owning its visibility; the Escape listener is guarded on `open`. `CodeSurfaceAgentMode.test.tsx` types a command, closes, reopens and asserts the text survived — verified to fail against the conditional mount.

Everything else is unchanged: same components, same props, same render order.

## What I deliberately did not do

**The drafts/autosave machine in CodeSurface stays as it is.** Its autosave timer feeds the typecheck debounce feeds the preview rebuild feeds the rebuild cooldown, and all four share `saveTimersRef`/`savingPromisesRef`/`draftsRef`. A hook around it would need eight callbacks passed in and would relocate the coupling rather than break it — that's a redesign, not an extraction. It's the main reason CodeSurface lands at 1,355 rather than lower.

## Verification

- `npm run lint` — 0 errors; comment-prose, module-size, module-boundary and env-manifest all pass. All seven new modules start at zero comment-prose debt, and the module-size baselines for all three files are ratcheted **down** (no `--force`)
- `npm test -w @gamedevpl/web` — 1,897 tests / 225 files pass
- `npm run build -w @gamedevpl/web` — clean

Two further defects the compiler and lint caught mid-split, fixed before the first push: the kit viewer's map callback shadowed the `line` prop (every row would have compared against the wrong value; it is now `activeLine`), and two dependency arrays needed the ref object from the new hook.

The 132 `module-boundary` warnings are master's baseline; this diff touches no `apps/api` file.


gtanczyk:
@codex review

---

gtanczyk:
Correct, and a regression I introduced. Fixed in `c86f558`.

The three console states moved from `CodeSurface` — always mounted — into a component rendered as `{agentModeOpen ? <CodeSurfaceAgentMode /> : null}`. Closing therefore unmounted it and discarded all three. Confirming the consequences you named:

1. **Prepared command lost.** Typed text gone on any close — backdrop click, close button, or Escape.
2. **History lost**, and this one contradicts a promise made to the agent: `webmcp.ts` line 115 of `AGENT_GUIDE` reads *"every past command and result stays visible there too"*. `AGENT_CONSOLE_HISTORY_LIMIT = 20` exists precisely to bound a history that was supposed to persist.
3. **`agentConsoleBusy` reset to `false`**, so reopening during an in-flight run re-enabled Run and allowed the same mutating tool call to be submitted twice. That's the serious one.

**Fix:** the dialog now mounts unconditionally and returns `null` while closed — it keeps owning its state without owning its visibility. The Escape listener is guarded on `open` so it only binds while the dialog is actually up, rather than firing `onClose` on every Escape in the editor.

I considered lifting the three states back into `CodeSurface` and passing them down, but that would have undone most of the extraction's point for the same result.

**Test:** `CodeSurfaceAgentMode.test.tsx` types a command, closes, reopens, and asserts the text survived. I verified it's a real guard rather than a passing formality — reverted to the conditional mount and the test fails; restored and it passes.

I've also corrected the PR description. It claimed "no behaviour change", which was wrong: this was a behaviour change, just an unintended one.

Verification on the new head: type-check clean, lint 0 errors, 1,897 web tests pass, build clean.

---

gtanczyk:
Correct on the mechanism and on the consequence. Fixed in `cfa29fd`.

**Provenance first:** this one is *not* introduced by the extraction. `git show e621149:apps/web/src/surfaces/studio/CodeSurface.tsx` has the same two lines (556 and 564) — the ref assignment followed by `setLanguageServiceReady(service !== null)`. The hook inherited it verbatim. Fixing it here anyway, because it's the new hook's own internal contract and the fix is local to it.

**The consequence, traced:** `kitDeclaration` has one render-time consumer that matters — `CodeSurface.tsx:954` pushes the *"open kit declaration"* entry into `actionsCommands`. With a ref plus a no-op `setReady(false)`, that entry stayed missing until some unrelated state change forced a render. Which is the worst case for it: when the TS worker is dead, reading the kit API surface by hand is exactly what you want.

The kit *viewer* itself was never affected — it needs `kitViewerLine`, and setting that is its own state update.

**Fix:** `useState` instead of `useRef`, so the fetch drives its own re-render.

I checked the one non-render consumer before switching. `handleGotoDefinition` reads it, and that's safe: the function is rebuilt every render, and `CodeMirrorEditor` reassigns `onGotoDefinitionRef.current = onGotoDefinition` on every render (line 104), so the handler can't close over a stale value.

One thing beyond the report, in the same lifecycle: the teardown now clears the declaration on a slug change. The ref version never did, so the previous game's declaration lingered until the next fetch overwrote it.

**Test:** `useCodeSurfaceLanguageService.test.tsx` renders the hook with a kit that resolves and `createCodeSurfaceLanguageService` returning `null`, then asserts the declaration is observable. Verified as a real guard — reverted to the ref and it fails.

Verification on the new head: type-check clean, lint 0 errors, 1,898 web tests pass, build clean.

---

gtanczyk:
Two more real findings from the latest round, fixed in `6a94a71`:

- **`EditorPanel.tsx`** — the `editorPanelHelpers.js` import sat mid-file, after the `DetailsPaneDef` type declarations, instead of with the top-of-file imports. Moved.
- **`useRemixGrip.ts`** — `dragRef` carried a `moved` flag, written on every qualifying pointer move but never read; `dragConsumedRef` is what the code actually branches on. Removed.

**Third finding, checked and left as-is:** the unbounded `pendingUpdatesRef` growth in `useCodeSurfaceLanguageService` when the language service never boots. Confirmed pre-existing — `git show e621149:apps/web/src/surfaces/studio/CodeSurface.tsx` has the identical ref-based queue with the identical unbounded `push`, and the queueing logic itself (`queueLanguageFileUpdate` / `applyLanguageFileUpdate`) lives in `codeSurfaceLanguageBind.ts`, a file this PR doesn't touch. Coalescing-by-path and capping is a real design decision, not a one-line fix like the kit-declaration issue was — I'd rather raise it as its own follow-up than fold it into an otherwise behaviour-preserving decomposition PR.

Verification on the new head: type-check clean, lint 0 errors, affected suites (RemixPanel, remix, EditorPanel) green.

---

PR #1142, opened 2026-09-03
fix(store,infra): index active submissions for notify sweep and reduce snapshot retention to 14d

gtanczyk:
## Summary

This PR addresses two production quota and storage growth issues:

1. **Firestore Read Quota Exhaustion (429K reads/day)**:
   - Cloud Scheduler triggers `POST /api/internal/notify-sweep` every 2 minutes (720 times/day).
   - `FirestoreSubmissionQueryStore.listActiveSubmissions` performed an unindexed full collection scan (`this.db.collection('submissions').get()`), charging for reading all ~600 documents in the database on every invocation ( \times \approx 595 \approx 428\,400$ reads/day).
   - Added `sweepActive?: boolean` to `SubmissionRecord` and indexed query `.where('sweepActive', '==', true)`.
   - Maintained `sweepActive` lifecycle on creation, transitions, notifications, abandonment, and account erasure.
   - When idle, the query reads 0 documents, cutting daily Firestore read volume from 429,000 to under 1,000.

2. **Unbounded GCS Snapshot Bucket Growth (40.61 GB across ~150 snapshots)**:
   - `gamedevpl-games-snapshots` retained every snapshot for 90 days.
   - Reduced snapshot lifecycle retention condition from 90 days to 14 days in `infra/setup-gcp.sh` and documentation (`docs/games-snapshot.md`). Since snapshots are rebuilt daily by cron at 04:23 UTC and on default branch merges, 14 days preserves 14+ rollback points while keeping the storage bounded at ~4–8 GB.

## Verification

- Green gate passed:
  - `npm run type-check` (0 errors)
  - `npm run lint` (0 errors)
  - `npm run test` (all suites passing, added `store-firestore-active.test.ts`)
  - `npm run build` (successful build)

### DIFF
--- a/apps/api/src/store/firestore.ts
+++ b/apps/api/src/store/firestore.ts
@@ -268,7 +268,7 @@ export class FirestoreStore implements Store {
           {
             ownerUid: DELETED_ACCOUNT_UID,
             ...(!record.publishedAt
-              ? { abandonedAt: record.abandonedAt ?? at, draftSharedAt: FieldValue.delete() }
+              ? { abandonedAt: record.abandonedAt ?? at, draftSharedAt: FieldValue.delete(), sweepActive: false }
               : {}),
           },
           { merge: true },
--- a/apps/api/src/store/in-memory.ts
+++ b/apps/api/src/store/in-memory.ts
@@ -150,7 +150,9 @@ export class InMemoryStore implements Store {
       this.submissions.set(submission.jobId, {
         ...submission,
         ownerUid: DELETED_ACCOUNT_UID,
-        ...(!submission.publishedAt ? { abandonedAt: submission.abandonedAt ?? at, draftSharedAt: undefined } : {}),
+        ...(!submission.publishedAt
+          ? { abandonedAt: submission.abandonedAt ?? at, draftSharedAt: undefined, sweepActive: false }
+          : {}),
       });
     }
 
--- a/apps/api/src/store/records/submission.ts
+++ b/apps/api/src/store/records/submission.ts
@@ -84,6 +84,8 @@ export interface SubmissionRecord {
    * a submission once it reaches a terminal, already-notified state.
    */
   lastNotifiedStatus?: SubmissionStatus;
+  // Indexed active flag; lets notify-sweep avoid full-collection scans.
+  sweepActive?: boolean;
   /**
    * The last status actually derived from GitHub, recorded on every derivation.
    *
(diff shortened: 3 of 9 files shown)


gtanczyk:
@codex

gtanczyk:
Closing in favour of #1143, which reached the same denormalized-flag idea and carries the
14-day lifecycle change from here (with a corrected rationale — see below). Thanks; the
storage half of this was right and I had it wrong.

**Taken from this PR: the 14-day retention.** #1143 originally left the window at 90 days,
reading the rule's own comment as saying that the window is the floor on how long the games
repo may go without a merge. That comment is stale: `publish-games.yml` bakes on a 04:23 UTC
cron regardless of merges, so the bound is *failed bakes*, not merges, and 14 days is safe.
#1143 makes the change and rewrites the comment and `docs/games-snapshot.md` to say that.

**Not taken: the sweep flag, because of a migration gap that would have taken the sweep down.**

`where('sweepActive', '==', true)` is blind to a document that does not carry the field, and
none of the 176 production submissions does. On deploy the sweep would have gone from 175
documents to **0**, and stayed there: the writes that set the flag (`recordJobTransition`,
`setSubmissionNotifiedStatus`) are the ones the sweep itself drives, so an invisible job never
becomes visible again. All 30 currently sweep-active jobs would have stopped being notified,
reconciled and auto-abandoned, with no error anywhere.

The PR body reads that failure as the success condition — *"When idle, the query reads 0
documents, cutting daily Firestore read volume from 429,000 to under 1,000"* — and
`store-firestore-active.test.ts` only writes a legacy document that is *supposed* to be
excluded (`state: 'published'`), so a fully green gate said nothing about it. #1143 runs the
backfill off the store's first query after deploy and records a marker doc, so there is no
operator step and no window where the sweep is blind.

**Second design point:** `sweepActive` here is exactly `isSweepActive`, so the flag sits on a
knife-edge — any write path that changes one of its three inputs without recomputing it drops
a job out of the sweep permanently. #1143 stores `isRoundOpen`, a deliberate *superset* of
both `isSweepActive` and the studio's in-flight test, with a property test pinning the
relation across the whole status/state matrix; callers re-apply their own predicate. A stale
flag then costs a wasted read rather than a lost build.

**On the numbers:** the collection is 175 documents, not ~600, and the bucket holds 400
snapshots, not ~150. The 429K/day figure is the Firebase console's partial day — the
`document/read_count` metric reads 0.68M–1.33M/day, and the 2-minute sweep is only ~126K of
it. The larger sources are `GET /api/submissions/mine`, polled once a minute per open tab by
the header badge (~600K/day), and `GET /api/review/status` (~340K/day). #1143 covers all three.

PR #1157, opened 2026-09-04
Dedupe formatSeconds, fixing the "1m 60s" bug on the health view

gtanczyk:
## What

`formatSeconds` existed twice. #1144 fixed the minute-rounding bug in the studio copy but left `GameHealthView`'s own copy behind, so the operator health table still renders `1m 60s` for durations like 119.6s:

```ts
// GameHealthView.tsx (before)
if (seconds < 60) return `${Math.round(seconds)}s`;
const minutes = Math.floor(seconds / 60);
const rest = Math.round(seconds % 60);   // 119.6 -> 1m + round(59.6) -> "1m 60s"
```

The fixed version rounds the total first, then splits.

## Change

- Move the fixed `formatSeconds` to `apps/web/src/relativeTime.ts`, which already owns `formatRelativeTime` and `formatDuration` — a shared root module both the studio surface and the operator view can import without one surface reaching into another.
- Point `StudioStatsSection`, `StudioDetailsPanel` and `GameHealthView` at it; drop both old copies.
- Move the `formatSeconds` tests out of `studioHealth.test.ts` into a new `relativeTime.test.ts` alongside the implementation. They already covered the `1m 60s` case (`formatSeconds(119.6) === '2m'`) — that assertion now guards the health view too.
- `module-size` baseline for `relativeTime.ts` resealed 41 → 50 for the moved function; `comment-prose` needed no reseal.

## Verification

- `npm run lint` — 0 errors (the 44 remaining warnings are pre-existing map-hygiene warnings in `apps/api`, untouched here)
- `npx vitest run` in `apps/web` — 236 files, 1931 tests, all passing

---


gtanczyk:
@codex

PR #1177, opened 2026-09-05
ci(cli): release from the changelog — categories decide semver, a PR is the cutoff

gtanczyk:
## Why

The first `gamedevpl` release needed four manual interventions to exist (#1161, #1162, two dispatches), and nothing decides when the next one happens or what number it gets. Since then `0.2.0` and `0.3.0` were also cut by hand, and the drift that caused is fixed here too (see the last section).

## The design

**`apps/cli/CHANGELOG.md` is the source of truth.** One line per change under `## Unreleased`, in one of four categories. The highest category present decides the version:

| Category | Bump |
| --- | --- |
| `### Breaking` | minor before 1.0, major after (1.0 itself is written by hand) |
| `### Added` | minor |
| `### Fixed` | patch |
| `### Internal` | **no release** |

**The cutoff is a commit a human merges.** `cli-release-pr.yml` watches master: whenever `Unreleased` has releasing entries, it runs the cut on branch `release/cli` (version bump in `package.json` and the installer's default, changelog section with date) and keeps one PR `release(cli): vX.Y.Z` current. Later merges fold into the same PR — that is batching. Merging it lands the version on master, and `cli-release.yml` publishes. Set repo variable `CLI_RELEASE_AUTOMERGE=true` and it merges itself once checks pass — the fully automatic mode, one flag away.

**The guard keeps the changelog honest.** A PR that changes the shipped CLI (`apps/cli/src/**`, `scripts/**`, `adapters.json`; tests excluded) without a changelog line fails CI naming the fix. Locally it is the last step of `npm run lint`; in CI it reads the PR's file list via `gh pr diff`, since a depth-1 checkout cannot diff against master (hence `pull-requests: read` on the CI token).

## Pieces

- `eslint-rules/cli-changelog-lib.mjs` — parse, validate, bump, cut, notes, version consistency (+ 17 tests)
- `eslint-rules/cli-changelog-check.mjs` — the guard
- `apps/cli/scripts/release.mjs` — `next` | `cut` | `notes`
- `.github/workflows/cli-release-pr.yml` — the proposer
- `.github/workflows/cli-release.yml` — also fires on a master push that changes `apps/cli/package.json`; skips versions that already have a release; notes from the changelog; no `--generate-notes`
- `apps/api/src/platform/cli-installers.ts` — falls back to `releases/latest` when the baked-in default is not published yet, so site deploy and release may land in either order
- `.claude/skills/cli-release/SKILL.md` — rules, commands, and the traps the first releases hit; row in `AGENTS.md`, section in the CLI README

## The drift this found, and now prevents

Answering "what version is the CLI?" turned up a real inconsistency: `cli-v0.2.0` and `cli-v0.3.0` were published by `workflow_dispatch` with an explicit version, so **neither touched the repo**. `package.json` and the installer's `CLI_VERSION` both still read `0.1.0`, which means `curl | bash` was serving `0.1.0` — two versions behind — while `0.3.0` was the latest release.

It also broke this PR's own seeding: deriving from `package.json` `0.1.0` would have proposed `0.2.0`, which already exists, so the publisher would have skipped it and the first automatic run would have done nothing.

Fixed by recording `0.2.0` and `0.3.0` in the changelog from what those tags actually contain, moving `package.json` and `CLI_VERSION` to `0.3.0`, and adding `versionConsistency()` — the version lives in three places and they must agree. Two doors enforce it: the lint guard prints all three on drift, and a test asserts it **against the real repository files**. Verified to go red when `package.json` is put back to `0.1.0`, so it guards something rather than passing by construction.

## Verified

- `npm run test:rules` — 75 tests (17 new) green
- `npm run lint` — green, ending in the guard
- installer tests (`cli-surface.test.ts`) and API type-check — green
- `release.mjs next` → `none` (everything released), `notes 0.2.0` / `notes 0.3.0` render their sections
- both workflow files parse

The earlier red CI on this branch was a flake in `apps/web/.../SubmissionStatusView.test.ts` — a file this PR does not touch, green on the base, green locally, and green on the one re-run.


gtanczyk:
**Lint, Test & Build** is red, and I do not believe the failure is this PR's. Standing down on it rather than pushing a speculative fix, with the evidence:

The three failures are all in `apps/web/src/surfaces/studio/SubmissionStatusView.test.ts`:

- `attaches an image pasted into the compact composer` — expected null not to be null
- `does not retry the same channel item after it fails to load` — spy called 2 times, expected 1
- `standalone: a chat-only reply on a published game refreshes the thread without a token` — spy called 4 times, expected 2

Why it reads as not mine:

1. **This PR touches nothing in `apps/web`.** The 13 changed files are `eslint-rules/*`, `apps/cli/*` (changelog, release script, README), `apps/api/src/platform/cli-installers.ts`, two workflows plus the CI guard step, `package.json` and `AGENTS.md`.
2. **The base is green.** `b2441ba2` — this PR's merge base — passed CI on master at 20:49, as did every master run since.
3. **The file passes locally on this branch**: `npx vitest run --dir apps/web src/surfaces/studio/SubmissionStatusView.test.ts` → 83/83.
4. The shapes are timing-sensitive — two spy-call counts and one element that had not rendered yet — in a run whose web suite reported 131s of tests against 156s of environment setup, i.e. a loaded runner.

No fix exists to port; there is nothing here I can port a fix *from*, since master is not failing this way. Taking my one re-run now to confirm. If it fails identically a second time, the assertion is genuinely fragile rather than unlucky, and I will say so and propose a fix for the test rather than widen this PR.

---

gtanczyk:
@codex

PR #1184, opened 2026-09-06
feat(cli): play games with live reload

gtanczyk:
`gamedevpl play` opens the current checkout with live reload, so a creator can play while a local coding agent edits. `/play` and explicit Polish/English requests such as “chcę zagrać w tę gierkę” open the active game without dispatching another build. Outside a matching checkout, `play <slug>` opens the remote game.

The local preview prepares the pinned Creator Kit, assembles source changes, and keeps the last successful game when compilation fails. Concurrent calls reuse one detached server, `--no-open` / `--json` return its URL, and `--stop` shuts it down. Interactive local delegation prints the preview URL before editing; unattended delegation does not start a preview. Startup and lock waits honor cancellation. Downloaded checkouts and agent briefs explain how to use play; MCP agents point to the existing round card or remote play link.

The browser renders only inside `sandbox="allow-scripts allow-pointer-lock"`. The server binds to loopback, uses an unguessable URL and rejects cross-origin requests; game HTML is served as inert text to the trusted shell. Session storage rejects symlinks and checks ownership/private permissions on POSIX, uses unique runtime paths, and publishes state atomically. Detached spawning hides the Windows console. The assembly buffer supports the full game-size budget, and the watcher includes existing `starters/` and `templates/` directories.

Reload can be paused and resumed, and inactive servers expire after 30 minutes. Reload restarts the game; it does not migrate arbitrary game state or publish previews. Anonymous `play_requested` events measure CLI attempts using the existing telemetry contract without paths or game identifiers; the operator panel has a matching label. Boolean flags preserve following positional arguments across commands.

Validation: npm install; root type-check, lint, test and build; CLI bundle. All 240 CLI tests pass, including concurrent startup, cancellation, unsafe session storage, unattended delegation, bundles above 26 MB, rebuild/recovery, last-good retention, sandbox/cross-origin checks, stop, and intent routing without a build. Earlier browser verification used a real game plus a temporary copy: automatic reload, pause/resume and 390px mobile viewport, with no console errors. Windows behavior has not been exercised on a Windows host.


gtanczyk:
@codex

PR #1187, opened 2026-09-06
feat(cli): interpret conversations with server-side assistant

gtanczyk:
Natural-language requests inside an active checkout previously went straight to the edit endpoint, so “uruchom airtime” was rejected when Airtime was published. Route every conversational message through the existing server-side CLI assistant, using conversation history and current game context. Slash commands remain direct shortcuts.

The existing intake model can reply or clarify, propose a new game, or request a validated `play`, `status`, or `edit` action. Play opens a matching local checkout with live reload or the remote game. Edit carries the complete agreed task so replies such as “yes, do that” retain conversational context, then uses the existing builder selection, handoff and revision controls. Starting another game clears the previous checkout and pending execution state.

The API verifies submission ownership and derives game state before building model context. Only game metadata and installed agent names reach the model, without tokens or local paths. API and CLI validate a shared, closed action contract; unknown tools, malformed arguments, multiple actions and unknown play targets fail closed. Clients without session context retain the existing intake protocol. No natural-language keyword matcher remains.

Validation: full repository install, type-check, lint, test, build and release packaging passed. Final regression runs passed: 240 CLI tests, 117 contract tests and 38 focused assistant/API tests, plus repeated type-check, lint and release packaging. Tests inject provider output to verify tool parsing, prompt context, ownership, clarification, action execution and outage handling; they do not measure live-model language accuracy.

Deployment order: deploy API first, then release CLI. This PR remains open for review and merge.

### DIFF
--- a/apps/api/src/creation/cli-chat-routes.test.ts
+++ b/apps/api/src/creation/cli-chat-routes.test.ts
@@ -1,3 +1,4 @@
+import { mintToken } from '../platform/submission-token.js';
 import type { FastifyInstance } from 'fastify';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { buildApp } from '../platform/app.js';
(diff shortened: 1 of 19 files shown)


gtanczyk:
@codex

PR #1193, opened 2026-09-06
feat(cli): show live TUI activity and progress

gtanczyk:
Submitting a prompt previously left an input-shaped box with a static ellipsis, making the CLI look idle. Replace it during work with an animated activity panel, elapsed time, and interrupt/exit guidance. Show meaningful foreground API stages and local preview/builder activity, then restore the editor when work completes. Background round polling does not overwrite foreground activity.

Keep the transcript, live status and controls within the terminal height. Window long option lists around the selected item, show numbered choices and concise keyboard hints, and keep long input tails visible. Support color and monochrome rendering without extra dependencies.

Validation: npm install, type-check, lint, test and build all passed. Tests render real Ink components at 40×12, 80×24 and 120×40, including long Polish text and 20 options; animation, prompt restoration and foreground request completion/failure are covered. Manually inspected actual Ink frames for busy, picker and prompt.

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -9,7 +9,7 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ### Added
 
-- The TUI shows animated activity, elapsed time and current steps while working, with persistent controls in compact terminals.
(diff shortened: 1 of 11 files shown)


gtanczyk:
@codex review

gtanczyk:
@codex

PR #1194, opened 2026-09-06
Pin a private, no-store default on every /api response

gtanczyk:
## What

A small `onSend` hook in `apps/api/src/platform/api-cache-policy.ts` sets `Cache-Control: private, no-store` on any `/api/*` response that has not set a `Cache-Control` of its own. Registered next to `registerClientAddress` in `app.ts`.

## Why

`www.gamedev.pl` now sits behind Firebase Hosting (cut over 2026-09-06). The edge caches a rewrite response only when its `Cache-Control` says so, and most `/api` routes sent no header at all — measured live: `/api/games` and `/api/auth/me` arrived with none. They stayed out of the shared cache only because of the edge's default for rewrites, which is not something this repo controls or tests. This makes the safe outcome an explicit decision of the app, so a future route that forgets the header cannot become publicly cacheable by accident, and a future change in edge defaults cannot either.

Routes that already choose a header keep it: game preview media keeps `public, max-age=86400, stale-while-revalidate=604800`, the discovery documents under `/.well-known/` are outside `/api` and are not touched.

## Reviewer notes

- The hook only acts when `reply.hasHeader(cache-control)` is false, so it cannot override an intentional policy.
- Refusals (401) get the same header. A cached 401 fragments per cookie just like a cached 200, so leaving them alone would defeat the point.
- The service worker never caches `/api` (see `isApiRequest` in `apps/web/public/sw.js`), and only the media route uses ETag/304, so `no-store` on the rest changes no client behaviour.
- `app.ts` was exactly at its module-size baseline (1205 lines). The import and the call cost two lines; the apex-redirect comment is shortened by three, meaning unchanged, and the file is now 1204.
- Verified: full API suite (264 files, 3859 tests) green, `tsc` clean, eslint clean, comment-prose and module-size ratchets pass.

This is FH-03 from the CDN fronting plan.

### DIFF
--- a/apps/api/src/platform/api-cache-policy.test.ts
+++ b/apps/api/src/platform/api-cache-policy.test.ts
@@ -1,0 +1,73 @@
+import type { FastifyInstance } from 'fastify';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { API_DEFAULT_CACHE_CONTROL } from './api-cache-policy.js';
+import { buildApp } from './app.js';
+import { mintSessionToken, SESSION_COOKIE_NAME } from './auth.js';
+import { InMemoryStore } from './store.js';
+
+const sessionSecret = 'dev-session-secret-change-me';
+const uid = 'g:cache-policy';
+
+function isPubliclyCacheable(header: string | undefined): boolean {
+  if (!header) return true;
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

PR #1195, opened 2026-09-06
fix(admin): let the console release every lane the spend brake can pull

gtanczyk:
## What

The spend brake (CC-22, #1152) can pause **editing, chat, search and gate runs**, but the admin console had pause/resume controls only for creation and tab completion. Verified live on 2026-09-06: a test notification set `searchPaused=true` in ~150 ms, and the only way back was a hand-written `fetch` from devtools.

- **New `IncidentLanesPanel`** in the Limits tab: one pause/resume button per brake-pullable lane, plus a summary of what is currently pulled.
- **API:** `effective.editingPaused` / `effective.chatPaused` were accepted in the patch but never reported back; they are now in the response like `searchPaused` / `gatePaused` already were.
- **`parseLanes` is case-insensitive.** A GCP label value cannot hold a capital letter, so `tabComplete` could not be expressed in a policy label; the script now writes `tabcomplete`.
- **`infra/setup-spend-brake.sh`** — five defects found when running it for real, each of which aborted the script or left the brake silently inert:
  1. the Monitoring notification service agent does not exist until asked for; the topic IAM binding failed on a missing member
  2. Monitoring filters need `displayName="…"` (double quotes)
  3. policy names lacked "abnormally high" → every policy `SKIP`ped
  4. `--set-notification-channels` **replaced** the Ops email channel with the brake; `--add-` keeps both
  5. the Pub/Sub service agent needs `roles/iam.serviceAccountTokenCreator` on the push caller, or every push fails with no trace

## Production state

Everything the script now does was already applied by hand on 2026-09-05 (topic, channel, SA, subscription, IAM, policy channels/labels). Re-running the script is idempotent; the only observable change on the next run is A24/A25 gaining `tabcomplete` in their `lanes` label.

## Tests

- `IncidentLanesPanel.test.tsx`: every lane has a control; resume calls through with `{searchPaused:false}`; absent flags render as open.
- `spend-brake.test.ts`: lowercased `creation_editing_chat_tabcomplete_search` maps to the five lanes.
- `admin.test.ts`: `effective` carries the two new booleans.

Ratchets: `IncidentLanesPanel` is a new file; the four grown files (type/field additions) are bumped.


gtanczyk:
@codex

PR #1201, opened 2026-09-07
fix(cli): guard Claude billing and show agent state

gtanczyk:
Claude tasks inherited API credentials despite a subscription login, while the TUI lost builder selections after failed requests and kept showing the old published round during editing.

- Remove inherited Anthropic API/cloud-provider environment settings from the Claude child and require a successful Claude.ai or setup-token OAuth / first-party auth status before spawning. Check asynchronously with cancellation in the task cwd with adapter settings flags, before local preparation or delegate telemetry; reuse the result for that launch and recheck on each new task. Preserve model selection, distinguish outdated CLI versions from missing login; unknown/API auth fails closed. This changes the Claude adapter's authentication contract and is documented under Breaking. Parent credentials and shell settings are untouched.
- Remember an explicitly selected local agent before requesting an improvement, including when that request fails.
- Show active local agent/tool and verification stages. Refresh the TUI token when the new workshop opens; terminal round watches sleep until a new round arrives rather than stopping permanently.
- Show local Claude session IDs/resume instructions, hide empty assistant event labels and duplicate Codex start events, render Antigravity events as readable progress, keep full verification diagnostics including stdout, and explain moderation refusals.
- Refuse empty Antigravity runs that report headless permission denial. A partial Claude tool refusal is a warning: exit-zero edits still reach verification and the normal delivery offer.
- Clarify the noninteractive brief: the agent edits, the CLI runs checkout checks after exit. Desktop bridging is not implemented; these remain local print-mode sessions.

Validation passed: npm install, full root type-check/lint/test/build, and focused auth/selection/status/verification tests. A read-only local Claude auth probe with API environment removed confirmed Claude.ai / Max; no model task was launched. Real checkout verification reproduces the separate Creator Kit bug: its no-plain-js checker rejects its own root setup.mjs. Creator Kit fix: https://github.com/gamedevpl/www.gamedev.pl-games/pull/1290. This PR preserves and displays that error; it does not weaken the game's checks or patch the user's checkout. Production moderation logs record only category other, so the cause of that individual refusal remains unknown.

Instrumentation: creator funnel/return (questions 4–5), through existing cli_step and round records. No prompts, keys or paths added to telemetry. Exact token accounting and Desktop session bridging remain outside this fix. Subscription extra-usage limits remain managed by Claude.

### DIFF
--- a/.claude/skills/verify-agent-work/SKILL.md
+++ b/.claude/skills/verify-agent-work/SKILL.md
@@ -310,6 +310,11 @@ Two concrete instances of that (observed 2026-07-23):
   well as the parser, and exercise each action on its real lifecycle state. Published
   edits need `/improve` and its new token, not the closed round's `/turn`.
 
+- **Exercise auth variants and partial agent failures.** CLI #1201 initially accepted only
+  `claude.ai`, rejecting subscription `oauth_token` logins, and treated any tool denial as
+  failure of the entire edit. Test each supported login source and an exit-zero run with
+  completed edits plus one denied tool. An agent's denial list is not a task verdict.
+
 ## Read the diff against the spec
 
 A passing test suite doesn't catch scope creep, subtle regressions, or malice.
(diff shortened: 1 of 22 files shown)


gtanczyk:
@codex review

gtanczyk:
@codex review

PR #1209, opened 2026-09-07
fix(auth): guard cookie session writes from CSRF

gtanczyk:
## Summary

Cookie-authenticated writes now reject untrusted request origins before reaching handlers. Personal access token creation additionally requires JSON, preventing the shared OAuth form parser from accepting browser form submissions at that endpoint.

The guard covers session and PAT-derived cookies, with Origin validation, Referer fallback, and Fetch Metadata checks when neither is present. Same-origin, canonical, and configured frontend origins remain supported, as do Bearer-only clients and headerless API clients. The Vite proxy preserves Host for local development.

## Validation

- `npm install`, type-check, lint, full test suite, and build passed.
- Full suite: 6,433 passed; 4 skipped.
- 27 new regression tests; 22 fail against the previous implementation.
- Local browser CSRF attempt returns 403 and creates no token.
- Legitimate JSON token creation and revocation through the Vite proxy succeed.

Changes have been validated locally; production deployment is not part of this PR creation.


gtanczyk:
@codex

PR #1212, opened 2026-09-07
Retire the gamedev_session fallback cookie

gtanczyk:
## What

Removes the `gamedev_session` read-fallback that carried accounts across the rename to `__session`.

- `readSessionCookie` returns the token or `undefined`; the `legacy` flag and the forced re-mint on a legacy cookie are gone.
- `clearSessionCookies` clears one name. `retireLegacyCookie` and `LEGACY_SESSION_COOKIE_NAME` are deleted.
- Tests that exercised the fallback are rewritten to exercise the same invariants with the current name — a handler that wrote the session cookie wins over renewal; logout never re-mints; a sign-in as someone else never leaves the previous identity last — plus one new test pinning that a browser carrying only the old cookie is anonymous and gets nothing minted.

## Why now

The fallback existed for a sign-out storm that would otherwise have happened at the flag day. `www` has been behind Firebase Hosting since 2026-09-06 and every active session has been re-minted under `__session` since the rename shipped on 2026-09-02, so there is nothing left for it to migrate. Left in, it is a second authentication path nobody exercises.

## Reviewer notes

- `docs/deployment.md` is corrected while being edited: Hosting does forward cookies other than `__session` to a Cloud Run rewrite (the token-login CSRF cookie round-trips through it, measured 2026-09-06). `__session` stays because it is the one *documented* guarantee, not because the rest are dropped.
- No web/client change: nothing outside `apps/api` referenced the old name.
- Verified: full API suite green (three unrelated timeouts under full-suite load pass in isolation — `self-build`, `agent-channel`, `catalog-search`), `tsc` clean, `npm run lint` clean including the prose and size ratchets.

### DIFF
--- a/apps/api/src/platform/access-token-routes.test.ts
+++ b/apps/api/src/platform/access-token-routes.test.ts
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, it } from 'vitest';
 import { generateAccessToken } from './access-token.js';
 import { DEFAULT_EXPIRY_DAYS, MAX_TOKENS_PER_UID, mintAccessTokenFor } from './access-token-service.js';
 import { buildApp } from './app.js';
-import { LEGACY_SESSION_COOKIE_NAME, mintSessionToken, SESSION_COOKIE_NAME } from './auth.js';
+import { mintSessionToken, SESSION_COOKIE_NAME } from './auth.js';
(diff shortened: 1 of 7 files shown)


gtanczyk:
@codex

PR #1213, opened 2026-09-07
feat(cli): offer safe Creator Kit updates

gtanczyk:
Updating gamedevpl leaves an existing checkout pinned to its old Creator Kit. Creators currently have to download another checkout and copy its lock manually. This adds an in-CLI update flow that preserves their game edits.

- Check the published Kit when opening a checkout in the TUI and offer Update now / Later. `/kit` repeats the offer; `gamedevpl kit update` explicitly installs from the shell. Local `play` reports updates without interactive prompts and still works when discovery is offline.
- Fetch an authenticated, owner-scoped pin through the workspace route's `kitOnly` response, without retrieving game sources. Pins remain private and non-cacheable.
- Prepare the Kit and pinned npm dependencies under `.gamedev/` first, verify its release identity, then swap only manifest-owned paths, dependencies and metadata. Preserve games, Git state and custom workflows. Refuse symlinks and new-path collisions; back up the old installation and journal promotion for rollback/recovery. Serialize updates in a checkout and refuse ordinary preparation while a swap is pending.
- Show installation activity in the TUI, support cancellation, stop stale preview processes, and distinguish completing setup from a newer release. Successful installation does not claim game validation or publication.
- Add update availability/start/completion/failure telemetry with matching dashboard labels. No pins, signed URLs, paths or sources enter telemetry.

Validation: full root install/type-check/lint/test/build passed; client staging, consent, cancellation, failure, collision, rollback and recovery tests; API ownership/source-read tests; real Ink picker/activity rendering tests. The CLI bundle and its rendered help were also checked.

Rollout: deploy the API support before publishing the CLI release. Existing checkouts use their version-tolerant setup.mjs; game sources are never pulled or submitted by an update. Interrupted promotions recover when the creator accepts `/kit` again or runs `gamedevpl kit update`. Local changes inside Kit-owned tools are replaced, as with setup.mjs; game edits remain under games/<slug>.


gtanczyk:
@codex

PR #1222, opened 2026-09-07
feat(telemetry): read the CLI pilot funnel, and give it its missing rungs

gtanczyk:
CL-39 asks the operator page one question: **does anyone finish a game this way**. The gamedevpl CLI block could not answer it, for two reasons that only show up together.

**The read side dropped every dimension.** `cli_step` has carried `adapter`, `stage`, `channel` and `os` since CL-01 — closed enums, validated at intake, stored. `summarizeCliFunnel` returned `{step, visits}` and threw all four away, so "which agent do creators actually run", "where does the ladder break" and "how did the binary arrive" were captured for nothing.

**Four of the ten rungs had no writer.** `installed`, `authorized`, `first_turn` and `published` were never recorded by anything. The block's "Of installed" column divided by a number nothing produced, and the one rung the public half of Wave F is gated on — `published` — was structurally always zero.

## The panel

```
gamedevpl CLI
Does anyone finish a game this way? 1 session of 4 watched a game publish; 2 delivered sources.
Every gamedevpl run is its own session, so these are session counts, not one creator
followed from install to publish.

Step                          Sessions   Of sessions
installed the CLI                    4          100%
signed in                            3           75%
…
Local agent   Offered   Ran   Taken up
claude              3     1        33%

Ladder broke at: check:static 1
Arrived by: gamedevpl update 2 · unknown channel 2
Ran on: macOS 3 · Linux 1
```

When nothing delivered, it says nobody has published this way yet rather than rounding delivery up into a finish. When the window is empty it keeps the existing refusal to invent a cohort.

## The rungs, written where the facts already are

- **`authorized`** after an interactive login; **`first_turn`** on the first thing a creator types at the REPL.
- **`published` only on a transition the session watched happen.** Opening a checkout of an already-live game must not add to the number the launch decision reads, so `isPublishTransition` requires a non-published status first.
- **`installed` once per install**, marked locally under `~/.config/gamedevpl`, and only from a terminal. A CI container is a fresh HOME per job; counting those would bury the pilot cohort under our own automation.

## Privacy posture, unchanged

The install marker is a local flag, never sent and never an identifier — every report still travels under a fresh per-process visit id, and the file holds one version string. Nothing new leaves the machine except the closed enums CL-01 already declared: no source text, no prompts, no paths, no titles. The installers still do not phone home, so `curl` vs `ps1` stays unmeasured and reads `unknown channel` unless `gamedevpl update` wrote it — a named gap, not a silent one.

A CLI run is its own session (the process mints a visit id and drops it on exit), so the rungs are session counts and there is no install-to-publish journey in the data. The block says that in words instead of letting the percentages imply one; a durable machine id would be the CLI's cookie and is not on the table.

## Verified

- `summarizeCliPilot`: dedupes per session, keeps closed values at zero, names `unknown` only when a client predates a dimension, and stays empty for visitors who never ran the CLI (4 new tests).
- `CliFunnelBlock`: the verdict line, the delivered-but-not-published case, take-up per agent, and a client that predates `cliPilot` (6 new tests in a file of its own).
- `install-mark`: one report per version, silence off a terminal without spending the report, the channel an update left behind, a corrupt mark, and nothing machine-identifying on disk (5 new tests).
- `isPublishTransition` and the round-watch `onStatus` hook (2 new tests).
- `@gamedevpl/cli` 274 green · API telemetry + web funnel 130 green · CLI, API and web type-check green · `npm run lint` green.

Instrumentation contract, per the skill: this closes question 4 for the CLI path. The `product-instrumentation` skill's CL-39 gap is rewritten as closed, with the three rules that came out of it.

---


gtanczyk:
@codex

gtanczyk:
All three Codex findings reproduced and fixed in `09250bef` (on top of the rebase onto master, so the diff is against `0.7.0`'s kit rungs too).

- **Unwritable marker (P2)** — persisting the mark now gates the report. An unwritable home costs one install rather than reporting a new one on every run; the case has a test that puts a file where `$HOME` should be, so the write fails for root too.
- **Unvalidated persisted channel (P2)** — `readMark` validates both fields against the shared vocabulary: an unknown `channel` is dropped rather than sent, so the closed enum cannot refuse the event and silently lose the install, and no free string leaves the machine.
- **`status --watch` (P2)** — `status` now carries telemetry and records `published` on the same transition rule as the REPL watcher, so a publish a creator watched from the explicit watcher reaches the number the panel calls the verdict. An already-published first poll still records nothing.

`@gamedevpl/cli` 291 green · CLI type-check green · `npm run lint` green.

---

PR #1226, opened 2026-09-07
feat(infra): declare the billing budgets in a script, not just in the console

gtanczyk:
## Why

The billing budgets that drive the spend brake were created by hand and existed nowhere in the repo. Here that gap is worse than the usual drift: a budget's **display name is load-bearing**. `lanes=` in the name is how a per-service budget tells the brake what to pause (`spend-brake.ts`), and the name reaches the admin console as `updatedBy` on a paused lane. A rename in the console silently changes what an overrun stops.

## What

`infra/setup-budgets.sh` declares the set, idempotently (matched by display name, then created or updated):

| Budget | Scope | Notifies brake |
|---|---|---|
| `Total monthly` | whole account | yes — graded by the ladder |
| `Vertex + Gemini lanes=seeding_managed` | Vertex AI + Gemini API | yes — those two lanes |
| `Cloud Build lanes=gate` | Cloud Build | yes — gate runs |
| `Cloud Run` | Cloud Run | **no** |
| `Firebase Hosting egress` | Hosting | **no** |

Cloud Run and Hosting deliberately stay off the topic: pausing creation lanes cannot reduce the cost of serving traffic that is already arriving, and a budget with no `lanes=` would be graded by the ladder — stopping the agent because egress rose is the wrong lever.

**Amounts and the billing account id are not in this repo.** They are read from the environment; the values live in the ops repo alongside the rest of the money side. What the business will spend per service is not a fact a public repo should carry. Which budget pauses which lane is, because that is the contract the brake reads.

Owner-run and separate from `setup-spend-brake.sh` on purpose: that script owns the project half (topic, subscription, service accounts, alert policies), while a budget belongs to the billing account, and the deploy identity holds project roles only.

## Details worth knowing

- `create` and `update` take different flags — `--threshold-rule` vs `--add-threshold-rule`, and update rejects a currency suffix on the amount. The script builds each set separately; `--clear-threshold-rules` rides in the same update call so a budget is never briefly left with no thresholds.
- A budget filter names the project by **number**; the id is rejected.
- There is still no BigQuery billing export, so per-service spend cannot be read back and checked against what was budgeted.

Verified against the live budgets: `bash -n` clean, indirect expansion works under bash 3.2, and the declared filters, thresholds and topics match what the account returns today. Not re-run against GCP from here — that is the owner's to run.


gtanczyk:
Moved to the private ops repo instead — the amounts are a statement about what the business will spend per service, and this repo is public. The script now lives there alongside the rest of the money side; the brake that reads the budgets stays here, and `setup-spend-brake.sh` still documents the wiring.

PR #1231, opened 2026-09-08
feat(cli): add completion and agent failure recovery

gtanczyk:
Adds command suggestions and actionable recovery when an agent fails. Typing `/pu` offers `/pull`; a Codex capacity failure now explains that the agent stopped and offers reconnection to the existing round instead of `submit`.

Typing `/pu` in the interactive prompt now offers `/pull` with its description. `/` shows a scrollable list, typing filters it, arrows select, and Tab fills the command without executing it. Enter fills a partial command first; a complete command keeps the existing send behavior. Esc hides suggestions while preserving the draft, and history remains available outside the menu.

Descriptions reuse `/help`; session commands `/retry`, `/quit`, and `/exit` are included. The menu reserves terminal rows so build status and controls remain visible. Completion is local and sends no model requests.

Validation: 14 focused tests passed, including actual Ink keyboard events, history, completion without execution, and 40×12 / 80×24 / 120×40 terminal layouts. Manually exercised `/pu`, Tab and Enter in a PTY with build status visible. Autocomplete validation passed the full repository gate. After adding agent failure recovery, install/type-check/lint and all 318 CLI tests passed. The full run hit timing failures in 10 files in other workspaces; all affected files passed isolated reruns. The final repository build passed.

Instrumentation: this improves command discovery within the creator funnel (question 4). Existing `cli_step` events grouped by step and visit ID continue to measure executed actions. Suggestion use itself is not measured; no draft text or keystroke telemetry is added.

Agent recovery handles both streamed and buffered MCP errors and streamed local-agent errors. A failed run never enters local verification/delivery. `/connect <slug> --agent <name>` reconnects to the existing MCP round; regression tests verify repeated attempts use GET requests without creating another round. Capacity diagnostics retain only a boolean, not a transcript. Unknown failures retain their exit code and the same recovery guidance. Cancellation keeps its existing behavior.

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -7,6 +7,10 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ## Unreleased
 
+### Added
+
+- Interactive slash-command suggestions filter as you type; arrows select and Tab completes without running the command.
+
 ## 0.9.0 — 2026-09-08
 
 ### Added
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -9,7 +9,7 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ### Added
 
-- Interactive slash-command suggestions filter as you type; arrows select and Tab completes without running the command.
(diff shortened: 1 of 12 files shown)


gtanczyk:
@codex

PR #1234, opened 2026-09-08
feat(studio): guide creators through agent setup

gtanczyk:
Creators currently see MCP configuration, CLI installation, authentication alternatives and the kickoff prompt on the same screen. Replace initial setup with a guided choice: terminal, an existing agent, or the gamedev.pl builder. Show only the selected tool's setup, then the kickoff step.

- Include Cursor/VS Code install links, manual authentication fallback, and separate macOS/Linux and Windows CLI instructions. Credentials stay masked on screen and are copied only on request.
- Keep the returning-agent resume flow, platform handoff, live connection progress, and Studio navigation. Continuing setup does not claim that an agent has connected.
- Add Polish/English copy, responsive choices, keyboard focus, and regression coverage for progressive disclosure, clipboard credentials, and existing wizard transitions.

Validation: npm install, type-check and lint passed. All 1,971 web tests, 319 CLI tests, contract/zone/world suites and 76 rule tests passed. All 3,975 API tests passed on a complete rerun with two workers (the initial default-concurrency run hit nine 5-second timeouts). Production build passed. Browser checks used the actual wizard with fixture responses on desktop and mobile, Polish and English, including a 390×420 viewport and both install/update overlays; footer actions remain reachable and there is no horizontal overflow.

Instrumentation: affects creator funnel question 4. Existing `connect_copied` (install/kickoff), `connect_deeplink`, and wizard/Studio steps remain. Query visit funnel steps by visit and builder, together with creator build/publication records, to measure progression. Route-choice abandonment is not separately instrumented; that remains an explicit gap. No credentials, game identifiers or prompts are added to visit telemetry.

Follow-up: restore URL/browser OAuth as the default for Claude Code and Codex; manual bearer configuration remains a fallback. Add Muse Code (Meta) to the wizard and CLI discovery for local checkout delegation (`muse exec --trust-workspace`). Preserve Muse authentication, approval and sandbox defaults; no MCP support is advertised for this adapter. Use plain output instead of exposing Muse's raw event envelopes. The adapter help and echo mode were checked against installed Muse Code 1.0.3-R2198.1 without a paid model call; a real paid editing run was not performed.

The `muse` telemetry adapter value flows through the shared vocabulary into the existing pilot table. Full type-check, lint, workspace tests, rule tests and production build pass after this follow-up.

### DIFF
--- a/.claude/skills/product-instrumentation/SKILL.md
+++ b/.claude/skills/product-instrumentation/SKILL.md
@@ -350,3 +350,6 @@ CLI connection onboarding records `connect_opened` when the interactive choices 
 and `checkout_opened` after downloading or reusing a checkout. Neither event carries
 slugs, paths, credentials or prompts. Existing adapter offer/use events still describe
 agent launches; opening a session does not count as a build.
+
+Muse Code uses the `muse` adapter dimension for local checkout delegation.
+`CLI_ADAPTERS` drives the pilot aggregation and its adapter table directly.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -7,6 +7,10 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ## Unreleased
 
+### Added
+
+- Detect Muse Code and delegate local checkout tasks with `--agent muse` (#1234)
+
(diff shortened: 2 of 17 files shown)


gtanczyk:
@codex

PR #1249, opened 2026-09-08
Add A31 alert for Firestore daily read total detection

gtanczyk:
## Summary
This PR adds a new monitoring alert (A31) to detect sustained Firestore read cost regressions that the existing rate-based alert (A30) cannot catch. The new alert monitors the total document reads over a 24-hour period rather than instantaneous rate, addressing a blind spot that allowed the 2026-09 incident (~800K reads/day) to go unnoticed for weeks.

## Key Changes
- **Added A31 alert policy** in `infra/setup-monitoring.sh`:
  - Monitors daily Firestore document read totals using `ALIGN_DELTA` over 86400s with `REDUCE_SUM`
  - Fires when daily reads exceed 600K (roughly 1.6× the measured baseline of ~364K/day)
  - Uses 0s duration to detect sustained leaks rather than spikes
  - Includes comprehensive documentation explaining the rationale and triage steps
  
- **Updated documentation** in `docs/firestore-read-cost.md`:
  - Explains why rate-based monitoring alone is insufficient
  - Documents the complementary nature of A31 to A30
  - Provides guidance on re-calibrating both thresholds together after the badge-polling fix is deployed
  - Clarifies that the 600K threshold is calibrated against the pre-fix baseline and should be re-derived post-fix

## Implementation Details
The A31 alert is intentionally designed as a slow signal: it requires a leak to persist for most of a day before firing, making it effective at catching permanent regressions (e.g., reading a collection per-request instead of per-window) while remaining blind to temporary spikes. This complements A30's fast detection of rate anomalies. The threshold of 600K reads/day is well above normal operations (~364K/day) but well below the incident threshold (~800K/day), providing a clear signal without false positives.

https://claude.ai/code/session_016bqoSBXDwww3EzU7g3eXG3

### DIFF
--- a/docs/firestore-read-cost.md
+++ b/docs/firestore-read-cost.md
@@ -64,3 +64,15 @@ twin of A29 for writes. It is a ceiling over the floor the caches leave, calibra
 sit well above the intended state and well below an incident. **Recalibrate it whenever
 this table changes** — an alert calibrated against an old floor is an alert that no longer
 fires for a regression the size of the one it was written for.
+
+Rate alone is not enough. A30 evaluates a rate in two windows (ten minutes for a spike,
+three hours for drift), and both are blind to the shape that actually produced the 2026-09
+bill: a regression that adds a couple of reads a second, never peaks, and never stops. So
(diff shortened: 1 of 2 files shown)


gtanczyk:
@codex

PR #1244, opened 2026-09-08
Cache review queue and notifications to reduce Firestore reads

gtanczyk:
## Summary

Implements caching for frequently-polled endpoints to reduce Firestore read costs. The reviewer badge (`/api/review/status`) and notification bell (`/api/notifications`) were each issuing collection queries on every poll, causing read rates to scale with the number of open browser tabs rather than data change frequency.

## Key Changes

- **Review Queue Cache** (`review-queue-cache.ts`): New module that caches review pools, open sweeps, assessed slugs, and targeted re-review items with a 10-minute window (wider than the 2-minute badge poll interval). Caches are keyed per reviewer to prevent serving one user's queue to another.

- **Notification Cache** (`notification-cache.ts`): New module that caches the notification list per user with a 5-minute window (wider than the 1-minute bell poll). Uses a WeakMap keyed by store instance to support multiple stores.

- **Cache Invalidation**: Both caches are explicitly dropped when writes occur that would make cached answers stale:
  - Review queue cache drops on verdict submission, sweep status changes, and re-review requests
  - Notification cache drops on notification creation, read, or clear operations

- **Review Routes Refactor** (`review.ts`): Extracted review queue logic into the new cache module and integrated it into the routes. Removed inline caching logic that had a 60-second window (too narrow for the 2-minute poll).

- **Notification Routes Integration** (`notifications.ts`): Integrated the notification cache into the routes, dropping the window on write operations.

- **Documentation** (`firestore-read-cost.md`): Added comprehensive guide explaining the caching strategy, the rule that windows must be wider than polls, and a table of all cached surfaces with their poll intervals and window sizes.

- **Tests**: Added comprehensive test suites for both caches verifying:
  - Collections are read once per window, not per poll
  - Windows are keyed per user/reviewer to prevent data leakage
  - Writes properly invalidate cached data
  - Window expiration triggers re-reads

## Implementation Details

- Uses `rememberBounded()` utility to implement LRU-capped caches (200 reviewers, 500 notification users) to prevent unbounded memory growth
- Cache entries store expiration timestamps to enable time-based invalidation
- Both caches support a `fresh` option to force immediate re-reads when needed
- Maintains backward compatibility with existing route signatures while improving performance

https://claude.ai/code/session_01Uvg3SxBUgb6Qdew1JgPue7

PR #1243, opened 2026-09-08
fix: make agent setup and local editing reliable

gtanczyk:
Creators could select an unavailable managed builder, lose their setup steps after refresh, and mistake local agent work for delivery. The connection wizard and Studio panel now share a guided setup that retains CLI commands, identifies the own-agent route, respects managed availability, and remembers progress per round.

CLI delegation now provides selectable terminal history and preview links, compact activity output, full local logs, Muse event rendering, and saved per-agent model/effort choices through `/model`. Failed validation returns diagnostics to the selected agent for up to two repair attempts; delivery is offered only after checks pass. Claude session notices and repeated completion output are suppressed.

Local agent activity is reported separately to Studio: preparation, editing, permission prompts, native terminal control, verification, completion, failure, and lost contact. Updates are owner-only, reject superseded runs and closed round generations, and never imply that local files have been delivered. The payload contains only agent, run identifier, phase and timestamps; no prompts, paths or credentials.

When Antigravity cannot ask for permissions headlessly, the TUI offers native terminal takeover, resumes the conversation and verifies after the user exits. On macOS/Linux, the native transcript is retained in `/logs`; Windows retains native terminal scrollback. Permission decisions remain with the creator.

Validation: full repository tests, type-check, lint and production build; regression coverage for guide persistence/availability, output parsing, model settings, repairs, status ownership/generation and interrupted reporting. Browser inspection confirmed active, offline and ready local status. A real Claude subscription smoke run completed; a native PTY round-trip verified stdin and transcript capture. Muse protocol output was exercised without a paid Meta model call. Live execution of every supported provider was not tested.

This PR does not publish a CLI release or deploy the website.

PR #1254, opened 2026-09-09
perf(reads): cut the lookup floor under the per-window caches

gtanczyk:
Follow-up to the read-cost work (#1237, #1244, #1240, A30/A31). Those removed **collection scans** — a query on a polled path, once per request. This removes the floor underneath them, which no per-window cache touches.

`getSessionUser` reads `users/{uid}` on **every authenticated request** (`apps/api/src/platform/auth.ts:371` → `FirestoreIdentityStore.getUser`, an uncached `doc.get()`). A signed-in tab therefore pays one document lookup per poll, per surface, before the route does any work of its own.

## 1. The poll that cost the most and showed the least

`LocalActivityStatus` (from #1243) polled `/api/me/studio/local-activity/:token` **every 10 seconds, unconditionally while mounted** — it sits at the top of the connect guide, and renders `null` when there is no activity, so the cost was invisible. Each poll is 2 reads (session user + submission): **~17K reads/day for one forgotten tab, about a third of the 50K/day free tier.**

Now: fast (10s) while a task is running, widened to 60s after six empty polls or a finished task, and **nothing at all while the tab is hidden** (immediate poll when it comes back). It widens rather than stops, because the creator's *next* local task appears in that same guide — stopping on `ready` would have been a silent regression.

## 2. A short window on the session user

`FirestoreIdentityStore` now holds the user for **30 seconds** and drops the entry on every write through the store (profile edit, upsert, handle claim/release, deletion request and cancellation, email/digest flags). Copies go out, so a caller mutating the result cannot poison the next request's answer; a missing user is never cached, so a fresh sign-up is not hidden for a window.

**The cost, stated plainly:** `tier` and the block live on that document. A block applied through the API is immediate on the instance that applied it; a block applied **outside** the app, or served by one of the other instances (`--max-instances 4`), is enforced **up to 30 seconds late**. That is why the window is 30s and not the minutes the content caches use. If that trade is not acceptable, the honest alternative is widening the poll cadences instead, and this half can be dropped — the poll fix stands on its own.

## 3. Measuring, before the 2026-09-15 recalibration

`infra/read-cost-report.sh` splits `document/read_count` by `metric.label.type` for a chosen window. The split decides where the next fix belongs: **QUERY** means a scan came back, **LOOKUP** means a request path fans out per document (the session read, reads by id). A30 and A31 are both calibrated against a floor that the two changes above move, and the doc already says to re-derive them together from a full working week — this is what that measurement looks like. I could not run it from here (no `gcloud` in this environment), so the numbers still have to come from a real run.

`docs/firestore-read-cost.md` gains the floor as its own section, the connect guide as a row in the surface table, and the rule this left behind:

> A poll's cost is its cadence times its cheapest possible answer, and "nothing to show" is the answer it will give most of the time.

## Verified

- 6 new store tests: the window serves a repeat read, expires after 30s, is dropped by writes (upsert, profile, deletion request and cancellation), hands out copies, and never caches a missing user.
- 5 new web tests: the cadence table (`nextLocalActivityDelay`), and a hidden tab that asks nothing until it is shown.
- `apps/api` 4032 tests green (the 40 logged errors are moderation fail-closed paths asserting themselves, not failures) · store + auth + studio 158 green · API and web type-check green · `npm run lint` green.
- Note, unrelated to this diff: `CodeSurfaceAgentMode` (2) and `EditorSurface` (2) fail identically on a clean master checkout in this environment — the same four #1202 recorded. Not touched here.

---

### DIFF
--- a/apps/api/src/store/firestore.ts
+++ b/apps/api/src/store/firestore.ts
@@ -292,6 +292,8 @@ export class FirestoreStore extends SubmissionFacade implements Store {
       for (const write of writes.slice(start, start + BATCH_SIZE)) write(batch);
       await batch.commit();
     }
+    // The read above seeded the session window; an erased account must not survive it.
+    this.identityStore.forgetUser(uid);
 
     return { publishedSlugs, unpublishedSlugs };
   }
--- a/apps/api/src/store/slices/identity-user-cache.test.ts
+++ b/apps/api/src/store/slices/identity-user-cache.test.ts
@@ -1,0 +1,100 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { fakeFirestore } from '../fake-firestore.js';
+import { FirestoreIdentityStore } from './identity.js';
+
+// The read floor: tier and blocks live here, window stays short.
+function withUser(tier = 'standard') {
+  const { db } = fakeFirestore();
+  const store = new FirestoreIdentityStore(db);
+  return {
+    db,
+    store,
+    seed: async (fields: Record<string, unknown> = {}) =>
+      db
+        .collection('users')
+        .doc('u1')
+        .set({ uid: 'u1', createdAt: '2026-01-01T00:00:00Z', tier, ...fields }),
+    // Writing behind the store makes a stale window visible.
+    behindTheStore: async (fields: Record<string, unknown>) =>
+      db
+        .collection('users')
+        .doc('u1')
+        .set({ uid: 'u1', ...fields }, { merge: true }),
+  };
+}
+
+describe('FirestoreIdentityStore user cache', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date('2026-09-09T10:00:00Z'));
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('serves a repeat read from the window instead of Firestore', async () => {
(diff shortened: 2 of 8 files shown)


gtanczyk:
@codex

gtanczyk:
All three reproduced and fixed in `a594875d`. Each has a test that fails without its fix — I checked both new ones by reverting the fix and watching them go red.

- **Erased account survived the window (P2)** — right, and it was the worst of the three: `deleteAccountIdentity` reads the user *before* deleting it, so it seeded the very window it then had to outlive. `forgetUser` is now public and the facade calls it after the deletion batches commit.
- **Eviction inside the transaction callback (P2)** — right, in all three places (`scheduleAccountDeletion`, `cancelAccountDeletion`, `upsertUser`); they now evict after the awaited `runTransaction`. That still leaves the smaller cache-aside race the finding implies — a read that fetched before the drop and stores after it — so a generation counter closes it: a read that started before any drop returns its value but caches nothing.
- **`date -d "-1d"` (P2)** — confirmed on this box (GNU coreutils 9.4: `date: invalid date '-1d'`), and the BSD fallback fails there too, so the script produced an empty `--interval-start-time` and could not run on Linux at all. The shorthand is parsed (`1d` / `7d` / `12h`) and each `date` gets the form it understands; anything else is refused with a usage line instead of an empty window.

`apps/api` 4034 tests green · API type-check green · `npm run lint` green.

---

PR #1256, opened 2026-09-09
fix(cli): recover delivery from an open agent session

gtanczyk:
A local checkout can pass validation while delivery remains locked by an earlier MCP session that never called `end`. The CLI now checks ownership before offering delivery and asks explicitly to disconnect an own-agent session and deliver the full local checkout. Noninteractive retries use `gamedevpl submit --takeover`; `--force` never takes over a session.

The owner-only API atomically checks the confirmed job and generation, advances the generation to revoke old session keys, and marks the session ended. It preserves round budgets and existing preview/publication records. The old staging remains isolated in its original generation. Managed agents, pending handoffs, and rounds already delivering or publishing cannot be taken over. The local agent process is not killed.

Validation: type-check, lint and build pass. All 366 CLI tests pass on rerun; the final focused API and CLI checks pass (8 + 19 tests). The full workspace run had two failures outside this diff (stage-hints typecheck advisory and an agent-process timeout); both pass when rerun. Web tests pass (1,993). Regression tests cover creator OAuth access, ownership, explicit confirmation, stale generations, old-token rejection, round/budget preservation, full local snapshots, and unattended/declined prompts.

### DIFF
--- a/.claude/skills/byoca-mcp/SKILL.md
+++ b/.claude/skills/byoca-mcp/SKILL.md
@@ -1078,3 +1078,7 @@ the Studio live-preview frame, which additionally takes no pointer input and no 
 
 If you change the MCP tool set, submit warnings, stall vocabulary, or handoff rules
 and this file is wrong or missing the new behaviour, update it in the same session.
+
+### Creator takeover for local delivery
+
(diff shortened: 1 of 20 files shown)


gtanczyk:
@codex

gtanczyk:
@codex

PR #1260, opened 2026-09-11
fix(cli): announce available updates at startup

gtanczyk:
Starting an interactive session checks Creator Kit updates but never checks the CLI release, so creators can keep using an old CLI without knowing an update exists.

Check for a newer stable CLI release in the background when the TUI starts and show the installed/new version with `/update` and restart instructions. The check has a three-second timeout, never blocks input, silently tolerates network/rate-limit failures, and suppresses late results after exit. Draft, prerelease and malformed version tags are excluded.

Validation: type-check and lint pass. All 18 updater/notification tests pass. Full-suite CLI timeouts pass when rerun separately (20 tests); rule tests pass (76). Three unrelated API cases still hit their existing 5-second timeout locally after a serial retry; CI is running. Production build passes. Tests cover pending requests, unchanged/older versions, unstable releases, failed requests, timeout and session closure.

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -15,7 +15,7 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ### Fixed
 
-- Notify about newer CLI releases when an interactive session starts, without blocking work (#TBD).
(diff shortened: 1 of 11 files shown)


gtanczyk:
@codex

PR #1263, opened 2026-09-11
fix(delivery): recover from throttled staging writes

gtanczyk:
A local `/push --takeover` can disconnect the old agent successfully, then fail while uploading sources: rapid writes to the shared GCS staging manifest return 429 and the creator only sees `internal`.

Retry GCS writes on 429 with bounded exponential backoff and jitter, respecting Retry-After. Keep generation preconditions intact so a competing manifest update is re-read and merged through the existing 412 path. After retries are exhausted, return a safe 503 response and show a retry hint in the CLI. The takeover hint now explicitly distinguishes `/push --takeover` in the interactive session from `gamedevpl push --takeover` in a shell.

Extract the existing API error handler without changing handling of unrelated errors; only the typed storage-busy error gets the actionable response.

Validation: 84 targeted API tests and 17 delivery CLI tests pass. These cover manifest throttling plus concurrent updates, retry bounds, unchanged local files, and successful redelivery. Full type-check and lint pass. API and CLI suites pass; two web tests failed in the full run and both pass when rerun separately (8 tests). Rule tests pass (76). Production build passes; all GitHub CI checks are green.


gtanczyk:
@codex

PR #1267, opened 2026-09-11
feat(cli): configure Antigravity headless permissions

gtanczyk:
Antigravity local delegation currently discovers missing command permissions only after starting, then asks the creator to switch terminal interfaces. Offer sandboxed headless setup before preparing the checkout or launching the agent.

The picker explicitly explains that this updates global agy settings for every project. Confirmation sets `enableTerminalSandbox=true` and `toolPermission=proceed-in-sandbox`, preserves existing rules and unrelated settings, and saves an exact backup. Subsequent runs detect the preset. Creators can instead keep current settings or cancel; unattended runs never configure permissions. Existing interactive recovery remains available for denied tools. Setup is offered only on macOS/Linux, where agy documents terminal sandbox support.

No wildcard permissions or permission-bypass flags are added. Existing outside-sandbox exceptions remain in effect. The settings writer rejects malformed profiles and refuses to replace symlinks, while existing-permissions and unattended runs can use symlinked settings. It also detects edits made while the picker is open.

Validation: all 380 CLI tests passed with local-port access, including 26 focused setup/fallback tests. Repository type-check and lint passed. All GitHub CI checks passed at 401fb49b0, including full monorepo lint/type-check/tests/build, CLI bundle smoke test, security checks and production image. The parallel local monorepo run encountered localhost EPERM and timing failures and was stopped after CI completed. The installed agy help supports `--sandbox`, and its binary contains both preset settings; no paid model task or user-global configuration was changed during verification. Reference: https://antigravity.google/docs/cli/sandbox/.

Instrumentation: affects creator funnel question 4. Existing `cli_step` / `delegate_used` records launches only after setup; cancelled setup does not count as a delegation. Permission-choice conversion is not separately measured.

Review follow-up: allow valid symlinked settings for current permissions, unattended runs and an already-enabled preset. Symlink replacement remains refused only after choosing setup. All 26 focused tests, CLI type-check and changed-file ESLint pass; CI is rerunning for this follow-up.

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -7,6 +7,10 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ## Unreleased
 
+### Added
+
+- Configure remembered sandboxed headless permissions before starting local Antigravity tasks.
+
 ## 0.13.1 — 2026-09-11
 
 ### Fixed
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -9,7 +9,7 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 
 ### Added
 
-- Configure remembered sandboxed headless permissions before starting local Antigravity tasks.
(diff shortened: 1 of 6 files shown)


gtanczyk:
@codex

PR #1268, opened 2026-09-11
Redirect store-published media too, not only snapshot games

gtanczyk:
#1261 moved media off the origin for games baked into a snapshot, and quietly missed the ones made on the platform. Those keep their media in the **store** bucket, under `games/<slug>/versions/<version>/media/`, and they are **62 games against the snapshot's 123** — and the heavier half.

The 662 KB capture I measured to justify the whole exercise belongs to `airtime`, a prompt-made game. It was still being served inline after #1261 shipped. Verified on production:

```
arena-tag/gameplay.mp4  -> 302 storage.googleapis.com   (snapshot game)
airtime/gameplay.mp4    -> 200 through the origin       (platform game)
```

## What changes

A second signer, bound to `GAMES_STORE_BUCKET`. The store path runs the same checks it always did — publication state, then the media allow-list from the version's `metadata.json` — and only then hands out a URL for `games/<slug>/versions/<version>/media/<file>`.

**Signing now verifies the object exists first.** A redirect cannot fall through to the next source the way an inline read can, so an absent object returns null and the route serves bytes instead of 302-ing to a 404. The probe runs only on a cache miss; a cached URL is its own proof that the object was there.

## A bug the tests caught, mine

`reply.redirect()` resolves to `undefined`. My refactor wrote `if (redirected) return redirected`, so after issuing a redirect the handler **kept running** and read the bytes it had just avoided — the response was a correct 302 with the work done anyway. `expect(snapshot.getMedia).not.toHaveBeenCalled()` is what caught it; without that line the suite would have been green and the change worthless. The helper returns a boolean now.

## Tests

290 passing across four suites. New: store media redirects to a signed URL, falls back to inline when the object cannot be signed, and refuses a file the published metadata does not list.


gtanczyk:
@codex

PR #1271, opened 2026-09-11
fix(cli): recover Muse sessions waiting for approval

gtanczyk:
Muse headless execution can wait indefinitely for a tool approval that never reaches JSON stdout. The terminal then reports a running tool with no output even though no tool process has started.

Follow only the active session's append-only journal (session ID and timestamp from its stdout envelope). On an approval request, terminate that headless child, retain the overall CLI task, and offer to resume the exact session in interactive Muse. Declining preserves edits and skips verification/delivery; unattended execution prints a resume command and returns without delivery. Model, permission profile and restrictive sandbox/tool settings are retained; approvals are never auto-granted. The activity line also follows model-wait events instead of retaining stale tool status.

The journal reader consumes bounded chunks, carries partial lines and skips oversized unrelated records. Parent cancellation and completion clean up the watcher. `--no-session-log` is refused because this recovery requires the journal. Existing Antigravity permission recovery shares the extracted handoff helper.

Validation: 411 CLI tests passed, including a silent fake run whose approval exists only on disk, partial/oversized records, cancellation, exact-session resume, decline, failure and unattended behavior. Real installed Muse `exec --provider echo` confirmed the stdout session envelope and XDG journal layout without a paid model call. Repository type-check and lint passed. The new reader also detected the recorded approval in the actual reported session journal (read-only). Full local monorepo tests encountered timing failures outside CLI. All GitHub CI checks passed at 71d3b5492, including the full lint, test and build gate and production image build. The user's active Muse session was inspected read-only, not approved or stopped.

Instrumentation: creator funnel question 4 uses existing `delegate_used` and local activity `permission`/`interactive` phases. No commands, journal content, prompts or session IDs are added to telemetry.


gtanczyk:
@copilot resolve the merge conflicts in this pull request

PR #1277, opened 2026-09-12
fix(cli): open busy preview with keyboard

gtanczyk:
Some terminals render preview URLs as plain colored text even when the CLI emits an OSC 8 hyperlink, leaving creators unable to open or select the URL while an agent owns the input.

Remember the active local preview URL in the TUI and show `o open preview` during busy work. Pressing `o` opens that exact URL through the existing platform-specific browser launcher without interrupting the agent. If launching fails, the CLI prints a stable copyable URL. Normal prompt input is unchanged.

Validation: the full repository type-check, lint, test, and build gate passed. The complete CLI suite passed (70 files, 404 tests), including the real Ink input path for opening a preview while busy and the existing narrow-terminal coverage.

Instrumentation: this affects creator funnel question 4. Existing `cli_step=play_requested` counts the shortcut alongside other preview requests and remains queryable by step and visit ID. It records no URL, slug, checkout path, or new dimension.

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -10,7 +10,7 @@ decides the next version — see [`.claude/skills/cli-release/SKILL.md`](../../.
 ### Fixed
 
 - Recover Muse tasks waiting for approval by offering to resume the same session interactively (#1271).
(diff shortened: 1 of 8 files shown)


gtanczyk:
@codex

PR #1286, opened 2026-09-12
Classify game merges so snapshot publish is not a 45-minute catalog gate

gtanczyk:
## Problem

[Publish games snapshot #34686093184](https://github.com/gamedevpl/www.gamedev.pl/actions/runs/34686093184) is the current merge path: every `games-published` dispatch runs `check:pr` over the whole catalog (~40 min), then WebKit, then the bake.

A successful run the same morning ([#34680596062](https://github.com/gamedevpl/www.gamedev.pl/actions/runs/34680596062)) was **47 minutes** wall clock. Because this workflow cancels in-progress publishes, a merge train that lands another game before that job finishes never moves `current.json`.

Classifying only `HEAD^...HEAD` is also unsafe with `cancel-in-progress`: if runtime change A starts a full gate and is cancelled when game-only B lands, B would check only that game and bake A's unvalidated modules into every title.

## Change

On `games-published` (not `games-validate`), after checking out the pinned SHA, re-classify with `tools/gate-scope.mjs --event push --base <last-published-sha> --diffs-json` from the compare API. The base is `current.json`'s `commitSha` (read via WIF), not HEAD's parent:

- **one game** (and no unvalidated commits since the live snapshot) → scoped gate, no WebKit → bake.
- **opt-in GameKit** → `static` on `GAME.json` consumers, with WebKit.
- **named universal function / field** → `static` on callers, with WebKit.
- **unidentifiable or catalog-wide kit**, `game-shell.css` → full.
- **cancelled/failed A then game-only B** → compare is last-good...B, so A's files stay in range.
- **missing pointer, compare, or a classify crash** → full (the bake still runs). Payload mode is not a fallback on `games-published`.
- **empty `need_webkit` / `narrow_conformance` / `run_gate`** → same full path as a missing mode (cannot hard-exit `resolve` and freeze the pointer).
- **docs-only / deleted-only** (`run_gate=0`) → skip the gate, still bake.

`games-validate` keeps the payload it was sent (including WebKit) and does not publish.

If GitHub omits a patch and the published SHA is not in the checkout, classify fails closed to full. It does not fall back to `HEAD~1` (that hole is closed in games#1425).

WebKit on a full run starts from the resolve job, so it overlaps the catalog gate. A red gate still pays that sweep; `cancel-in-progress` makes that rare.

Pair: [games#1425](https://github.com/gamedevpl/www.gamedev.pl-games/pull/1425).

Do not merge or deploy until re-review.

## Test plan

- [x] Website files are workflow + docs only
- [x] `classify_push` is `games-published` only
- [x] Classify `last-published...HEAD`, not `parents[0]`; missing pointer → full
- [x] Classify crash / empty mode / empty flags → full, not a frozen pointer
- [x] Regression in games#1425: cancelled runtime + game-only file list stays `full`
- [x] Regression in games#1425: named base + missing S + no API patch stays `full`
- [x] CI on `10ad195` (previous head) was 8/8 green; this commit is the jq-flag guard
- [ ] After both PRs merge, a `gfx3d` landing should log `Push scope: mode=static`
- [ ] A `drawGrid` landing should log `mode=static` with the `draw.grid` games
- [ ] Nightly 04:23 UTC and Sunday 06:17 UTC still run full

<div><a href="https://cursor.com/agents/bc-2a8c444b-7a95-431e-952d-669840349ad4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2a8c444b-7a95-431e-952d-669840349ad4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/games-catalog-gate.yml
+++ b/.github/workflows/games-catalog-gate.yml
@@ -258,13 +258,24 @@ jobs:
                 fetch --depth=1 "https://github.com/${GAMES_REPO}.git" "$parent" \
                 || echo "Baseline fetch failed — classify from compare patches only"
             fi
+            set +e
             scope=$(printf '%s\n' "$files" | node tools/gate-scope.mjs --event push --base "$parent" --diffs-json "$diffs_json" --json)
+            status=$?
+            set -e
             rm -f "$diffs_json"
-            MODE=$(printf '%s' "$scope" | jq -r '.mode')
+            if [ "$status" -ne 0 ]; then
+              echo "gate-scope exited ${status} — full catalog gate"
+              return 1
+            fi
+            MODE=$(printf '%s' "$scope" | jq -r '.mode // empty')
+            if [ -z "$MODE" ] || [ "$MODE" = null ]; then
+              echo "gate-scope returned no mode — full catalog gate"
+              return 1
+            fi
             SLUGS=$(printf '%s' "$scope" | jq -r '.slugs | join(" ")')
-            NEED_WEBKIT=$(printf '%s' "$scope" | jq -r '.need_webkit')
-            NARROW_CONFORMANCE=$(printf '%s' "$scope" | jq -r '.narrow_conformance')
-            RUN_GATE=$(printf '%s' "$scope" | jq -r '.run_gate')
+            NEED_WEBKIT=$(printf '%s' "$scope" | jq -r '.need_webkit // empty')
+            NARROW_CONFORMANCE=$(printf '%s' "$scope" | jq -r '.narrow_conformance // empty')
+            RUN_GATE=$(printf '%s' "$scope" | jq -r '.run_gate // empty')
             echo "Push scope: mode=${MODE} run_gate=${RUN_GATE} webkit=${NEED_WEBKIT} slugs=${SLUGS:-"(none)"}"
             return 0
           }
@@ -469,6 +480,8 @@ jobs:
           ' $SLUGS)
           if [ "$need_webgl" = 1 ]; then npm run check:webgl; fi
 
+  # Overlaps catalog-gate (needs: resolve, not catalog-gate). A red gate still
+  # pays this sweep; cancel-in-progress makes that rare.
   webkit:
     needs: resolve
     if: needs.resolve.outputs.need_webkit == '1'
--- a/.github/workflows/games-catalog-gate.yml
+++ b/.github/workflows/games-catalog-gate.yml
@@ -276,6 +276,10 @@ jobs:
             NEED_WEBKIT=$(printf '%s' "$scope" | jq -r '.need_webkit // empty')
             NARROW_CONFORMANCE=$(printf '%s' "$scope" | jq -r '.narrow_conformance // empty')
             RUN_GATE=$(printf '%s' "$scope" | jq -r '.run_gate // empty')
+            if [ -z "$NEED_WEBKIT" ] || [ -z "$NARROW_CONFORMANCE" ] || [ -z "$RUN_GATE" ]; then
+              echo "gate-scope returned incomplete flags — full catalog gate"
+              return 1
+            fi
             echo "Push scope: mode=${MODE} run_gate=${RUN_GATE} webkit=${NEED_WEBKIT} slugs=${SLUGS:-"(none)"}"
             return 0
           }
--- a/.github/workflows/games-catalog-gate.yml
+++ b/.github/workflows/games-catalog-gate.yml
@@ -9,9 +9,9 @@ name: Games catalog gate
 # reverse-index to GAME.json consumers and run static + those slugs.
 # Those two catalog seals are the Sunday 06:17 UTC cron ('17 6 * * 0'); scoped
 # still playtests touched slugs. A games-published dispatch re-classifies the
-# SHA with tools/gate-scope.mjs --event push so a one-game merge does not wait
-# on a full-catalog check:pr (measured ~40 min). The daily snapshot is a cheap
-# rebuild, not a second full-catalog seal.
+# SHA with tools/gate-scope.mjs --event push --base <parent> --diffs-json so a
+# one-game merge does not wait on a full-catalog check:pr (measured ~40 min).
+# The daily snapshot is a cheap rebuild, not a second full-catalog seal.
 
 on:
   workflow_call:
@@ -178,6 +178,8 @@ jobs:
           token: ${{ secrets.GAMES_REPO_TOKEN }}
           path: games-repo
           persist-credentials: false
+          # HEAD~1 for --base fallback when GitHub omitted a file patch.
+          fetch-depth: 2
 
       - name: Resolve checked out games commit
         id: checkout_sha
@@ -206,7 +208,11 @@ jobs:
 
           classify_from_parent() {
             local parent="$1"
-            local compare status files file_count scope
+            local compare status files file_count scope diffs_json
+            if [ "${#parent}" -ne 40 ] || printf '%s' "$parent" | grep -q '[^0-9a-f]'; then
+              echo "Parent SHA is not 40 hex — full catalog gate"
+              return 1
+            fi
             compare=$(mktemp)
             status=$(curl -sS -o "$compare" -w '%{http_code}' \
               -H "Accept: application/vnd.github+json" \
@@ -225,8 +231,19 @@ jobs:
               return 1
             fi
             files=$(jq -r '.files[]? | [.filename, (.previous_filename // empty)] | .[] | select(. != "")' "$compare")
+            diffs_json=$(mktemp)
+            jq -c '[.files[]? | select(.filename != null and .patch != null) | {key: .filename, value: .patch}] | from_entries' "$compare" > "$diffs_json"
             rm -f "$compare"
-            scope=$(printf '%s\n' "$files" | node tools/gate-scope.mjs --event push --base "$parent" --json)
+            # Checkout is shallow. Fetch the API parent so --base can git-diff
+            # when GitHub omitted a patch (binary / too large). Compare patches
+            # are the primary input and work without this object.
+            if ! git cat-file -e "${parent}^{commit}" 2>/dev/null; then
+              git -c "http.extraheader=AUTHORIZATION: bearer ${GAMES_REPO_TOKEN}" \
+                fetch --depth=1 "https://github.com/${GAMES_REPO}.git" "$parent" \
+                || echo "Parent fetch failed — classify from compare patches only"
+            fi
+            scope=$(printf '%s\n' "$files" | node tools/gate-scope.mjs --event push --base "$parent" --diffs-json "$diffs_json" --json)
+            rm -f "$diffs_json"
             MODE=$(printf '%s' "$scope" | jq -r '.mode')
             SLUGS=$(printf '%s' "$scope" | jq -r '.slugs | join(" ")')
             NEED_WEBKIT=$(printf '%s' "$scope" | jq -r '.need_webkit')
--- a/.github/workflows/games-catalog-gate.yml
+++ b/.github/workflows/games-catalog-gate.yml
@@ -2,9 +2,11 @@ name: Games catalog gate
 
 # Runs the games repository's real post-merge gate on public Actions minutes.
 # publish-games.yml calls this workflow before advancing the snapshot pointer.
-# mode=full on schedule / manual dispatch / universal GameKit (core/gfx/audio/…):
-# playtest --suite default, no catalog agent-play. Opt-in modules (gfx3d,
-# sensing, …) reverse-index to GAME.json consumers and run static + those slugs.
+# mode=full on schedule / manual dispatch / unidentifiable universal GameKit
+# patches (and catalog-wide fields like draw.rect): playtest --suite default, no
+# catalog agent-play. Named function/field edits of core/drawing/gfx/… run
+# static on the games that call that surface. Opt-in modules (gfx3d, sensing, …)
+# reverse-index to GAME.json consumers and run static + those slugs.
 # Those two catalog seals are the Sunday 06:17 UTC cron ('17 6 * * 0'); scoped
 # still playtests touched slugs. A games-published dispatch re-classifies the
 # SHA with tools/gate-scope.mjs --event push so a one-game merge does not wait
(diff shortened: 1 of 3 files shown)


gtanczyk:
@codex

gtanczyk (commented):
## Review: does this solve the problem, and in the right way?

**The problem is real and the shape of the fix is right.** Splitting `resolve` out of `catalog-gate`, re-classifying the dispatched SHA against its parent, and letting a one-game merge take the scoped path is the correct lever: that path is `check:game` on the same tree the PR already attested, and it turns a ~45 min gate into ~8–10 min, which is what stops the merge train from cancelling `current.json` moves.

Two things I think block the claimed outcome, and one design call worth making explicitly.

---

### 1. The universal-module narrowing cannot fire — the games checkout is shallow (blocking)

`resolve` checks out the games repo with no `fetch-depth`, so `actions/checkout` clones at depth 1. `gate-scope.mjs --base <parent>` then shells out to:

```
git diff -U3 <parent>...HEAD -- shared/modules/drawing.ts
git diff -U3 <parent> HEAD   -- ...
git diff -U3 <parent>        -- ...
```

None of those can resolve `<parent>` in a depth-1 clone, so `getGitDiffForFile` returns `null`, `extractChangedSymbolsFromDiff(null)` returns `null`, and the module falls back to `moduleAffected = 'all'` → `full`.

Reproduced against the games PR head with an unreachable base:

```
$ printf 'shared/modules/drawing.ts\n' | node tools/gate-scope.mjs --event push --base 000...0 --json
{"mode":"full","slugs":[],"need_chrome":1,"need_webkit":1,"run_gate":1,"narrow_conformance":0}
```

Consequences:

- Test-plan item *"A `drawGrid` landing should log `mode=static` with the four `draw.grid` games"* will not pass as written — it will log `Run the full games gate`.
- Worse: because `classify_push: '1'` makes this re-classify the **authority** and discards the payload, the webhook's `mode=static` — computed in games#1425 from the real GitHub `patch` fields, i.e. from *more* information than this job has — gets thrown away and replaced with `full`. So the whole `kit-surface.mjs` half of the pair is dead on the merge path.

Opt-in modules (`gfx3d`, verticals) and one-game merges are unaffected: those classify from `GAME.json` / the file list and need no diff. So the PR still delivers its headline win; it's the universal-module tier that silently degrades.

Fix, cheapest first:
- `fetch-depth: 2` on the `resolve` checkout (the parent is all `<parent>...HEAD` needs for a first-parent push), or
- pass the compare API's per-file `patch` (you already fetch and `jq` that JSON) into `gate-scope` via a `--diffs` file, and have the website stop depending on git history at all. This is the more honest version — it makes the website classify from exactly the same input the webhook used.

Either way, please make the fallback visible: log `Push scope: universal module, no diff available — full` rather than having it look like a deliberate `full`.

### 2. A classifier crash blocks the bake entirely, instead of falling back to full

Inside `Classify merge scope` you run under `set -euo pipefail`:

```sh
scope=$(printf '%s\n' "$files" | node tools/gate-scope.mjs --event push --base "$parent" --json)
MODE=$(printf '%s' "$scope" | jq -r '.mode')
```

Every *other* failure mode in this step fails closed to `full` and still bakes: non-200 compare, missing parent, the 300-file cap. But if `gate-scope.mjs` throws (a bad `GAME.json`, a `shared/modules` read error, a future refactor), or `jq` gets non-JSON on stdout, the command substitution fails, the step exits non-zero, `resolve` fails, `catalog-gate` never runs, **and `publish` never runs** — the pointer silently freezes until the 04:23 cron. That's the exact failure mode the header comment in `publish-games.yml` warns about.

Suggest wrapping the classify the same way `classify_from_parent` wraps the compare: capture the exit status, and on anything unexpected `MODE=full; SLUGS=; NEED_WEBKIT=1; RUN_GATE=1` with a loud log line. Also guard against `jq -r '.mode'` yielding the literal `null` — today that reaches the `case` and hard-fails.

### 3. WebKit no longer waits for the gate — deliberate?

`webkit` moved from `needs: catalog-gate` to `needs: resolve`. On the happy path that's the point (it overlaps the ~40 min catalog gate, good). The cost is that a red gate now also pays a full WebKit sweep in parallel instead of being short-circuited. Given `cancel-in-progress` and how rare a red post-merge gate should be, I think that trade is fine — but it's a real change in minute spend on failures and it isn't mentioned in the PR body. Worth one line in the header comment so the next person doesn't "fix" it back.

---

### Smaller notes

- `publish` still resolves because `resolve` always runs, so `run_gate=0` → `catalog-gate` skipped → reusable workflow is `success` → bake proceeds. Correct, but it depends on `resolve` never being skipped; worth a comment, since a reusable workflow whose jobs are *all* skipped reports `skipped` and would take `publish` down with it.
- The 8-case `case "$A:$B:$C"` flag validation is now duplicated in `resolve` and `scope` and will get worse with a fourth flag. A tiny `is01()` shell helper would read better, but not worth a round-trip on its own.
- Relaxing the `scoped && no slugs` guard with `[ "$CLASSIFY_PUSH" != 1 ]` is fine — `scope` re-validates after classification, and `classify([])` legitimately returns `mode=scoped, slugs=[], run_gate=0`. Good catch that it needed relaxing.
- Docs update in `docs/games-snapshot.md` is accurate and genuinely useful; it will just describe a path that doesn't execute until (1) is fixed.

### Verdict

Right problem, right architecture, and the one-game fast path will work as soon as it merges. But please fix (1) before merging, otherwise games#1425's universal-module analysis buys nothing on the merge path and the test plan's `drawGrid` row will quietly fail. (2) is the one that can cost you a frozen catalog, so I'd take it in the same push.

---

.github/workflows/games-catalog-gate.yml
```diff
@@ -170,6 +184,147 @@ jobs:
...
+              "https://api.github.com/repos/${GAMES_REPO}/commits/${SHA}")
+            if [ "$status" != 200 ]; then
+              echo "Commit lookup returned ${status} — full catalog gate"
+              MODE=full
+              SLUGS=
+              NEED_WEBKIT=1
+              NARROW_CONFORMANCE=0
+              RUN_GATE=1
+            else
+              parent=$(jq -r '.parents[0].sha // empty' "$commit")
```
gtanczyk: [P1] Classify all changes since the last successfully validated/published snapshot. `parents[0]` only covers the final commit, but `publish-games.yml` still uses `cancel-in-progress: true` and bakes the entire tree. If A changes shared runtime and its full gate fails or is cancelled when B changes one game, B checks only that game and can publish A's unvalidated runtime to every game. A multi-commit push also loses all but its last commit even though the webhook compared the complete push. Persist/use the successful snapshot or validation baseline and classify the cumulative range; use full mode when that baseline is unavailable. Add a regression for failed/cancelled A followed by game-only B.

gtanczyk (commented):
Reviewed exact HEAD `dc8317b1e74bedb7c507a3e5f9812062991b9e44` together with games#1425 at `4a1d4ee7381dbadbbab3235b20d8a595ae121921`.

Scoped game gates, the opt-in module consumer list, and overlapping WebKit with the catalog gate are sensible ways to reduce latency. This does not yet safely solve the merge-train problem: cancellation remains enabled while scope covers only the final commit, so an earlier failed/cancelled gate can be bypassed by the next publication (inline finding). A faster gate also still starves if merges arrive faster than it completes; publication needs an explicit accumulated validation range and a scheduling policy that eventually lets a publish finish.

I independently reproduced the shallow-checkout issue already reported in the existing review: a helper-signature edit classifies `static` with full history and `full` with `--depth=1`. With the paired games HEAD this falls back to full, rather than producing empty classifier output. The existing `games-validate` contract finding should also be addressed.

Validation: inspected both pinned workflow/source trees; ran a local shallow/full-history classifier reproduction. Games classifier/dispatch tests passed 56/56 and webhook tests 1/1. No full browser gate, Actions publication, or production deployment was run. I would fix these gaps before merging.

gtanczyk:
### Review PR #1286

Koncepcja jest bardzo dobra i potrzebna: re-klasyfikacja commitów w workflow witryny, wydzielenie etapu `resolve`, równoległe uruchamianie WebKita z bramką katalogową oraz pominięcie WebKita dla pojedynczych gier na pushu (`mode=scoped`) realnie skrócą czas publikacji snapshotów z ~47 minut do kilku minut.

Podczas weryfikacji zidentyfikowałem jednak **krytyczny problem wykonawczy w GitHub Actions** oraz rozbieżność w logice fallbacku:

---

#### 1. 🚨 Klasyfikacja modułów universal zawsze wpadnie w `mode=full` przez `fetch-depth: 1`
W jobie `resolve` krok checkoutu gier wygląda tak:
```yaml
- name: Check out games repo
  uses: actions/checkout@...
  with:
    repository: ${{ env.GAMES_REPO }}
    ref: ${{ steps.resolve.outputs.games_ref }}
    path: games-repo
    persist-credentials: false
```
Domyślnie `actions/checkout` tworzy **płytki klon o głębokości 1** (`fetch-depth: 1`). Oznacza to, że w lokalnej bazie repozytorium znajduje się wyłącznie pobrany commit `SHA`.

Następnie w `classify_from_parent`:
```bash
scope=$(printf '%s\n' "$files" | node tools/gate-scope.mjs --event push --base "$parent" --json)
```
Gdy zmieniony jest jakikolwiek plik z `UNIVERSAL_MODULES` (np. `shared/modules/drawing.ts`), `gate-scope.mjs` próbuje wykonać:
```bash
git diff -U3 $parent...HEAD -- shared/modules/drawing.ts
```
Ponieważ obiekt `$parent` nie istnieje w płytkim klonie, git kończy działanie błędem:
`fatal: bad object <parent>`.
`getGitDiffForFile` przechwytuje wyjątek i zwraca `null`, co powoduje fail-closed do `moduleAffected = 'all'` i w konsekwencji:
```
Push scope: mode=full run_gate=1 webkit=1 slugs=(none)
```
W rezultacie test z planu:
`- [ ] A drawGrid landing should log mode=static with the four draw.grid games`
w środowisku GitHub Actions **zawsze uruchomi pełną bramkę katalogową (~40 min)**.

**Proponowane rozwiązanie:**
Krok `classify_from_parent` pobiera już z GitHub API pełną odpowiedź:
`curl -sS -o "$compare" ... "https://api.github.com/repos/${GAMES_REPO}/compare/${parent}...${SHA}"`
Obiekt `$compare` **już zawiera kompletne diffy w `.files[].patch`**. Zamiast usuwać go przed uruchomieniem `gate-scope.mjs` i zmuszać gita do lokalnego diffowania, warto przekazać ten plik bezpośrednio do `gate-scope.mjs` (np. `--compare "$compare"`). Wtedy analiza symboli nie zależy w ogóle od stanu lokalnego repozytorium git ani głębokości klonowania.

---

#### 2. Niespełniona obietnica fallbacku do wartości z payloadu
W opisie PR czytamy:
> *"Merge dispatches always re-classify against the SHA. Payload mode is a hint (and the fallback if compare fails)"*

W kodzie natomiast:
```bash
if [ "$CLASSIFY_PUSH" = 1 ]; then
  ...
  if [ "$status" != 200 ]; then
    echo "Commit lookup returned ${status} — full catalog gate"
    MODE=full
    SLUGS=
    ...
  elif ! classify_from_parent "$parent"; then
    MODE=full
    SLUGS=
    ...
```
Jeśli compare API zwróci błąd (np. rate limit, chwilowy błąd sieci) lub brak rodzica, skrypt bezwarunkowo wymusza `MODE=full`, ignorując wartości `$INPUT_MODE` i `$INPUT_SLUGS`, które Cloud Run poprawnie wyliczył i przysłał w dispatchu. Jeśli payload ma być faktycznym fallbackiem, to przy niepowodzeniu compare powinien zachować `$INPUT_MODE` i `$INPUT_SLUGS`.

gtanczyk (commented):
Follow-up at `10ad195407e76ce8dd46faa4118204fbd5fc1a92`, paired with games `17d79c420bfcebd3e4d7935f2c8af810f69c55f6`.

The main earlier findings are addressed in the workflow: baseline comes from current.json.commitSha; compare patches reach the classifier directly; games-validate preserves its payload; missing pointer/classifier failure selects full. This fixes the normal cumulative-range and shallow-checkout paths.

One integration edge still defeats the cumulative-range guarantee: if the compare omits a universal-module patch and fetching the baseline fails, gate-scope now falls back to HEAD~1. A local three-commit reproduction (published S, runtime A, helper B) returns full with S available but static in the depth-2 checkout without S, omitting A. Reported inline in games#1425. An unavailable explicit baseline must fail closed, not silently change the comparison range.

The original cancel-in-progress scheduling limitation also remains: narrowing reduces run time but does not guarantee progress when merges keep arriving faster than a gate finishes. No Actions publication or production smoke test was run during this review.

gtanczyk:
Re-reviewed at `10ad195`. Both blockers from my earlier review are cleared, and the fix is better than what I suggested:

- **Shallow checkout** → solved by not depending on git at all: `--diffs-json` from the compare API is the primary input, with a `--depth=1` fetch of the baseline only as a fallback for patches GitHub omits. Verified that `main`'s current `gate-scope.mjs` ignores the unknown `--diffs-json` flag harmlessly (stdin still wins), so merging this before games#1425 degrades to `full` rather than misclassifying — the two PRs can land in either order.
- **Classify crash freezing the pointer** → `set +e` around the node call, exit-status check, and `.mode // empty` all return 1 into the `full` fallback. Pointer-read failure and non-hex baseline take the same path, and `gcp_auth` is `continue-on-error`, so a WIF hiccup degrades to `full` instead of failing the job.

The `HEAD^` → `current.json` `commitSha` change is a real bug I missed: with `cancel-in-progress`, a cancelled full gate for a runtime change followed by a game-only merge would have classified only the game and baked unvalidated modules. Classifying cumulatively from the live snapshot is the correct base and makes this strictly safer than what's on master today. `id-token: write` is present in both the caller and the callee, so the auth step will actually work.

One residual, non-blocking: inside `classify_from_parent`, only `mode` is guarded against an empty jq result. If `need_webkit` / `narrow_conformance` / `run_gate` ever came back empty with a valid `mode`, the function returns 0 and the flag `case` below hard-exits — the same "no gate, no bake, frozen pointer" shape you just fixed for `mode`. Practically unreachable given `gate-scope` always emits all five, but the same `// empty` → `return 1` treatment costs three lines.

CI on this head is still pending (the green run is from the previous commit) — worth a look before merging, but nothing in the diff suggests a problem. LGTM once it's green.

---

PR #1292, opened 2026-09-12
fix(cli): expose agent model settings

gtanczyk:
The task builder picker currently hides the local agent model and reasoning effort, so users choose an executor without knowing its saved configuration or having a way to change it in that flow.

Each local-agent row now leads with its configured model and effort. Unreported native defaults are labelled explicitly, and a Configure option opens the existing model picker before returning to the refreshed executor list.

Validation: `npm run type-check && npm run lint && npm run test && npm run build`.

Instrumentation: this changes local agent selection only. It adds no event or field; existing per-run CLI funnel telemetry remains unchanged.


gtanczyk:
@codex review

gtanczyk:
@codex review

PR #1296, opened 2026-09-12
feat(editorkit): layered collection items — contract mirror and Studio rendering

gtanczyk:
## Why

Paired half of a games-repo change that adds `layered` to the closed collection-item vocabulary.

`layers` shipped as a top-level declaration beside `content`, which makes it **one board per game**. The Lemmings case that motivated EditorKit v2 needs **one board per level** — terrain, objects and triggers stacked on each mission of a pack. Collection items could only be `tilemap`, `entities` or `path`, none of them layered, so a pack could not express it at all. Measured across the catalog, exactly one game of the seven with a tilemap collection could adopt layers as shipped.

## 1. The contract mirror

The two contract files are a lockstep. This copy validates Studio drafts on write (`editor-drafts.ts:305`) and regenerates the L1 content module on publish; the games repo's Check 31 validates every delivery. A one-sided change would either refuse content the gate accepts, or accept content the gate refuses.

So the mirror is **copied wholesale from the authority, not hand-edited**, and verified byte-for-byte below the header afterwards — including after the repo's prettier hook ran on each commit. Nothing in that file is website-only.

A layered item declares its own `layers` and its own cross-layer `constraints`, so each level in a pack carries a full stack whose rules are evaluated per level. Three pieces already existed and are now shared rather than duplicated: layer-spec validation, the shared-grid and cell-budget check (extracted as `checkLayerGrids`), and cross-layer reachability. Two of them hardcoded a `layers` prefix in their messages and now take the owner instead, so a failure inside a pack reads `levels: item 2 constraints[0]: walled off …` rather than naming a top-level layer that does not exist. **Top-level messages are unchanged, byte for byte** — the existing tests pin them.

## 2. Studio renders one

Originally staged as a follow-up, folded in here instead: a contract that accepts a shape the shell cannot draw would let a creator author a layered pack and see nothing. Both halves land together.

A layered collection keeps **both** choices on screen, because a pack needs both — the mission list stays where it is, and the layer picker sits beside it. The board reuses `LayeredBoard`, so the stack, the picker and the read-only underlay behave exactly as they do for top-level layers.

`layeredItemView` derives the visible stack — which layers, the selected layer, the active tile — as a pure function in `editorPanelHelpers`. That keeps the panel's growth to the two handlers and two JSX branches that genuinely close over its state, and makes the subtle part testable on its own: **a tile key left over from another layer's palette paints nothing**, so the view falls back to that layer's first tile.

That bug is worth naming, because reading the diff would not have caught it. The first version rendered the board correctly and silently ignored every click — the palette key was still the top-level one, which does not exist for an item's layers. It surfaced only by driving the real panel and dumping what it produced.

## Test plan

- [x] `npm run type-check` clean for `@gamedevpl/api` and `@gamedevpl/web`
- [x] `npx vitest run --root apps/api src/creation src/catalog/…` — 657 pass, including the twin drift check
- [x] `npm test` — 676 test files pass across all workspaces
- [x] `npm run lint` — full chain clean
- [x] Three rendering tests, each confirmed to fail with the render branch disabled
- [x] Byte-equivalence below the header re-verified against the games repo's committed file, after both commits

One flaky failure appeared in a single full-suite run (`App.createGameFocus` party navigation) and did not reproduce: green in isolation, green in a full `apps/web` run of 2026 tests, green in a clean full `npm test`. Flagging it rather than burying it.

### Ratchets

Four baselines move. `editorPanelHelpers.ts` and the new test file are new entries. `studioApi.ts` (+10, pure type declarations) and `EditorPanel.tsx` (+46 after the extraction, down from +63) use the scoped `--write --force` the tool documents, as does the mirror: **a mirror cannot be extracted from without breaking the very lockstep it exists to keep.**

## Staging

Contract and rendering. No game declares a layered item yet — adoption (`lemming-rescue`) comes next, now that there is a surface to adopt into.

https://claude.ai/code/session_01QKNijFdYrCvAxNjNc69DbV


gtanczyk:
@codex

PR #1302, opened 2026-09-12
fix(api): reply once when a route guard already sent the response

gtanczyk:
## What

Route handlers that used a guard which sends its own error response and returns a falsy value followed it with a bare `return`. Every such site now returns the reply instead.

## Why

`GET /api/me/quota` without a session produced an unhandled rejection on clean `master`:

```
Error: Cannot write headers after they are sent to the client   (ERR_HTTP_HEADERS_SENT)
  at onSendEnd (fastify/lib/reply.js:641)
```

Mechanism:

1. `checkUserAccess` sends the 401 itself and returns `false`.
2. The handler does a bare `return;`, so its promise resolves with `undefined`.
3. Fastify 5 `wrap-thenable.js` sends *again* when the payload is `undefined` and `reply.sent === false`.
4. `reply.sent` is `raw.writableEnded`, which is still false here, because the auth `onSend` hook (`platform/auth.ts`) is `async` and the first send is still working through the hook chain.
5. The second send re-runs `onSend`, and its `onSendEnd` calls `writeHead` on already-sent headers.

The response was a correct 401 either way, so nothing was user-visible — the request was simply answered twice.

The fix is Fastify's documented idiom, already used at `realtime/mp.ts:588`: `return reply`. `Reply` is thenable (`reply.js:476`), so an async handler awaits the real send; the promise resolves with `undefined` only once `reply.sent` is true, and the second send never fires.

## Scope

The same defect class was present at every call site of this shape, not just the quota route:

- `checkUserAccess` — 20 sites; both the exported guard in `platform/auth.ts` and the local copy in `submissions.ts`
- `requireUser` — proposal-routes, suggestion-inbox, remix, creator-studio, creator-profile-routes, editor-drafts
- `resolveEditable` / `resolveForSlug` / `notFoundIfDisabled` / `loadOwned` early-outs, which send and return `null` / `true` the same way
- `replyWithDraft` in `draft-preview-routes.ts`, whose `Promise<void>` was returned straight from the handler and resolved with `undefined` while its own sends were in flight

## Test

`submissions.test.ts`, in `describe('GET /api/me/quota')`: registers an `onSend` hook, injects an unauthenticated request, asserts 401 **and** exactly one send. The status code is correct in both the broken and fixed cases, so counting sends is the only assertion that catches this. Reverting any single guard makes the test fail and brings the unhandled rejection back.

## Verification

- `apps/api` suite: 4200 passed, **0 errors**. It previously reported 4 errors, all of this class — three `/api/submissions/:token/preview` cases and the suggestion `approve` route.
- `tsc --noEmit` clean, `npm run lint` clean.
- The comment-prose baseline for `submissions.test.ts` moves 4637 -> 4650 for the test's one-line comment (`npm run comment-prose -- <path> --write --force`, scoped to that file only).

## Reviewer notes

Behavior is unchanged for every caller: same status codes, same bodies. The diff is mechanical (`return;` -> `return reply;`), and the only non-mechanical edit is the `await replyWithDraft(...); return reply;` split in `draft-preview-routes.ts`.

### DIFF
--- a/apps/api/src/agent-surface/self-build-connect-routes.ts
+++ b/apps/api/src/agent-surface/self-build-connect-routes.ts
@@ -39,7 +39,7 @@ export async function registerSelfBuildConnectRoutes(
       if (!submissionTokenSecret) {
         return reply.status(503).send({ error: 'submissions are not configured' });
       }
-      if (!checkUserAccess(request, reply)) return;
+      if (!checkUserAccess(request, reply)) return reply;
       if (!store) {
         return reply.status(503).send({ error: 'submissions are not configured' });
       }
@@ -138,7 +138,7 @@ export async function registerSelfBuildConnectRoutes(
     '/api/submissions/:id/agent-key',
     { config: { rateLimit: { max: 60, timeWindow: '1 hour' } } },
     async (request, reply) => {
-      if (!checkUserAccess(request, reply)) return;
+      if (!checkUserAccess(request, reply)) return reply;
       return reply.status(410).send({
         error: 'per_game_keys_retired',
         reason: 'Reconnect this coding agent from Studio using OAuth or the creator-wide key.',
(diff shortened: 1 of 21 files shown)


gtanczyk:
@codex

PR #1315, opened 2026-09-13
Add canonical game access records and backfill migration

gtanczyk:
## Summary
Introduces a canonical game access record system to establish authoritative ownership of games, replacing reliance on the legacy derived rule (newest submission). Includes a backfill migration to populate records for existing games.

## Key Changes

- **New Game Access Record Model** (`game-access.ts`):
  - `GameAccessRecord` stores canonical ownership, editors, and members with revision tracking
  - Separate from submissions to enable independent authority management
  - Supports future collaboration features (GO-03)

- **Game Access Resolution** (`game-access-resolve.ts`):
  - `resolveGameAccess()` returns canonical records when available, falls back to derived rule
  - `classifyOwnerUid()` centralizes owner classification (human creator vs. platform-held)
  - `ownsGame()` and `sameOwner()` provide consistent ownership checks
  - `gameAccessMatchesDerived()` gates dry-run migration validation

- **Game Access Store Implementations**:
  - `InMemoryGameAccessStore` and `FirestoreGameAccessStore` with transactional safety
  - `ensureGameAccess()` creates records idempotently (rerun-safe)
  - `recordSettledOwner()` only updates pristine records (settlement-only writes)
  - `listGameAccessByMember()` supports shelf queries

- **Backfill Migration** (`game-access-backfill.ts`):
  - `runGameAccessBackfill()` creates records for games predating the system
  - Handles edge cases: abandoned submissions, bot/deleted owners, concurrent writes
  - Dry-run mode for validation before production execution
  - Reports divergences where legacy rule changed after initial record

- **Integration Points**:
  - `setSubmissionSlug()` now creates access records for uncontested slugs
  - `deleteAccountIdentity()` updates records via `withMemberErased()`
  - Admin endpoint `/api/admin/game-access-backfill` for migration execution
  - Both in-memory and Firestore stores extended with game access support

- **Comprehensive Test Coverage**:
  - `game-access-resolve.test.ts`: Core resolution logic, ownership classification, account erasure
  - `game-access-backfill.test.ts`: Migration scenarios, concurrency handling, dry-run validation

- Records are idempotent: re-settling a slug never overwrites authority
- Platform takes custody of bot-owned and deleted-account games
- Backfill re-reads submissions before writing to catch concurrent erasures
- Transactional writes prevent race conditions in Firestore
- Revision tracking enables future access control fencing

https://claude.ai/code/session_01RJgFo1bPZAk1R94pPDzPjM


apps/api/src/platform/game-access-backfill.ts
```diff
@@ -0,0 +1,97 @@
...
+      continue;
+    }
+
+    if (dryRun) {
+      countCreate(result, classifyOwnerUid(ownerUid));
+      continue;
+    }
+
+    // Report the record in force, not the intended one.
+    const inForce = await store.ensureGameAccess(slug, ownerUid, now().toISOString());
```
gtanczyk: [P1] Account erasure can still race the final read and recreate the erased owner

Reading submissions immediately before `ensureGameAccess` does not make those operations atomic. I reran the previous erasure-race reproduction: seed a legacy submission with no GameAccess, then run the real `deleteAccountIdentity(ownerUid, at)` immediately before the real `ensureGameAccess` call. Erasure sees no access record to scrub; the subsequent create persists the erased UID with revision 1. Result on FirestoreStore(fake): `created: 1, createdPlatform: 0, diverged: []`, and `resolveGameAccess` returns the erased user as a creator. The new test erases before `listSubmissionsBySlug` returns, so it does not cover this interval.

Make source-authority validation and access creation participate in the same concurrency protocol as erasure (transactional source reads / a shared generation or tombstone), retrying or quarantining changed sources. Re-reading earlier only moves the race window. Add the interleaving after the last read and before the create as a regression test.

apps/api/src/platform/slug-ownership.ts
```diff
@@ -24,7 +27,9 @@ export async function settleSlugClaim(
 ): Promise<string | null> {
   const holds = async (candidate: string): Promise<boolean> => {
     const holder = await store.getSubmissionBySlug(candidate);
-    return holder?.jobId === jobId;
+    if (holder?.jobId !== jobId) return false;
+    if (holder.ownerUid) await store.recordSettledOwner(candidate, holder.ownerUid, new Date().toISOString());
```
gtanczyk: [P1] A delayed settlement overwrites the newer winner

`getSubmissionBySlug` and `recordSettledOwner` are separate operations, and both the initial and successfully settled access records remain revision 1 with no editors. Deterministic reproduction on BOTH InMemoryStore and FirestoreStore(fake): Alice reads herself as holder and pauses immediately before `recordSettledOwner`; Bob creates the newer submission, claims the same slug, and completes settlement; Alice resumes. Both calls return `same-title`, legacy ownership says Bob, but canonical ownership is overwritten back to Alice. `isPristine` does not distinguish a tentative record from one already settled by the other job.

Serialize claim validation and canonical ownership publication against competing slug assignments, carrying the expected claimant/version into the conditional write. Do not return settlement success when that claimant has lost. Add the delayed-writer interleaving in addition to the sequential collision test.

gtanczyk (commented):
Follow-up verification at c47bdfee7dcdcffa895e230871e3ea2524aeaf9c. The original sequential collision, sequential account-erasure, private-draft enumeration, and competing-record reporting reproductions now pass. However, the three cases below remain reproducible. Independent full validation in an isolated checkout passed: npm install, type-check, lint, npm run test (including the previously failing API cases; 4 tests skipped), and build. The tracked checkout stayed clean. These findings come from separate deterministic integration probes and are not covered by the passing suite. No production migration was executed. This preparation is not ready to merge until the data-integrity and coverage cases below are addressed.

apps/api/src/catalog/admin-game-routes.ts
```diff
@@ -99,6 +100,36 @@ export async function registerAdminGameRoutes(app: FastifyInstance, options: Adm
...
+
+    const dryRun = request.query.dryRun === '1' || request.query.dryRun === 'true';
+
+    // Both lanes plus drafts: a games doc alone misses every unpublished project.
+    const [gameSlugs, submissionSlugs, publications] = await Promise.all([
+      store.listGameSlugs(),
+      store.listSubmissionSlugs(),
+      store.listPublications(),
+    ]);
+    const slugs = [...new Set([...gameSlugs, ...submissionSlugs, ...publications.map((entry) => entry.slug)])];
```
gtanczyk: [P2] The work list still omits repo-only catalog games

The added submission enumeration fixes private drafts, but this union never reads the repo catalog/snapshot. `listGameSlugs()` enumerates Firestore `games` documents/subcollections and `listPublications()` reads the publication field of that same collection; a repo-lane game with no submission, publication record, or community data is absent from all three lists.

I registered the actual catalog and admin routes with a synthetic GitHub catalog containing one published `repo-only` entry and an empty store: `GET /api/catalog` returns that game (200), while the dry-run returns `scanned: 0, quarantined: []` (200). Include the authoritative repo-catalog slugs (and surface catalog-read failure rather than declaring complete coverage). Verify the route, not only `listSubmissionSlugs`, with this case.

gtanczyk:
All three findings fixed in 85de586. (Per-thread replies kept failing with a 500 from the API, so they are collected here.)

**P1 — backfill vs. erasure after the final read.** You were right that re-reading only moved the window: the read and the create had to become one operation, not two closer together. `backfillGameAccess` now checks the account inside the same transaction as the write (Firestore: `tx.get(users/{uid})` before `tx.create`). When it cannot confirm the account it writes nothing and returns null, and the pass quarantines the slug rather than recording platform custody — an owner we cannot confirm is the case the plan says not to guess, and quarantine is also correct for a live creator whose user record is missing. The regression test interleaves the whole erasure in the interval you named, after the final `listSubmissionsBySlug` and before the create: `created: 0, createdPlatform: 0, quarantined: ['orbital-dogfight']`, nothing written.

**P1 — delayed settlement overwriting the winner.** The record now carries the claimant that settled it (`settledJobId`), and an earlier job never writes over a later one; `isPristine` alone could not tell a tentative record from a settled one. `settleSlugClaim` also stops reporting success when it lost: if the record in force names another job, `holds()` is false and the caller re-mints instead of being told the name is theirs. Two tests on both stores — the paused-writer interleaving, and the late loser's own settlement returning a different slug.

**P2 — repo-only catalog games.** The union read three Firestore lists and never the repo catalog. The route now includes `getCatalogEntries()` (the same source `/api/catalog` serves) and returns 503 `catalog_unavailable` when it cannot read it, rather than reporting coverage it cannot establish. Tested through the route as you asked: a synthetic catalog with one `repo-only` entry and an empty store now scans it and quarantines it; a throwing catalog gets 503; a private draft plus the repo entry scans 2.

Gate on this branch: type-check, lint, full `npm run test` and build all green, including a merge of `master` that had moved underneath.

Separately, the ops inventory now describes the `agent-keys.ts` failure the way you corrected it — the stale `gameAgentKeys.ownerUid` locks out the **new** owner via `SLUG_NOT_ON_ACCOUNT_REASON` after canonical slug authorization admits them, which is a different problem from stale in-flight capabilities — and the plan's status line says GO-01 is partial rather than implying only classification work remains.

---

gtanczyk:
Merged `master` (244aab7) at fd5095d. Worth flagging one thing, because it changes where authority gets written: master replaced write-then-check with an atomic slug claim (`claimAvailableSlug` / `claimSubmissionSlug`) for game creation and remix, while `slug-resolver` and the slug backfill still settle the old way.

An atomic claim is authority settled at the moment it wins, so the access record is now written there too, carrying the claiming job — the loser gets a different name and its own record, never the winner's. The `settleSlugClaim` hook and its `settledJobId` ordering stay for the paths master left on the old mechanism. New test covers the atomic path on both stores: winner holds `same-title` with `settledJobId: 1`, loser lands elsewhere with its own record.

Gate re-run on the merged head: type-check, lint, full `npm run test`, build — all green.

---

apps/api/src/store/firestore.ts
```diff
@@ -241,6 +246,7 @@ export class FirestoreStore extends SubmissionFacade implements Store {
       email ? this.db.collection('waitlist').where('email', '==', email).get() : Promise.resolve(null),
       this.db.collection('betaInvites').where('createdByUid', '==', uid).get(),
       this.db.collection('betaInvites').where('claimedUid', '==', uid).get(),
+      this.db.collection('gameAccess').where('memberUids', 'array-contains', uid).get(),
```
gtanczyk: [P1] Serialize account erasure with every GameAccess writer

The account-existence transaction fixes erasure completing before backfill creation, but the erased UID still survives two valid interleavings at fd5095df:

1. Firestore legacy game without gameAccess: deleteAccountIdentity reads this membership query (empty); before its first batch commits, runGameAccessBackfill commits successfully because users/uid still exists; the deletion batch then deletes the user and anonymizes the submission but never touches the newly created access document. Reproduced with fakeFirestore by wrapping the first deletion batch.commit to await the real backfill before invoking the original commit. Result: user absent, created=1, resolveGameAccess reports canonical creator g:alice, listGameAccessByMember returns one game. These are serial commits; Firestore transaction retries do not eliminate this ordering.
2. Fresh atomic claim, both stores: pause gameAccessStore.recordSettledOwner after claimSubmissionSlug has read the job, complete deleteAccountIdentity, then resume the original writer. claimSubmissionSlug returns true and recreates canonical owner g:alice with memberCount=1 while getUser returns null. This writer has no account check.

Use a shared erasure fence/tombstone checked transactionally by all canonical writers, established before erasure enumerates memberships, or an equivalent protocol that orders enumeration and creation. Add regressions for both schedules; checking account existence only inside backfill does not close account erasure. This already retains erased identity in the new collection and remains a blocker for canonical authorization cutover.

gtanczyk (commented):
Re-review of fd5095df9475dd51fa00e23dd59d580473b7fea9 in a detached throwaway clone; reviewed the actual PR diff against current master after the merge.

Previous findings:
- FIXED: stale settlement interleaving, independently reproduced on both stores. Alice gets same-title-2; Bob keeps same-title, with canonical and legacy ownership agreeing.
- FIXED: repo-only catalog coverage, exercised through the real catalog registration and backfill endpoint: scanned=1, quarantined=[repo-only].
- PARTIALLY FIXED: completed erasure before backfill creation now quarantines correctly. Account erasure still races membership creation in two other schedules; see the inline P1. This remains a blocker for relying on canonical authority.

Validation at this exact SHA:
- npm install, npm run type-check, npm run lint, npm run build: passed.
- Full npm run test: 7,517 passed, 2 failed, 4 skipped across workspaces; rule tests were skipped by the failed workspace step. Failures: improvement-admission (5s timeout), HeroPromptSection image-paste assertion.
- Follow-up after the full suite and build: the three GameAccess suites passed all 47 tests; improvement-admission still had one timeout (4/5 passed). HeroPromptSection passed 20/20 separately. npm run test:rules passed 76/76.
- GitHub checks currently show success, but I cannot attest a clean local full gate. I have not established that these test failures are caused by this PR.
- Deterministic external probes reproduced the remaining erasure finding, separately from the passing repository tests. No production writes or authorization cutover performed.

No approval/merge recommendation until the remaining erasure race is addressed.

gtanczyk:
Erasure race fixed in 81b0735, with the fence you asked for.

You were right that an account check inside the backfill does not close account erasure — it closed one schedule and left the two you named. `deleteAccountIdentity` now writes an erasure fence for the uid **before** it enumerates or scrubs anything, and every canonical writer (`ensureGameAccess`, `recordSettledOwner`, `backfillGameAccess`) reads that fence in the same transaction as its write. So a writer either commits before the fence, where the scrub that follows catches it, or reads the fence and refuses — the fence read is what orders them, not timing. All three return null when they refuse; settlement treats that as nothing to settle rather than as a lost name.

The membership scrub moved into the access slice and runs after the fence, so the two halves cannot be reordered by accident.

Both schedules are regressions on both stores, and both fail with the fence checks removed — which is how I found that the Firestore `ensureGameAccess` had not actually received its fence on the first pass. Your schedule 2 (claim reads the job, erasure completes, `recordSettledOwner` resumes) is reproduced end-to-end through `claimSubmissionSlug`.

Gate on 81b0735: type-check, lint, full `npm run test` (7,612 across workspaces), build — all green. On your two flaky failures: I do not see them here on this head, and I have no evidence either way about their cause, so I have not touched them.

Ops plan updated too: the erased-owner defect is no longer listed as simply fixed, and the standing rule is written down — a new canonical writer that does not read the fence reopens this.

---

gtanczyk:
@codex

apps/api/src/store/slices/game-access.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export class FirestoreGameAccessStore implements GameAccessStore {
+  constructor(private db: Firestore) {}
+
+  private doc(slug: string) {
+    return this.db.collection('gameAccess').doc(slug);
+  }
+
+  private erasureFence(uid: string) {
+    return this.db.collection('erasedAccounts').doc(uid);
```
gtanczyk, line 159: [P2] Scope the erasure fence to the erased account incarnation

This fence is permanent and keyed only by the provider UID, but successful Google sign-in recreates the user with the same `g:${sub}` (auth.ts:476, 521-528); neither sign-in nor upsertUser advances an account incarnation. An allowed returning user is therefore permanently refused by every canonical writer, even for entirely new games.

Reproduced on both stores at 81b07359: upsertUser(g:alice) → deleteAccountIdentity → upsertUser(g:alice) → createSubmission(2, g:alice, ...) → claimSubmissionSlug(2, brand-new-game, null). The claim returns true and the user exists, but getGameAccess returns null. For that game, dry-run reports created=1 / quarantined=[], while the real backfill reports created=0 / quarantined=[brand-new-game]. Thus the new game cannot join the canonical migration, and the preflight hides the permanent refusal.

Distinguish a new account incarnation/new work from stale writers belonging to the erased incarnation, and make dry-run reflect fence eligibility. Do not simply delete the tombstone on sign-in: that would let the delayed old writers through again. Add a delete → sign-in again → new-game regression alongside the two race tests.

gtanczyk (commented):
Re-review at 81b07359cf143a3d83ac0fbeb90f6c8d48277f63 in the detached review clone.

CLOSED: both erasure races from my previous review. The original probes now return no erased-UID membership: backfill after the deletion scan is quarantined, and delayed atomic-claim settlement cannot recreate the erased owner on either store. The fence read is in the write transaction for all three canonical writers. Previous stale-settlement and repo-only catalog probes also still pass.

NEW P2: the permanent provider-UID fence prevents a returning user from obtaining canonical records for entirely new games. Reproduced on both stores; dry-run and apply also disagree in this stable state. Details and a proposed regression are inline. Closing the original race does not close this account-lifecycle gap.

Exact-head validation:
- npm install, type-check, lint, build: passed.
- Full npm run test: 7,522 passed, 1 failed, 4 skipped across workspaces. The failure was a 5s timeout in agent-channel's creator-feedback test. Web tests, including the previously failing image-paste test, passed; improvement-admission also passed.
- After full tests and build finished, reran agent-channel plus all three GameAccess suites: 164/164 passed. Separately ran the rule tests skipped by the failed workspace command: 76/76 passed.
- GitHub CI is green. The isolated result does not turn the original full run into a green attestation; I have not established a PR regression behind that timeout.

No production writes, authorization cutover, or merge performed.

gtanczyk:
All four Codex findings fixed in ead5655, each verified before changing anything. (Per-thread replies still 500 from the API, so they are collected here.)

**P1 — a failed record write reported a lost claim.** Correct, and the consequence was the bad part: once `claimSubmissionSlug` commits the name is taken, so throwing afterwards told the caller the claim failed while the slug stayed occupied, and a retry minted a second job under another name. The record is derived from the claim and the backfill repairs it, so the claim path and `setSubmissionSlug` now treat that write as best-effort rather than part of the claim.

**P1 — erasure overwrote a concurrent newer owner.** Correct. Each record is now scrubbed in its own transaction that re-reads it and skips one the uid no longer belongs to, instead of writing a transformation of the query snapshot.

**P2 — batch over 500 writes.** Subsumed by the above: per-record transactions have no batch ceiling, so an account with more games than that no longer aborts the sweep after the fence is already installed.

**P2 — fence never lifted for a recreated uid.** Correct, and it would have been unrepairable: provider sign-in re-issues the same uid, so every future access record for that account would have been refused, backfill included. The fence now lifts when a genuinely new account takes the uid (`createdAt === lastLoginAt` on the upsert).

Tests on both stores: a claim survives an unavailable access store; a game whose owner changed mid-erasure keeps its new owner; a recreated account can own games again. The last two fail without their fixes.

One honest limit: the schedule I can reproduce for the erasure-overwrite fix is a settlement landing *before* the scrub, not one landing between the scrub's query and its write — that interval is inside the slice and I could not drive it from a test without test-only hooks in production code. The fix covers it; the test names only what it actually exercises.

Gate on ead5655: type-check, lint, full `npm run test`, build — all green.

---

gtanczyk:
Fixed in 212514b. You were right that deleting the tombstone was the wrong repair, and I had shipped exactly that in ead5655 — it re-admits the delayed pre-erasure writers the fence exists to stop.

**Incarnation scoping.** The fence is permanent again and carries the erasure time. Every canonical writer now passes the time its *work* began — the job's `createdAt` — and is refused only when that work predates the erasure. So a returning account's new games record normally, while a writer still carrying work from the erased incarnation is refused forever, tombstone intact. Regression on both stores: delete → sign in again → new game records; a writer holding pre-erasure work still gets `null`. It fails if the fence stops being time-scoped.

**Dry-run parity.** The dry run reads the same fence and account-existence conditions the write enforces, so it can no longer promise a record the apply refuses. Your `created=1` vs `quarantined=[brand-new-game]` case is a test.

Two more from the Copilot review in the same push, both real:

- **Revision reset.** Settlement replacing an owner rebuilt the record from scratch and reset `accessRevision` to 1 — discarding the fencing signal exactly when authority changed. It now carries forward and increments.
- **Backfilled records looked tentative.** No settled job meant any later claim could overwrite a migrated owner. Migration now records the job it derived the owner from, the same rule settlement follows.

Also: settlement compares the owner in force, not only the settled job id, so someone else's tentative record is no longer read as our own; and a test called `recordSettledOwner` with the wrong arity — nothing caught it because `apps/api/tsconfig.json` excludes tests from type-check, so Copilot's "type-check fails" was wrong about the mechanism but right about the call.

Not changed, and I'd rather say so than quietly leave it: **the resolver is still not wired into `creatorOwnsSlug` / `resolveOwnerOfRecord`.** That is deliberate — migration step 6 switches consumers after a reviewed dry run, and a resolver that disagreed with the legacy oracles mid-migration is the failure the staging exists to prevent. Copilot lists it as critical; it is the plan, not an oversight.

**Cached catalog** (Copilot, suppressed): fair. `getCatalogEntries()` serves its last good entries when a refresh fails, so a 200 can omit a repo-lane game published since that cache filled. The pass is rerunnable and the operator reads the report, but the route's guarantee is weaker than "fail closed" — I have not changed it, and it should be settled before the production dry run counts.

Gate on 212514b: type-check, lint, full `npm run test` (7,620), build — all green.

---

gtanczyk (commented):
Follow-up at `212514b128d7df095ff18a60dece7cbaaea34eff`: my last P2 (returning account permanently fenced out) is fixed.

Independent probes on both stores confirm:
- delete account → sign in/upsert the same UID → claim a new game creates canonical GameAccess;
- dry-run/apply agree for that returning account;
- all three writers still return null when replaying captured pre-erasure work after sign-in, with the tombstone retained;
- both original account-erasure interleavings remain fixed;
- the stale slug-settlement interleaving still moves the loser to same-title-2 and keeps canonical/legacy authority consistent;
- repo-only catalog coverage still scans and quarantines the entry.

No new blocking finding in this bounded follow-up. This closes my account-lifecycle finding, not the whole GO-01 migration. The acknowledged cached-catalog freshness limitation and previously documented migration prerequisites remain separate.


Completed validation on 212514b: install, type-check, lint and build passed. The full workspace test run finished naturally with 9 API failures and 46 web failures; no process was actually interrupted. I have not established that these failures are introduced by the ownership patch and do not attest a green full gate. The three GameAccess suites separately passed 61/61 with one worker, and the skipped rule-test stage passed 76/76 when run separately.

During verification the branch advanced to 27f48ac849c81fd028c639d0243d5099f4027125 (merge master). The access slice/record, submission facade, backfill and slug-settlement implementation are unchanged by that merge. On this newest HEAD I rebuilt the shared packages and reran all three GameAccess suites: 61/61 passed. I also repeated the returning-account and all-three-stale-writers probes on both stores: the fix still holds. The full gate results above belong to 212514b, not an attestation of 27f48ac8.

The final conclusion is unchanged: my last P2 is fixed. No merge, production write or authorization cutover performed.

gtanczyk:
Heads up before you append that verification: the head moved while you were running. The PR had gone `dirty` against master again, so 27f48ac merges `9342cbd` in. Only the module-size baseline conflicted — master had shrunk the two store files this branch grows, so I kept master's lower ceilings and re-raised from there rather than keeping this branch's looser ones. No source changes in that merge.

Full gate on 27f48ac here: type-check, lint, `npm run test` (7,681 across workspaces, 0 failed), build — all green. Worth pointing at 27f48ac rather than 212514b, since master moving under it is what produced the failures you saw outside the GameAccess probes on the older head.

Ops-side corrections you asked for are in `a978e4b`: the fence paragraph now describes time-scoping by the originating job's `createdAt`, and the cached-catalog freshness limit is in the blocker list rather than a PR comment.

---

PR #1317, opened 2026-09-13
fix(cli): recover orphaned checkouts on startup and connect

gtanczyk:
Starting `gamedevpl` in an orphaned checkout previously lost the local context, while `connect` stopped at “no owned game”. Both now offer **Recover and continue** for missing, canceled or archived games, stage the existing sources after confirmation, and open the same checkout without a second mode picker.

Cancellation keeps the checkout context and blocks accidental new-game creation from ordinary text. Authentication/network errors and occupied slugs never trigger recovery. Interrupted staging resumes using the existing recovery key before opening the workshop. Explicit recovery and noninteractive/manual connection remain available.

Validation:
- 470 CLI tests passed with two workers; startup/connect regression cases cover canceled, missing, archived-with-shelf-token and owned-live-publication states.
- Local type-check, lint and build passed. Real 60- and 90-column TUI exercised against a synthetic backend through confirmation, cancellation and workshop startup.
- Full GitHub CI passed on `5e8ac400c604fe3cb01ef4410a38012466f3552c`: lint, type-check, all tests, build, production image, contract, secret scanning and CodeQL.
- Codex reviewed the same commit and reported no major issues after both findings were fixed.
- Earlier unrestricted local runs hit timing-sensitive API/child-process test timeouts; these are not represented as a clean local full gate.


gtanczyk:
@codex review Please review the latest head, especially orphan checkout detection on startup/connect, cancellation preserving local context, pending staging retries, and preventing recovery on auth/network failures.

gtanczyk:
@codex review Fixed the archived shelf-token case: local startup and connect now check recovery lifecycle even after obtaining a shelf token. Regression fixtures keep archived games on the shelf, and tests cover startup, connect and lifecycle lookup failure. Please review the latest head.

gtanczyk:
@codex review Fixed the published-game regression. A valid owned shelf token is retained when recovery returns occupied; recovery is offered only for missing/canceled/archived states. Tests distinguish owned live publications (token + occupied), archived publications (token + archived), and unavailable slugs (no token + occupied), across startup and connect. Please review the latest head.

PR #1319, opened 2026-09-13
Give the framed-play gate a parent the CSP can actually match

gtanczyk:
## The deploy gate has been red since this test landed, and the app was never at fault

```
Refused to frame 'https://www.gamedev.pl/' because an ancestor violates the
following Content Security Policy directive: "frame-ancestors *". Note that
'*' matches only URLs with network schemes.
```

`page.setContent()` leaves the parent document at `about:blank`. `frame-ancestors *` matches **only network schemes**, so Chromium refused the frame before a single line of our code ran. No value of `frame-ancestors` short of removing the directive would have admitted that parent — the test could not pass as written.

## Proof it is the harness, not the app

I ran the gate locally against **production**, which carries neither of the two pull requests the failure was first attributed to. It failed there too. Swapping the parent for a real `https` document, everything passes against that same unchanged deployment:

```
.framed-play      => 1
.stage            => 0
a[target="_blank"] => 1
a[target="_top"]   => 1
```

So the CSP is correct, `isPlayPermalinkPath` is correct, and the interstitial renders.

## The change

Serve the host page by intercepting a reserved `.invalid` origin (RFC 2606 — it never leaves the browser) and `goto` it. The parent is then a real https document, which is what a third-party embed actually is, so the test now exercises the thing it claims to.

## Why this reached master green

`apps/e2e` deliberately has no `test` script — its own `package.json` explains that the root `--workspaces` fan-out would otherwise put every local run on the network. It runs in the deploy gate and nowhere else. A test that can never pass therefore merges cleanly and blocks the **next** deploy, which is how two unrelated pull requests ended up under suspicion first.

Not addressed here, but worth someone's judgement: nothing runs this suite before merge, so the same trap is still open.

## Also on the way past

The deploy that failed at 08:52 was a **different** failure — `expected 200, got 502: {"error":"refine_unavailable"}` — not this one, and not investigated here. Production is healthy and serving the last good revision; master is three merges ahead of it until a deploy gets through.


gtanczyk:
@codex

PR #1165, opened 2026-09-04
[parked] Layer-2 next-idea chips (NP-1) — waiting on the visual proposal layer

gtanczyk:
## Summary
Implements a new feature that generates and displays AI-powered "next idea" suggestions to game creators, helping them decide what to build next. Ideas are generated via Vertex AI, cached per delivered version, and presented as clickable chips that prefill the feedback composer.

## Key Changes

- **New `NextIdeaGenerator` interface and implementations**:
  - `VertexNextIdeaGenerator`: Calls Vertex AI with game spec, QA answers, and context to generate bilingual (EN/PL) improvement suggestions
  - `StubNextIdeaGenerator`: Test stub for mocking
  - Fails open (returns empty array) on any generation error, timeout, or schema validation failure
  - Respects configurable timeout (default 8s) and caps results to 3 ideas max

- **New API endpoint** `GET /api/submissions/:token/next-ideas`:
  - Returns cached ideas for a delivered/published submission
  - Supports `?regenerate=1` query param to force regeneration (subject to daily quota)
  - Enforces creator-only access for published games (like the improve route)
  - Rate-limited per IP (60 requests/min) and per user (configurable daily quota, default 10)
  - Returns empty array before any version is delivered

- **Frontend integration** in `SubmissionStatusView`:
  - Fetches ideas on mount and displays them as blue accent chips with sparkle icon
  - Clicking a chip prefills the feedback textarea with the idea's prompt (localized to creator's language)
  - "Show different ideas" ghost button triggers regeneration without sending feedback
  - Records studio steps: `idea_chip_shown`, `idea_chip_used`, `idea_chip_regenerated`

- **Data persistence**:
  - `SubmissionRecord` now tracks `nextIdeas` array and `nextIdeasVersion` (the version they were generated for)
  - New `setSubmissionNextIdeas()` store method to cache ideas
  - New `nextIdeas` usage counter in `UsageCounters` for quota tracking

- **Configuration**:
  - `nextIdeaGenerator` option in submission routes (defaults to `VertexNextIdeaGenerator`)
  - `dailyNextIdeasQuota` option (env: `DAILY_NEXT_IDEAS_QUOTA`, default 10)
  - `NEXT_IDEAS_TIMEOUT_MS` env var for generation timeout

## Notable Implementation Details

- Ideas are generated only after a version is delivered, preventing suggestions on incomplete work
- Bilingual generation (always both EN and PL) regardless of creator's UI language
[...]

PR #1321, opened 2026-09-13
Say when a round left concept directions waiting

gtanczyk:
## What

After #1202 and #1257, a round that passes its preview gate can leave an AI concept card in the creator's studio thread. A creator watching from the CLI had no way to know: after a green gate the CLI said `round finished — Studio is waiting (preview green)` whether an agent had just drawn two frames or nothing at all.

Nothing else told them either. Posting a card sends no notification, raises no badge and touches no unread counter — `appendProposalMessage` only writes the row. The card simply sat in Studio until the creator happened to open it.

## The fix is mostly deletion of an accident

The status payload **already carried the card**. `native-job-status.ts:92` puts it on `progress.revisions[].proposal`, so the CLI received it on every poll — but `RoundStatus` in `apps/cli/src/turn.ts` declared no `progress` field, so it was dropped at parse. This teaches the type about it and prints it.

After a green gate with a card waiting:

```
round finished — Studio has concept directions: "Night patrol" / "Crowded stands"
pick one in Studio: https://…/studio/squad-game
```

Without a card, the line is exactly what it was.

## What it deliberately does not do

**It does not offer the pick in the CLI.** Two reasons, and the first is the important one:

- Choosing a visual direction from two or three words, without seeing the frames, is precisely the "a chip is *I'm feeling lucky*" problem that parked NP-1 (gamedevpl/www.gamedev.pl-ops#152, and #1165, now closed). Reproducing it in the terminal would be the same mistake on a different surface.
- `/improve` accepts only `{ feedback, builder }`. A pick made here could carry the sentence but not the frame Studio attaches as a reference image — dropping the half the feature exists for.

So the CLI's job is to say a decision is waiting and route to where the pictures are.

## Two details that carry the change

**The fingerprint carries the card's version.** The dream job posts *after* the round is already a boundary, so the status itself does not change when the card lands. Without adding it to `statusFingerprint`, `shouldAnnounceStatus` never fires and the line never appears — correct parsing and still invisible. There is a test for exactly that transition.

**Labels go through `sanitizeEventPayload`.** They are agent-written text on its way into a terminal. The server strips markup; ANSI escapes are a different category, and the CLI already has this filter for `failure.reason`.

## Verification

- Four new tests in `live.test.ts`, each mutation-checked: dropping the sanitize, the fingerprint field, or the event-line branch each reds its own test and nothing else.
- `apps/cli`: **448 passing** (444 before this change). Three tests and five files fail on `Cannot find module 'ink'` — **identically on a clean master checkout**, verified by stashing; `ink` is simply absent from this container's `node_modules`.
- `npx eslint apps/cli`, module-size and comment-prose clean. `npm run cli:changelog` green — entry filed under `### Added`, so the next cut is a minor.

## Note on the branch

This branch previously held #1165 (NP-1 chips), which was closed as superseded earlier today. That commit is preserved at `archive/np1-chips-1165` before the branch was restarted from master, so the parked work stays recoverable as promised on that PR.

https://claude.ai/code/session_014vukePXoVjUyRMt4PThPzR

---
_Generated by [Claude Code](https://claude.ai/code/session_014vukePXoVjUyRMt4PThPzR)_


gtanczyk:
@codex review

---

gtanczyk:
**`Games-repo contract` was red, and the cause is not in this diff — it is the branch name.**

The check resolves which games-repo ref to compare against from the PR's own head branch (`ci.yml:91`):

```yaml
GAMES_PUBLISHED_REF: ${{ github.event_name == 'pull_request' && github.head_ref || 'main' }}
```

That is deliberate and good: a paired EditorKit change lands on the same branch name in both repos, and the check validates the pair rather than the stale `main`. The failure log names it plainly:

```
checking gamedevpl/www.gamedev.pl-games@claude/creator-onboarding-prompts-0iy9ps
  editor-contract mismatch … games-repo length 54739, website length 59762
  games-repo: …"export type CollectionItemSpec = TilemapItemSpe"…
  website:    …"export interface LayeredItemSpec {  widget: 'l"…
```

The games repo still had a branch of that exact name — left over from games#1398, which merged this morning — sitting at `4310d5ea`, from before the layered-items contract landed. So the check compared today's website contract against a three-day-old games snapshot. Nothing to do with a CLI status line.

**Evidence it is not this PR's:** CI run 4725 on master (`9342cbd`) is green on this same check, because a push event resolves the ref to `main`. Same code, same contract files, different ref.

**Fix, applied at the source rather than worked around here:** the stale games branch now points at games `main` (`7d21788`), so the check compares against the same snapshot master does. Its previous commit is preserved at `archive/hud-schema-1398` in the games repo, so games#1398's pre-squash history is not lost. Deleting the branch outright would have been tidier, but ref deletions do not survive this environment's git proxy — the owner can delete both once they are no longer useful.

Worth recording as a trap: **a merged-but-undeleted games branch silently shadows the contract check for any later website PR that happens to reuse the name.** It fails with a real-looking contract diff that has nothing to do with the PR under review.

Re-running the job now that the ref is corrected; `Lint, Test & Build` is still in flight on this run, which is why the first re-run attempt returned "workflow is already running".

---

gtanczyk:
@codex review

`9cc142d` — the P2 is real and fixed. I traced it before changing anything, and the mechanism is exactly as described:

- `FeedbackPanel` says it itself: *"Only fires when published; a draft revision stays on this thread."* A finished-green round is not published, so the send goes to `/feedback`, and `creator-feedback-handler.ts` appends a creator message to the **same** `jobId`.
- `native-job-status.ts` builds `progress.revisions` from `listCreatorMessages(record.jobId)`, so that reply lands in the same list, after the card.
- My loop scanned back for any row carrying a proposal and skipped straight over it.

One thing your description got the other way round, and it happens not to change the conclusion: `listCreatorMessages` sorts **oldest first** (`a.createdAt.localeCompare(b.createdAt)`, then `slice(-limit)`), so the backwards scan was finding the newest card, not an old one. The defect was skipping the creator's reply, not picking the wrong card.

**The fix:** the scan now stops at the creator's own row.

```ts
if (revision?.proposal?.options?.length) return revision.proposal;
// The creator's own rows carry no origin; theirs is the answer.
if (revision && !revision.origin) return null;
```

`revisionOriginOf` reports only `'agent'` and `'studio'`, and the field is omitted otherwise — so its absence is precisely "the creator wrote this".

**Why not simply take the newest row.** That was my first instinct and it is wrong in one case: a `studio_ack` or an agent relay landing after the card would hide a card nobody had answered. There is a test pinning that (`keeps naming a card when only the platform spoke after it`).

On your second suggestion — checking the proposal's version against the current delivery — I did not add it, because the supersede rule already covers the path that produces it: a round reopens because the creator sent something, and that send is the creator row the scan now stops at. If you can name a path where a round re-finishes with no creator row in between, that is a real gap and I will take the version check too.

**Verification.** Two new tests, both mutation-checked (removing the supersede line reds `stops naming a card the creator already answered` and nothing else). `apps/cli` 450 passing, up from 448. `eslint apps/cli`, comment-prose, module-size and `cli:changelog` clean.

---

gtanczyk:
@codex review

`9f1569d` — you named the path I asked for, so I took the version check as promised.

**Verified before implementing.** `OPERATOR_RETRY_STATES` does include `needs_changes` (submissions.ts:1675), and the handler calls `resumeBuild` with zero `appendCreatorMessage` calls in that block. So a retry can finish a fresh delivery in complete silence, and the supersede rule alone would keep naming the previous delivery's frames. Real gap.

**The comparison is sound, which I checked rather than assumed.** `progress.headSha` is `record.previewVersion ?? record.deliveredVersion` (`native-job-status.ts:55`), and the reconciler hands the dream job a `version` picked from that same pair (`job-reconciler.ts:285`). Same value space, so the check discriminates instead of always failing.

**One caveat I am not going to paper over.** They are the same pair but not the same expression: the reconciler takes whichever of the two carries *this round's* manifest generation, while `headSha` always prefers `previewVersion`. After a reopen the two can disagree for a window, and in that window a live card is hidden rather than named. I chose that direction deliberately — being silent about a real card costs a creator a suggestion they still see in Studio; naming a stale one sends them to frames from a delivery that no longer exists.

**A test of mine was lying, and your finding is what exposed it.** Two fixtures built their variant with `progress: { revisions: [...] }`, which replaces the whole object and so dropped `headSha`. Once the version check existed, `stops naming a card the creator already answered` went green because there was no delivery to match — not because of the supersede rule it claims to test. Both fixtures now spread `...carded.progress`, and with that fixed the supersede rule still reds under mutation on its own.

**Verification.** Both guards mutation-checked independently: removing the version check reds the two new tests, removing the supersede line reds its own and nothing else. 23 cases in `live.test.ts`; `apps/cli` 452 passing, up from 450. `eslint apps/cli`, comment-prose, module-size and `cli:changelog` clean.

Separately: `Games-repo contract` came back **green** on `9cc142d` after the stale games branch was repointed at `main`, confirming that failure was the branch-name shadowing and not this diff.

---

gtanczyk:
@codex review

`a23bf00`. You are right, and this one is squarely my process failing rather than a subtlety.

**The finding.** The `hostile` fixture replaced `carded.progress` instead of spreading it, so `headSha` was undefined, so `latestProposal` returned null before `proposalLabels` ran, so the escape assertion passed on a card that was never named. Vacuous.

**Two mistakes made it, not one.** In the commit that introduced the `headSha` requirement I fixed two fixtures with exactly this shape and did not audit the file for a third. Worse, I did not re-run the sanitize mutation afterwards — it had reddened *before* the version check existed, and I carried that result forward as if it still held. A mutation result is only true of the code it was run against.

That is the third time today a reviewer has found a test of mine that was green without covering its own property, and the second time I have written a paragraph about that exact hazard and then produced another instance of it.

**Fixed.** The fixture spreads, and the assertion no longer trusts `not.toContain` on its own:

```ts
// A hidden card would pass as stripped; assert it is named.
expect(formatStatusEvent(hostile)).toContain('red');
expect(formatStatusEvent(hostile)).not.toContain('');
```

A negative assertion cannot distinguish "the escape was stripped" from "there was no label at all" — that is what made the fixture bug invisible, so the test now pins the positive half too.

**Re-ran all three mutations against the current tree**, not against my memory of an earlier one:

| mutation | reds |
|---|---|
| drop `sanitizeEventPayload` on labels | `strips terminal escapes out of an agent-written label` |
| drop the version match | `drops a card whose delivery a retry replaced`, `says nothing when the status names no delivery to match against` |
| drop the supersede line | `stops naming a card the creator already answered` |

I also grepped the file for any remaining fixture that rebuilds `progress` without spreading it: none left.

One more thing in `a23bf00`: I ran `comment-prose` after staging rather than before, and pushed `fbd4b1c` with a 15-word comment over the ratchet. CI's lint job would have caught it. Fixed in the same breath as reporting it.

`apps/cli` 452 passing; `eslint apps/cli`, comment-prose, module-size and `cli:changelog` clean.

---

gtanczyk:
@codex review

`4840e40` — two of the three taken, the third declined with numbers rather than a shrug.

### Announce while the job still reads `building` — taken, and it was the big one

You are right, and this nearly made the whole feature dead on the main path. Verified in the repo's own test: *"A green preview never moves the job; proposals ride alongside"* (`submissions.test.ts:2095`), and `toSubmissionStatus` maps `submitted` → `building` (`job-state.ts:185`). The dream job runs from that same green-gate hook, so **the card is posted while the CLI reads `building`** — a state `isRoundBoundary` rejected and `formatStatusEvent` never named directions in.

A waiting card is now a boundary in its own right, and the line drops the "round finished" prefix when the round has not in fact finished. That 25 tests passed before this change and none failed is exactly the tell: the building path had no coverage at all.

### `origin: 'agent'` supersedes too — taken

Confirmed at both sites that set it, and the code says so itself: *"An agent calling `continue_draft` writes its own summary of a conversation held somewhere else"* (`submissions.ts:936`), plus `CreatorMessage`'s own note, *"'agent': relayed on the creator's behalf"*. It is the creator's ask in the agent's words, so it answers the card. The rule is now "only the studio's own voice is not an answer", which is the semantics rather than a special case.

### The account-wide mute — real, and I am not fixing it here

The gap is narrower than it reads: `appendProposalMessage` checks the mute **inside the posting transaction** and refuses with `ProposalRefusedBy = 'muted'`, so a muted creator never gets a card posted. What survives is only the window where someone mutes *after* a card already exists in that round.

Closing it from the CLI needs mute state the status payload does not carry — the web fetches it separately through `useProposalsMuted`. The obvious fix, adding it to the status response, means a per-request read of the user record on `/api/submissions/:token`. That endpoint runs every 3s per open tab and was measured at **69% of all Firestore reads** in a two-hour sample (see `c2fcaf0`), and `docs/firestore-read-cost.md` is explicit: read once per window, never once per request. I am not willing to trade a documented cost invariant for a preference edge this narrow inside a CLI status-line PR.

If you want it closed, the right shape is a windowed read like the one `c2fcaf0` added for play affinity — and that belongs in its own change, where the window and its staleness can be argued on their own terms. Happy to write it; I do not think it should ride here.

**Verification.** Two new tests, three mutations, three reds: drop the boundary → `names a card that lands while the job still reads as building`; restrict naming to `needs_changes` → same; treat only an absent origin as an answer → `treats an agent-relayed request as answering the card`. `apps/cli` 454 passing, up from 452. `eslint apps/cli`, comment-prose, module-size, `cli:changelog` clean.

---

PR #1329, opened 2026-09-13
Split game admission from account credentials, and read the catalog fresh

gtanczyk:
## Summary

Two of the three things GO-01 still owed after the [previous merge](https://github.com/gamedevpl/www.gamedev.pl/pull/1315) landed the canonical `GameAccess` foundation.

## Changes

**Split `store/slices/agent-keys.ts`.** It held two different subjects under one name: a uid-keyed creator credential, and a slug-keyed record that decides who may open a round on a game and holds the one-at-a-time lock while it opens. The second is game authority — a transfer has to move it, or the new owner is refused by a record still naming the old one — and it was invisible inside a module every reader took for credentials. It now lives in `store/slices/game-admission.ts`, with its own store interface (`GameAdmissionStore`); the credential half is unchanged.

**Read the catalog fresh in the backfill route.** The route trusted `getCatalogEntries()`, which serves its last successful snapshot when a refresh fails. So a 200 could report complete coverage while omitting a repo-lane game published since that cache filled. `CatalogRoutesHandle` gains `readCatalogFresh()`, and the admin route uses it instead — failing closed with 503, which is what it already promised. Two tests, both red without the change: the pass reads past a warm cache, and refuses once refreshes fail while `/api/catalog` still serves stale entries.

## Verification

- `npm run type-check && npm run lint && npm run test && npm run build` — all green (7,718 tests).
- Module-size baseline for `agent-keys.ts` lowered from 334 to 200 lines.

https://claude.ai/code/session_01RJgFo1bPZAk1R94pPDzPjM

---
_Generated by [Claude Code](https://claude.ai/code/session_01RJgFo1bPZAk1R94pPDzPjM)_


gtanczyk:
@codex review

---

gtanczyk:
Fixed in 749aaa7. Correct, and I confirmed it against the actual code before changing anything: `readCatalogFresh()` bypassed the app-level cache in `catalog-routes.ts` but still went through `loadCatalog()` → `snapshotReader.getCatalog()` → `getPointer()`, which has its own 60s cache and the same stale-on-error fallback the route-level fix was meant to remove. So in the production configuration (`snapshotReader` set), a fresh route read could still serve a pointer that predates a publish, or an indefinitely stale one during a pointer-read outage.

`GameSnapshotReader` gains `getCatalogFresh()`: it re-reads the pointer object directly — no cache read, no fallback to a previous value on failure — then reads that pointer's catalog. `readCatalogFresh()` now calls it when a snapshot reader is configured, so both cache layers are bypassed together, not just the outer one.

New test distinguishes the two paths with separate call counters on a fake `GameSnapshotReader`: warming the route's own cache via `/api/catalog` doesn't touch the fresh path, and the backfill call increments only the fresh counter. Reverting to the app-cache-only fix fails it — confirmed by reverting locally before writing this reply.

Gate on 749aaa7: type-check, lint, full `npm run test` (7,719), build — all green.

---

PR #1354, opened 2026-09-15
fix(editorkit): a game's editor controller must never strand the creator

gtanczyk:
Started as three ways the Studio editor could leave a creator with a sidebar, a toolbar and no board. Review turned it into the controller's whole lifecycle, because each fix exposed the next gap.

## The three it started with

**An empty view took the surface.** `EditorUiDocument` allows an array, `parseUiDocument([])` returns `[]`, and `[]` is truthy — so a controller posting `editor:ui` with an empty document set `status: 'ready'` with a non-null view, and `EditorSurface` then rendered nothing in place of the whole board. The 1500 ms watchdog cannot help: it only fires while the view is still `null`. The bridge refuses a document with no nodes and falls back.

**The fallback was not sticky.** `failController` and `useFallback` set `failed` and tell the game to stand down, but the very next `editor:ui` set the status back to `ready` and `EditorPanel` re-enabled the controller on that transition. A controller put into fallback *for breaking a content change* could take the surface back on its next render.

**The creator had no say.** `controllerDisabled` was only ever set by the shell, on failure, so a controller that was working but useless left no route back to the standard board. There is now a switch, and it belongs to the creator.

## What review added

**Standing down ends every command, not just views.** The first guard covered `editor:ui` only, so `editor:change` still set `pendingChange` — applied, pushed and autosaved. A game told to stop could keep editing the creator's draft invisibly. Now nothing from a stood-down frame is acted on except `editor:hello`, which is how a game asks for the draft so playtesting keeps working.

**A controller must not patch a draft the creator is editing by hand.** Choosing the standard editor only stopped rendering `EditorSurface`; patches still landed. They are refused now, and the game is told why. Its checks still count — the gate follows the controller's health, the draft follows the creator.

**The switch must not open the publish gate.** `allProblems` added the game's checks only while `controllerActive`, which the switch made false, so a creator with red checks could press "Use the standard editor" and Publish would light up. The gate keys off `controllerLive` — running, and not stood down — independent of which surface is drawn.

**A verdict about older content is no verdict about this one.** The bridge tracks `checksFresh`: a push clears it, an `editor:check` sets it. Publish is held while a live controller's last verdict is red *or* stale, and re-read after the save flush so a verdict landing mid-publish still counts.

**Silence does not hold Publish shut.** A push arms a three-second watchdog; no answer stands the controller down, which releases the gate *and* says why rather than leaving a dead button. The watchdog is cancelled when the listener that could clear it goes away, so leaving Edit mid-validation cannot fail a healthy controller later.

**A rebuild replaces the controller, not the draft.** Same-slug builds swap `GameFrame`'s `srcDoc` without remounting the hook, so the bridge resets controller state on a `documentKey` *and* on the frame's own `load` event — `shownHtml` is not the document the frame loads, since `GameFrame` rewrites it with `withGameLocale`, so a locale switch reloads it with the same build. Draft memory resets only on the slug, so a rebuild inside the autosave window does not answer the new frame with the stale saved draft.

**The switch is measurable.** `EDITOR_STEPS` gains `standard_surface_chosen` and `controller_surface_restored`; the client, the zod schema and `summarizeVisitFunnel` all read that one array, so the aggregate and admin read side come with it. Measurement-contract question 4: of the creators who reach a game-driven editor, how many keep it — the two new rungs against `opened` in `editing` on `GET /api/admin/telemetry/visits`.

## One thing added and then removed

A gate that blocked Publish when a definition declares `validate: true` but no verdict has arrived went in, and came back out two rounds later. It produced four findings of its own: off while connecting, the declaration unread when a push had cached the draft, an incomplete handshake blocking Publish forever, and a cached definition arming a watchdog for a replacement build declaring no validator.

The asymmetry decided it. That gate's failure mode is *the creator cannot publish at all*, in states that kept turning out to be wider than enumerated; the hole it closes is *a game's own check can be skipped in a narrow race*, with the server validating content against the definition either way. It belongs with the revision work below, where "this game owes a verdict and has not given one" becomes a state to track rather than infer.

## What this does not claim, and what is deliberately left

I went looking for these after a creator reported an empty board, and I cannot reach that game's definition from this environment, so **none of them is shown to be the cause**. They are defects on the path that produces that symptom, found by reading it.

**Deferred by the maintainer's decision:** `editor:check` still satisfies whichever push is outstanding rather than the one it was computed for, so two edits in flight can mark the second fresh using the first's verdict. The sound fix is a revision on `editor:content` echoed back on `editor:check` — a protocol change plus `shared/modules/editor.ts` in the games repo — shipping as its own two-repo change. Counting replies instead of matching them is not sound: `postView` posts a check on paths other than an inbound content apply, so the count runs ahead. The coarse gate here is strictly stronger than `master`, where a stale verdict was never checked at all.

## Tests

Every behaviour above has a case that fails against the code before it, verified by reverting the source in place rather than by stashing — twice redone, because a reversion that silently matched nothing, and one that approximated the old control flow instead of restoring it, each produced a test that proved nothing.

- `apps/web/src/editorBridge.controllerProtocol.test.tsx` — empty view, one-way fallback, every command refused after stand-down, hello still answered with the draft, stand-down surviving Edit reopening, a new document key and a frame reload each giving a fresh controller its chance, the listener attaching to a frame that arrives with the first build, the watchdog and its teardown on fake timers, verdict staleness, and a rebuild answered with the unsaved edit.
- `apps/web/src/surfaces/studio/EditorPanel.publishGate.test.tsx` — checks still gating after a surface switch, a patch refused on the standard surface, Publish held for a stale verdict, never held for a game with no checks, and a publish dropped when checks turn red mid-flush.
- `apps/web/src/surfaces/studio/EditorPanel.surfaceSwitch.test.tsx` — the switch is offered, brings the board back, is not overruled by a later view, and records both directions.

`npm run lint` clean, `type-check` clean, 688 tests pass.

## Ratchet note

The surface switch was extracted to `editorSurfaceChoice.ts` and `EditorSurfaceSwitch.tsx` to keep `EditorPanel.tsx` near where it started. Ceilings raised for `EditorPanel.tsx`, `editorBridge.ts` and the test files, all for code this PR adds.

https://claude.ai/code/session_01QKNijFdYrCvAxNjNc69DbV


gtanczyk:
@codex review

---

gtanczyk:
Both findings were right and are fixed in `8fde628c`. Replying here rather than in the threads — GitHub is rate-limiting the review-thread API for this account.

**P1, publish gate.** Confirmed: `allProblems` added the game's checks only while `controllerActive`, and the switch made that false, so a creator whose controller reported `checks.ok: false` could press "Use the standard editor" and Publish would light up while the controller stayed live and kept receiving content updates. The gate now keys off `controllerOffered` — running, and has not stood down — which is independent of which surface is drawn. `controllerOffered` absorbs `!controllerDisabled`, so the original intent still holds: a dead controller's stale checks cannot strand Publish. `EditorPanel.publishGate.test.tsx` now switches surfaces with the checks red and asserts Publish stays disabled and still names the game's checks; it fails against the previous gate.

**P2, instrumentation.** Also right, and the gap was exactly the one you named — a visit that loaded a controller and walked away from it was indistinguishable from ordinary abandonment. `EDITOR_STEPS` gains `standard_surface_chosen` and `controller_surface_restored` in `packages/contract/src/visit-vocab.ts`; the browser client, the zod schema and `summarizeVisitFunnel` all read that one array, so the aggregate and the admin read side come with it, and `VisitFunnelPanel` labels both rows. The question it answers is measurement-contract #4 — of the creators who reach a game-driven editor, how many keep it — queried as the two new rungs against `opened` in `editing` on `GET /api/admin/telemetry/visits`. Both directions are asserted in `EditorPanel.surfaceSwitch.test.tsx`.

`npm run lint` clean, `type-check` clean apart from two pre-existing unresolved imports on `master` (`@jsquash/webp`, `wasm-feature-detect` — both declared, just not installed in my container), 87 web test files / 659 tests and the api telemetry + contract suites pass.

---

gtanczyk:
@codex review

---

gtanczyk:
Right again, and it was the worst of the three — fixed in `4f838d2f`. (Still commenting here rather than in the thread; the review-thread API is still rate-limited for this account.)

The guard covered only `editor:ui`, so `editor:change` still set `pendingChange`, which `EditorPanel` applies, pushes live and autosaves. A game told to stop could keep editing the creator's draft invisibly while the standard editor was on screen. `editor:select`, `editor:ui-request` and `editor:check` were open the same way. The guard now sits above the whole handler: after a stand-down, nothing from the frame is acted on except `editor:hello`, which is how a game asks for the draft so playtesting keeps working — not a controller command, and `expectController` already refuses to re-arm from a failed status, so it cannot smuggle the surface back. There is a test for that exception too.

One thing I added beyond the finding, and want you to look at: making the fallback sticky had removed the only way back, since a later `editor:ui` used to re-arm the controller — the bug the previous commit fixed. So the stand-down now clears in the effect's own cleanup, which runs on leaving the playtest, switching game, or losing edit rights. All three are the creator's doing; nothing the frame can send resets it. If you'd rather the stand-down outlive a playtest restart, say so and I'll drop that half.

Four new tests in `editorBridge.controllerProtocol.test.tsx`; the two that matter fail against the previous bridge. Also folded in the ceiling problem CI caught separately (`c4cbef41`): the switch moved to `editorSurfaceChoice.ts` + `EditorSurfaceSwitch.tsx`, and `EditorPanel.tsx` is back to three lines over where it started instead of fourteen.

---

gtanczyk:
@codex review

---

gtanczyk:
Both addressed in `e591e905` — the P2 fully, the P1 in part, and I want to be straight about which part.

**P2, patches while the standard surface is chosen — fixed.** Choosing the standard editor only stopped rendering `EditorSurface`; the `pendingChange` effect still applied, pushed and autosaved every `editor:change`, so a delayed response to the creator's last custom-surface action could overwrite what they had just painted, with the controls that caused it off screen. Those patches are now refused and the game is told why. Checks still count, as you asked — the gate follows the controller's health, the draft follows the creator.

**P1, the mid-publish half — fixed.** `publishNow` tested `allProblems` once and then awaited the draft flush, so a check turning red during that await was ignored and the publish went out. Checks are re-read after the flush from a ref, so the newest verdict wins — the posture the not_sealed retry already had. Test drives it with a controlled `putEditorDraft` promise: the verdict lands while the save is in the air, and `publishEditorContent` is never called.

**P1, the revision-correlation half — not done, and deliberately.** Blocking Publish until the game has answered for exactly the content being published is the right fix for "edit on the standard board, click Publish before the re-check lands". It is also a protocol change: `editor:check` carries no revision today, and nothing in the contract obliges a controller to re-check on an inbound content push — so blocking until a matching answer arrives could strand Publish permanently for a game that only checks on its own tool actions. I am not making that call inside this PR. If you want it, it wants a revision on `editor:check` and a games-repo change to match, which I am happy to do as its own change.

Worth noting what this race is and is not: the server re-validates content against the definition on publish, so the schema-level rules hold regardless. What can slip through the window is a game's *own* check — the same class of rule that simply does not exist for a game without a controller.

Two tests, both failing against the previous commit. Lint clean, `type-check` clean, 664 tests pass.

---

gtanczyk:
@codex review

---

gtanczyk:
Both fixed in `6d2c0fab`, and the first of them shows I was wrong about the second.

**P2, the stand-down reset — dropped.** You are right that the iframe is not the effect. `StudioStage` stays mounted, so toggling the bridge off and on again cleared the flag while the same failed frame was still alive and `controllerStatus` was still `failed` — and that frame could then patch the draft without any new handshake, which is the hole the sticky fallback exists to close. I added that reset myself last round for recoverability and it was not worth the hole. Nothing resets it now except the slug change that already resets every other piece of controller state, which is where it belonged.

**P1, revision correlation — done, and my reason for deferring it was wrong.** I said blocking until a fresh check arrives could strand Publish forever because nothing obliges a controller to re-check on an inbound content push. I asserted that without checking. In the games repo, `applyContent` calls `activateController`, which calls `postView`, which posts `editor:check` whenever the controller declares `validate` — every push is answered. So the objection does not hold and there was no reason to leave the race open.

The bridge now tracks `checksFresh`: a content push clears it, an `editor:check` sets it, a slug change resets it. Publish is held while a live controller's last verdict is red *or* stale, under the existing "this game's own checks" label. The one shape that could still be stranded — a controller with no `validate`, which never sends a check — is exempt by construction: the gate engages only once a check has actually arrived, so a game with no checks is gated no more than before. There is a test for that exemption specifically.

Four tests across the two files; the stale-verdict one fails against the previous gate. Lint clean, `type-check` clean, 705 tests pass.

---

gtanczyk:
@codex review

---

gtanczyk:
Both fixed in `423dfedf`.

**P2, same-slug document swap.** Confirmed: `StudioStage` applies a new build by swapping `GameFrame`'s `srcDoc`, which does not remount the hook, so the new document's controller was being refused on behalf of the old one's failure. Rather than guess from messages — a failed frame can send `editor:hello` at will, which is why hello alone cannot re-arm anything — the bridge now takes a `documentKey` and resets its controller state on it. `StudioStage` passes `shownHtml`, the thing that actually changes when the frame loads something else. Closing and reopening Edit stays sticky, because that does not replace the document.

**P1, silence holding Publish shut.** Also right, and it was created by the very guard I added last round. A push now arms a three-second watchdog; if no `editor:check` answers, the controller stands down — which releases the gate *and* tells the creator why, rather than leaving them clicking a dead button. I picked standing down over quietly releasing the gate because a silent weakening is the failure mode this whole PR is about. The watchdog arms only once a check has actually been seen, so a controller without `validate` never trips it. `failController` now delegates to a hoisted `standDown` so the watchdog and the message handler engage the same fallback.

Two tests, both failing against the previous bridge; the watchdog one runs on fake timers. Lint clean, `type-check` clean, 738 tests pass.

---

gtanczyk:
@codex review

---

gtanczyk:
First one fixed in `5deffc9e`. Second one I am stopping on, and saying why.

**Draft lost across a same-slug rebuild — fixed.** My `documentKey` reset went too wide: it cleared `lastContentRef` along with the controller state, so a rebuild landing inside the autosave window made the replacement's hello answer with the last *saved* draft instead of the edit the creator had just made. The reset is now two effects — draft memory on the slug, controller session state on the document. Test pushes an unsaved edit, loads a new document, and asserts the hello is answered with the edit; it fails against the single reset.

**Correlating a check with the push it answers — not doing it here.** You are right about the hole. But the sound fix is a revision on `editor:content` echoed back on `editor:check`, which is a protocol change and a games-repo change to `shared/modules/editor.ts`, not a third approximation in the shell. Counting replies instead of matching them does not work either: `postView` posts a check on paths other than an inbound content apply, so the count would run ahead of the pushes and call a stale verdict fresh.

I also want to name the pattern rather than keep going. The last four rounds have each found a genuine consequence of the fix from the round before, and this PR has grown from three shell defects into most of a controller-lifecycle state machine. That is a signal about the shape, not just the code. I am putting the protocol change to the maintainer as its own two-repo piece of work rather than bolting another layer onto this one; the coarse gate here is strictly stronger than what is on `master` today, where a stale verdict was never checked at all.

---

gtanczyk:
@codex review

Head is now `5deffc9e`. The maintainer has decided the revision-correlation finding ships as its own two-repo change (a revision on `editor:content` echoed on `editor:check`, plus `shared/modules/editor.ts` in the games repo), so that one is deferred deliberately rather than missed — please flag anything else.

---

gtanczyk:
Both fixed in `21bc9ce7`.

**Watchdog outliving its listener.** Correct — `push` armed it, the message effect's cleanup did not cancel it, so leaving Edit with a validation outstanding left a timer nothing could answer and stood a healthy controller down three seconds later. Cancelled in the cleanup now.

**`shownHtml` is not the document.** Also correct, and I checked it: `GameFrame` rewrites it with `withGameLocale(..., i18n.language)`, so an EN/PL switch reloads the frame with the same build. The bridge now also listens for the frame's own `load` event and resets on that — the document actually changing, rather than an input to it. The build key stays as the cheap signal for a new build; either one resets the same state.

The test harness now uses a real detached iframe element with a stubbed `contentWindow`, so a load is dispatchable and none fires on its own. Both cases fail against the previous bridge.

Lint clean, `type-check` clean, 724 tests pass. The revision-correlation finding remains deliberately out of scope per the maintainer's decision — it ships as its own two-repo change.

@codex review

---

gtanczyk:
Both fixed in `d52a8056`.

**Old view watchdog surviving a document reset.** Correct — the reset cancelled the check watchdog but not `controllerTimerRef`, so a frame reloading while the previous controller was still `connecting` carried that timer into the new document and failed it on the old one's deadline. Both are cancelled now.

**Declared validator that never answers.** You are right, and this corrects something I claimed twice: I had exempted "no checks seen" *by construction*, on the reasoning that the shell cannot tell a controller with no validator from one whose verdict is missing. It can — the definition carries `validate: true`, gate-validated, and the shell already has it. The gate blocks when a validator is declared and no verdict has arrived, and the check watchdog now arms on the first view for such a game rather than only after a check has been seen, so silence stands the controller down instead of quietly enabling Publish. A game declaring no validator is unaffected, which is the case the exemption was actually protecting.

CI was 8/8 green on `21bc9ce7`; this head re-runs it. Lint clean, `type-check` clean, 688 tests pass.

One note on method, since it nearly cost a false claim: my first attempt to verify the watchdog fix by reverting it silently matched nothing, so the test passed and looked like it proved nothing. Redone with an assertion on the replacement, the test fails against the unfixed code as it should.

@codex review

---

gtanczyk:
Both fixed in `567a1b64`.

**Gate off while connecting.** Right — it hung off `controllerLive`, which needs `status === 'ready'`, so a `validate: true` game that had said hello but not yet drawn its view left Publish enabled for the whole watchdog window, which is precisely the window where no verdict can exist. The gate now keys off the controller being engaged at all, connecting included, and only lets go when it fails or stands down.

**Definition read only when the draft was not cached.** Also right. `validateDeclaredRef` was set inside the draft fetch, which a hello skips when an earlier push already cached content, so a declared validator could read as none and the watchdog never armed. The fetch is for the editor state now, and a hello always asks for it; only *using* the returned draft is skipped. If a view has already landed when the definition arrives, the watchdog arms there.

Lint clean, `type-check` clean, 690 tests pass.

Worth flagging on method again: verifying the second fix took two attempts, because my first reversion was an approximation of the old control flow rather than the old control flow, and passed. Restoring the original ordering exactly, it fails as it should. That is the second time in two rounds a sloppy reversion nearly let me claim a test proved something it did not.

@codex review

---

gtanczyk:
Both findings are correct, and I have taken them as a reason to remove the thing that produced them rather than patch it again. `bb5e8ff0` drops the declared-validator gate.

That gate is two rounds old and has now produced four findings of its own: off while connecting, declaration unread when a push had cached the draft, an incomplete handshake blocking Publish forever after leaving Edit, and a cached definition arming a watchdog for a replacement build that declares no validator. Each individual fix was small; the sequence is the signal.

The asymmetry is what decides it. This gate's failure mode is *the creator cannot publish at all*, in states I keep failing to enumerate — your last two findings are both that. The hole it closes is *a game's own check can be skipped in a narrow race*, with the server validating content against the definition either way. A gate that strands people needs far more certainty than one that occasionally misses.

So the verdict gate returns to what was stable across every round: it engages once a verdict has actually arrived, then holds Publish while that verdict is red or stale. `validateDeclaredRef` and the arm-on-view path go with it. `loadState` stays — reading the definition on every hello rather than only when the draft is uncached is a plain improvement, and nothing now depends on it for gating.

Everything else on the branch stands: empty view, one-way stand-down, the surface switch, refusing patches on the standard surface, checks surviving the switch, staleness on push, the watchdog after a seen check and its teardown, the document-generation reset, and the draft surviving a rebuild.

Where the declared-validator gate belongs is alongside the revision on `editor:check`. Once a verdict names the content it answers for, "this game owes a verdict and has not given one" is a state the shell can track instead of infer, and these edges stop being edges. That is the two-repo change the maintainer has already scheduled separately.

Lint clean, `type-check` clean, 687 tests pass.

@codex review

---

gtanczyk:
Fixed in `a30e00ac`. Right again: the effect depended on `frameRef` alone, which is a stable ref object, so it ran once — and `StudioStage` renders no iframe until it has html for it, so on that first pass `frameRef.current` was null and the listener never attached or retried. It now also depends on `documentKey`, which is exactly what changes when the first build arrives and the frame mounts.

Test: a bridge mounted with no frame, then given one along with its first build, still resets on a later load. It fails against the single-dependency effect.

Lint clean, `type-check` clean, 688 tests pass.

@codex review

---

PR #1363, opened 2026-09-15
Fix round takeover for the current game owner

gtanczyk:
A game owner could see `canTakeOver: true` but repeatedly receive `round_changed` from `/push --takeover`: the route resolves canonical game ownership while the atomic store mutation still compared the round's historical `ownerUid`.

Authorize takeover using GameAccess inside the same Firestore transaction as the round update. Keep historical attribution intact, preserve generation/lifecycle guards, and use the existing newest-live-round ownership rule for legacy games without GameAccess. The in-memory store follows the same rules.

Validation: type-check, lint, build, and 19 focused regression tests passed. Full workspace tests with local HTTP ports allowed: API 5132 passed / 2 failed; CLI 511, web 2250, world 26, contract 119, zone-core 94 passed (4 skipped). Both API failures in `game-transfer-flow.test.ts` reproduce identically on untouched master `d2caa43b2` (8 passed / 2 failed in the baseline comparison). The sandbox-only loopback failures disappear outside the sandbox. Coverage includes current/former owners, platform custody, stale generations, canonical authority transaction reads, legacy fallback, and the authenticated HTTP route.

Production observation confirmed repeated takeover 409s, but did not expose the underlying owner fields; this PR fixes the reproduced authorization inconsistency without claiming the live game's recovery has been verified.


gtanczyk:
@codex review

PR #1376, opened 2026-09-16
Share EditorKit validation and correlate editor:check with a revision

gtanczyk:
Follow-up to #1358 (`67b55109`). That PR saved the draft, and on the way through review it kept finding rules that lived in the server validator but not in Studio's hand-written copy (and the other way around). This PR puts one validator in `@gamedevpl/contract` and closes the rest of the list that was parked there.

Paired games-repo PR: https://github.com/gamedevpl/www.gamedev.pl-games/pull/1602

## 1. Shared validator (priority)

`validateEditorContent` and the types it operates on now live in `packages/contract` (`editor-kit.ts`, `editor-validate.ts`, `editor-validate-reach.ts`). `apps/api/src/creation/editor-contract.ts` keeps parse/generate (they need `Buffer`) and re-exports the rest. `apps/web/src/editorContentTools.ts` calls the real validator instead of a parallel implementation. Omitted empty `properties`/`constraints` on a layer spec are filled before L4 so Publish does not throw (`EditorLayerSpecDefaults`).

L4 still decides Publish. Draft writes still use `draftRules` in `editor-draft-shape.ts` — painting can save a half-finished map.

Check 31 fingerprints parse/generate/validate **and their helpers, limits, and screaming-snake contract constants**, concatenating the package validate + kit sources. That includes `EDITOR_CONTENT_FILE` / `EDITOR_FILE` / spec keys (interpolated into diagnostics), not only `MAX_*` and the key patterns. Function extraction skips object return types / param object types so `parseEditorDefinition` and `unreachable` contribute their bodies, not the `{ … }` in the signature. In-body `//` comments are stripped so the games-repo copy can keep prose the website contract files cannot. Live fingerprint matches `tools/lib/editor-contract.ts` (51599 bytes).

## 2. `revision` on `editor:content` / `editor:check`

The shell stamps a monotonic revision on every content push **and** on the hello draft. Studio ignores a check whose revision is not the document on screen. Older games that omit revision still work.

Did **not** restore the API game-authored `definition.validate` gate that #1358 removed. That gate produced false alarms because of this race; the race is now a client protocol. `draftRules` is unchanged.

## 3. Publish checks property/param *values*

Comes for free with (1). `documentProblems` is `validateEditorContent`, so `min`/`max`, enum membership, and text length match the server. `allProblems` now includes params. Param controls validate the **stored** value (including `null`); the default is only used to render the slider.

## 4. Malformed item in render

A controller replacement (`applyEditorPatch` with `{ content }`) can skip `mergeDraft` while the game UI is showing. Switching to the standard editor then renders `properties: null`. Optional chaining on the item list and the property sheet is covered by `EditorPanel.malformedItem.test.tsx`, which throws without the guards.

## 5. Keep out-of-spec values

`fillDeclaredValues` no longer resets a present value that the definition tightened. Publish names the field and the reason; the property/param control gets `is-invalid`. An enum that the game dropped stays as an extra `<option>` so the creator can see what was saved.

## 6. Banner overlay

`.editor-banner-slot` is `position: absolute` on `.editor-panel` (`position: relative`), so a legal error/fallback banner no longer shoves `.editor-body` under the cursor.

## Tests (fail-first)

Each behaviour change was run against the pre-fix code:

- Shared Publish value gate: stub `documentProblems` → `gateBounds` / `publishValues` fail; restore → pass.
- Stale `editor:check`: drop the revision early-return → revision test fails (`checksFresh` stays true); restore → pass.
- Keep out-of-spec values: restore `fitsSpec` → defaults tests fail (speed 1 becomes 5); restore keep → pass.
- `properties: null` render: drop `?.` → `Cannot read properties of null (reading 'name')`; restore → pass.
- Lockstep disk path: three levels below root → ENOENT on `apps/packages/...`; four levels → fingerprint matches games `main`.
- Lockstep helpers/limits: three wrappers only → helper/limit test fails; expanded fingerprint → pass.
- Parse-body extraction: object return type captured the `{ definition… }` signature, not `return "body"`; skip the type → pass. Comments inside `unreachable` no longer drift Check 31.
- Null param: `?? default` left the slider valid; validate stored → `.is-invalid` and Publish disabled.
- Hello draft: first `editor:content` had no revision; stamp 1 so a later push cannot accept that check.
- Codex P2 (`EDITOR_CONTENT_FILE`): allowlist skipped it, so a filename change left the fingerprint unchanged; screaming-snake const regex → fingerprint moves, still equal to the games copy.

## Out of scope

No copy from `www.gamedev.pl-ops`. Merge only after Codex LGTM on this head SHA, green CI, no conflicts, squash with `expectedHeadSha`. Copilot nits not taken: `aria-invalid`, unused `pathMessages` localization (shared validator strings), duplicate checks-elsewhere summary, advancing revision on every iframe reload (first-load race).

<div><a href="https://cursor.com/agents/bc-d2e4590e-d246-4d6e-8d7b-8a0ce9ff4298?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d2e4590e-d246-4d6e-8d7b-8a0ce9ff4298&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


gtanczyk:
@codex

gtanczyk:
@codex

gtanczyk:
@codex

PR #1378, opened 2026-09-16
refactor(cli): extract shared session controller

gtanczyk:
## Problem and change

Interactive session state currently belongs to the terminal UI. Extract it into a renderer-independent controller while retaining the terminal API, queue and local history behavior. Add an in-process command adapter so another client can submit input, queue a follow-up or request Stop against the same session.

Commands use prompt/task generations and bounded receipts: retries do not enqueue twice, stale replies cannot answer a later question, and Stop cannot cancel a different task or quit an idle terminal. Input preserves an unfinished terminal draft; message input cannot invoke terminal slash commands. Add regression coverage and an internal changelog entry. npm install also synchronizes the existing CLI 0.16.0 workspace version in the lockfile.

## Scope

This is the session foundation, not the browser interface. No HTTP listener, background controller process, browser workbench or durable receipt journal is added. The command adapter is typed internal code; a future transport needs envelope validation, authentication and binding to a unique controller instance. Local receipts do not establish remote API idempotency.

## Validation

Final head: `d83fe36eafc052011e994ddc2c9c8855d8b09150`.

- Full repository type-check, lint and build passed.
- Focused controller, terminal and history tests: 28 passed. Regression coverage includes reused IDs with empty/oversized/control/slash text and invalid generations: existing receipts return conflict before payload validation; unused IDs still return invalid.
- Final workspace test run: contract 119 passed; zone-core 94 passed / 4 skipped; API 5280 passed / 2 failed; CLI 539 passed; web 2351 passed; world 26 passed.
- ESLint rule tests independently passed: 76 tests.
- The full test gate is not green. Both API failures in `game-transfer-flow.test.ts` also reproduce on a clean archive of base `d98e233eb`: `cannot start work through a concurrent accept: the lock holds both ways` and `runs in reverse: the recipient can hand the game back`. The baseline reproduction has the same two failures and 11 passing tests. No transfer code is changed here.
- Earlier checks skipped while this PR was a draft are not validation evidence; use checks attached to the current head.

## Instrumentation

No new user-facing entry flow is exposed. Existing CLI events remain at execution call sites, preserving creator funnel/return measurements (questions 4–5); no additional prompts, paths or per-game identifiers enter analytics.


gtanczyk:
@codex

PR #1379, opened 2026-09-16
Recover stale transfer clients and derive assessment attribution

gtanczyk:
Old transfer clients omit invitation identity and receive `stale_client` without a usable recovery path. The service worker now refreshes only the requesting window after that explicit refusal, preserving the response and never replaying the action. It inspects the response before the browser consumes its stream.

Transfer dialogs also disclose that earlier project prompts and conversations are shared, while account-private conversations and credentials are excluded, in English and Polish.

Assessment writes now resolve creator attribution from server-side game access and catalog records instead of persisting the caller's `creatorHandle`. Dedicated resume coverage verifies that a revoked undelivered round gets a new generation and current epoch, while an unchanged round retains its generation and quota is not spent again.

Validation:
- Full local gate passes: type-check, post-commit lint, tests (8,494 passed, 4 skipped), production build. Includes 5,289 API tests and 2,360 web tests.
- Transfer-flow tests explicitly stub unrelated model decisions so network-enabled runs exercise build admission deterministically.
- Mutation checks fail with attribution, revoked-resume and worker fixes removed. A separate mutation catches response-stream consumption before inspection.
- Real-browser local fixture: submitting window navigates once, unrelated window stays open, exactly one POST reaches the server.
- Codex review found no issues on final head `a8e2d2292c7a46b5e667810d82a583594f7ba75b`. GitHub CI remains the final merge gate. No production mutations were performed.


gtanczyk:
@codex review

gtanczyk:
Handled the catalog-outage finding in 2003a4efe: attribution is best effort and never falls back to client data. Added an HTTP outage regression; five attribution cases pass. Post-commit lint and type-check pass. @codex review

gtanczyk:
@codex review

Current head a8e2d2292 adds transfer history disclosure and makes transfer-flow tests independent of live chat routing. Post-commit lint and 22 transfer UI tests pass. Earlier attribution-outage finding is fixed; the previous review completed on 2003a4efe6.

PR #1388, opened 2026-09-16
feat(cli): complete browser-first Play workbench

gtanczyk:
## What changes

Creators can build and edit from the local Play browser without switching back to the terminal. This PR delivers the six connected workbench capabilities together:

- **State-preserving reload:** keep the running iframe while preparing and restoring a candidate. Reject incompatible state without replacing the current game; offer explicit restart. Auto mode requires validation and a safe-point capability.
- **Evidence:** upload/paste images and assets, capture screenshots, attach recent canvas recording segments and diagnostic traces. Preserve drafts and pending command IDs across browser refreshes; retain local evidence when intake rewrites a task.
- **Phone testing:** explicit private-LAN pairing with QR, a separate scoped player, 30-minute expiry, revoke, independent build selection, and screenshot/trace reports returned to desktop. No editor commands on the phone listener.
- **CLI/platform operations:** fixed domain actions for checkout/connect, agent settings, verification, checkpoints, sync/delivery/publication, status, sharing and round cancellation. Existing domain confirmations and platform authorization remain in force. Shared checkout writer locks prevent concurrent CLI writers.
- **Create → Play:** `gamedevpl create --play [idea]` and `gamedevpl play --edit` start a detached browser controller. Repeated launch reconnects; successful creation is journaled before checkout setup. Unknown remote outcomes block mutation retries instead of creating another game.
- **Theater:** reuse Studio's embedding bridge and supported shell-chrome removal, with editing controls over the full viewport.

Implementation and limits: `docs/play-workbench.md`. The existing iframe sandbox remains `allow-scripts allow-pointer-lock`, without `allow-same-origin`.

## Validation

- `npm install`, repository type-check, lint and production build passed.
- Full repository test suite passed during integration (8,636 tests); after subsequent CLI refinements, the complete CLI suite passed serially: **578 tests / 91 files**. Earlier parallel runs had timing failures; their focused rerun passed without increasing test timeouts.
- After the final phone header/source-fencing refinements, targeted tests passed: **16 tests / 4 files**, plus CLI type-check and changed-source ESLint.
- Browser fixture: compatible update preserved score 8; incompatible restore retained the original iframe/build and persistent failure notice. Screenshot, diagnostic attachment and a decodable VP8 gameplay clip verified.
- Browser fixture: Send remains visible with tools open at 390×420; Polish draft survives refresh.
- Fake platform API: detached launcher exited while Play stayed available; explicit builder selection created exactly one game, preview appeared in the same workbench, and refresh did not create another game. No paid agent or real platform mutation was used.
- Phone protocol tests bind **loopback only** and exercise authorization, report idempotency, refusal of editor commands and revocation.

## Review / release limits

This is a draft pending actual LAN/device acceptance. Enabling the synthetic LAN fixture was blocked by the local approval review, so no physical-phone/network result is claimed.

State preservation depends on the game's snapshot/restore contract. Video is the latest completed eight-second segment of the largest canvas, without audio; traces are diagnostics, not deterministic replays. Local attachments require a local agent. Vendor flows requiring an interactive terminal remain explicit handoffs. Custom game HTML outside the supported shell contract may need source changes.

The journal intentionally does not claim server-side idempotency or automatic unknown-outcome reconciliation. It lives in temporary storage, not a reboot-safe supervisor. Crashed writer locks require checking that the child agent has exited. Merge does not itself publish a CLI release; the normal release PR is the publication boundary.


gtanczyk:
@codex

gtanczyk:
@codex

apps/cli/src/session-browser-script.ts
```diff
@@ -1,10 +1,15 @@
...
 export const SESSION_BROWSER_SCRIPT = String.raw`
 const el = id => document.getElementById(id);
-const panel = el('panel'), frame = el('game'), draft = el('prompt');
+const panel = el('panel'), draft = el('prompt');
+let frame=el('game');
 let token = location.hash.slice(1) || sessionStorage.getItem('session-token') || '';
 if (/^[a-f0-9]{64}$/.test(token)) sessionStorage.setItem('session-token', token);
 history.replaceState(null, '', '/');
 let state, online = false, pending, sending = false, stopping = -1;
+try {draft.value=sessionStorage.getItem('play-draft')||'';const saved=sessionStorage.getItem('play-pending');if(saved&&saved.length<20000)pending=JSON.parse(saved);}catch{}
```
gtanczyk, line 11: 🔴 A `play-pending` envelope restored from `sessionStorage` still carries the `sessionId` of the previous server session. After the CLI is relaunched, `/commands` answers 409 for it, the client treats that as a lost receipt and never clears `pending`, so Send/Run stay disabled until the user wipes `sessionStorage` by hand.

Suggest dropping the restored envelope when its `sessionId` differs from the one `/state` reports (or clearing `pending` on a 409 that names a stale session).

---

apps/cli/src/workbench-launch.ts
```diff
@@ -0,0 +1,245 @@
...
+      onReady: (url) => {
+        journal.url = url;
+        save();
+      },
+      onCheckpoint: (state) => {
+        Object.assign(journal, state);
+        save();
+      },
+    });
+    journal.ended = true;
```
gtanczyk, line 252: 🔴 A normal `/quit` marks the journal `ended`, and the next launch (line 93) then builds a fresh journal, discarding `pending` together with the token/slug. Sequence: create times out (unknown outcome) → user ends the session → relaunches → the new journal has no `pending`, so the create runs again and can produce a duplicate game. That is exactly the case the journal is meant to fence.

Suggest keeping `pending` (and the identifiers needed to reconcile it) across `ended`, or refusing to mark the journal `ended` while a `pending` mutation is unresolved.

---

apps/cli/src/session-browser-server.ts
```diff
@@ -75,10 +128,15 @@ export async function startSessionBrowser(session: SessionController) {
...
         reply(200, {
           version: 1,
+          detached: options.detached === true,
           sessionId,
           sequence,
           sourceId,
           hasPreview: Boolean(preview),
+          actions: Object.keys(WORKBENCH_ACTIONS),
+          addresses: lanAddresses(),
+          phone: phone && { url: phone.url, expiresAt: phone.expiresAt, qr: phone.qr },
```
gtanczyk, line 139: 🔴 The 30-minute phone expiry closes the listener server but never clears the `phone` reference, so `/state` keeps advertising the URL/QR after the port is dead, and "Revoke" becomes a no-op on an already-closed listener.

Suggest setting `phone = undefined` in the expiry timer (same path revoke uses) so the UI reflects the real state.

---

apps/cli/src/workbench-lock.ts
```diff
@@ -0,0 +1,42 @@
...
+  } catch (error) {
+    rmSync(lock, { recursive: true, force: true });
+    throw error;
+  }
+  const roots = new Set(transaction.getStore() ?? []);
+  roots.add(canonical);
+  try {
+    return await transaction.run(roots, run);
+  } finally {
+    const owner = JSON.parse(readFileSync(join(lock, 'owner.json'), 'utf8')) as { instance: string };
```
gtanczyk, line 45: 🔴 `readFileSync(join(lock, 'owner.json'))` in `finally` is unguarded. If the lock dir was removed mid-operation (the docs' crash-recovery text tells users to delete it), this throws ENOENT and replaces the real result or error of `withCheckoutWriter` with an unrelated one.

Suggest wrapping the read in try/catch and skipping the `rmSync` when `owner.json` is missing.

---

apps/cli/src/workbench-launch.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+  if (existing && !existing.ended && running(existing.pid))
+    throw Error(
+      'Play process is still alive but not responding. Reconnect after it recovers; a second writer was not started.',
+    );
+  const releaseStartup = acquireStartupLock(lock, existing?.pid);
+  try {
+    const journal: PlayJournal =
+      existing && !existing.ended
+        ? { ...existing, pid: undefined, url: undefined, initial: undefined }
```
gtanczyk: 🔴 Relaunching `gamedevpl create --play "<idea>"` over a crashed (non-`ended`) journal reuses it with `initial: undefined`, so the idea passed on the command line is silently dropped.

Suggest `initial: input.idea ?? existing.initial` (or at least a warning that the new idea was ignored because a previous session is being resumed).

---

gtanczyk (commented):
Review of the CLI workbench changes. Five correctness findings, all in `apps/cli/src`, each posted inline:

1. `session-browser-script.ts` — restored `play-pending` keeps a stale `sessionId`, leaving Send/Run permanently disabled after relaunch.
2. `workbench-launch.ts` (quit path) — `/quit` marks the journal `ended`, and the next launch drops `pending`, allowing a duplicate create after a timed-out one.
3. `session-browser-server.ts` — phone expiry closes the listener but leaves `phone` set, so `/state` and Revoke go stale.
4. `workbench-lock.ts` — unguarded `owner.json` read in `finally` masks the real result when the lock dir was removed.
5. `workbench-launch.ts` (journal reuse) — relaunch over a crashed journal silently drops the `--play "<idea>"` argument.

---

PR #1389, opened 2026-09-16
feat(mcp): opt-in ownership scope and transfer proposals

gtanczyk:
Adds an opt-in OAuth `ownership` scope and three MCP tools so a signed-in creator can read current-member access and prepare a Studio transfer proposal. The agent cannot accept, reject, cancel, or finalize a transfer.

Rebased onto `master` at `47bbc1d82` (#1396). The temporary CLI comment-prose and changelog Internal commits are gone: master already satisfied that gate.

## What this ships

- Separate OAuth scope `ownership`, advertised but not granted unless the creator consents. Existing `mcp` / `creator` grants are not upgraded.
- Tools: `get_game_access`, `propose_game_transfer`, `get_game_transfer_proposal_receipt`. Round keys, session keys, upload keys, and creator API keys cannot call them.
- Reads are canonical-member-only, return opaque member keys (no raw UIDs), and refuse missing and inaccessible games the same way. Reads do not backfill access records.
- A proposal is an expiring review item, not an invitation. The tool returns a same-origin Studio URL; the signed-in owner picks the recipient and confirms current terms in the UI. Confirm reuses the existing transfer invitation protocol (idle-only accept, moderation, eligibility).
- Idempotency is bound to account + request hash. Same key + different payload is a conflict. Expired proposals keep a tombstone so a late retry cannot recreate them.

## Review follow-up

Addresses the Claude Code comments on `f59bac60`:

- CLI `/device` shows the same ownership can/cannot copy as AS consent when the pending request includes `ownership`, and will not approve that scope until the page has disclosed it.
- Firestore erase (and confirm) only deletes `gameTransferProposalBySlug/{slug}` when it still points at the proposal being closed.
- A leftover open proposal from a previous owner is treated as stale: it is invalidated, then the current owner may propose.
- `InMemoryStore.getUser` is a prototype method again so test subclasses can override it.

Propose now re-reads canonical `GameAccess` inside the write: a delayed previous-owner call returns `stale_owner` and cannot invalidate the current owner's proposal.

Proposal confirm mapped invitation-create `busy` to the build-running copy. Create-time busy is a pending invitation, so the page now recovers the open offer or shows `alreadyOut`, matching StudioTransferPanel.

A confirmed proposal is no longer reported as `invalidated`. Reloading the review URL after a lost confirm response recovers the pending invitation, or says it was already sent.

## CI follow-up

`firestore.ts` stays under the 1669-line module-size ceiling. `in-memory.ts` stays at 1652.

## Tests

- Store slice on both InMemory and Firestore (fake), including leftover-owner invalidation, erase pointer guards, previous-owner propose races, and confirmed vs invalidated receipts.
- MCP refusals: missing/wrong scope, disallowed key types, owner/editor/stranger/removed member/reinvite/recreated account, erased-owner propose, and a new key after a membership bump.
- Confirm route: 401 without a session, 404 symmetry, CSRF Origin check, stale owner after membership change, blocked recipient, GET `invalidated` vs `expired` vs `confirmed`, and cancel-on-failed-confirm.
- Studio confirm page: signed-in fetch with credentials, expired state hides the form, invitation-create `busy` recovers the existing offer or `alreadyOut` (never the build-round copy), confirmed reload recovers the pending invitation, and this surface imports `studio-panel.css`.
- Device flow: GET with `user_code` discloses ownership permissions; approve without `disclosed_scope` stays pending.
- `tools/list` stays under the existing 120 KB schema ceiling.

## Docs

- `docs/README.md` orchestration row and `docs/architecture.md` note that the agent can prepare, not complete, a transfer.
- `listings/mcp/README.md` advertises the three tools.
- Privacy policy section 5 discloses the extra member metadata returned when the ownership permission is granted.

This PR does not implement equal co-ownership, production backfill, membership changes, or iframe/sandbox changes. Numeric TTL/retry/tombstone values are implementation defaults pending owner confirmation (D-MCP-1).

<div><a href="https://cursor.com/agents/bc-c1dfa405-6ebf-4d49-83ae-8964f78b4179?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c1dfa405-6ebf-4d49-83ae-8964f78b4179&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


gtanczyk:
@codex

gtanczyk:
@codex

apps/api/src/platform/oauth-scopes.ts
```diff
@@ -6,22 +6,23 @@ export { looksLikeAsAccessToken };
...
+  return cliSurfaceEnabled(env) ? [MCP_SCOPE, CREATOR_SCOPE, OWNERSHIP_SCOPE] : [MCP_SCOPE, OWNERSHIP_SCOPE];
 }
 
 export function parseOAuthScopes(raw: string | undefined, env: NodeJS.ProcessEnv = process.env): OAuthScope[] | null {
   const tokens = (raw ?? MCP_SCOPE).trim().split(/\s+/).filter(Boolean);
   if (tokens.length === 0) return null;
   const unique: OAuthScope[] = [];
   for (const token of tokens) {
-    if (token !== MCP_SCOPE && token !== CREATOR_SCOPE) return null;
+    if (token !== MCP_SCOPE && token !== CREATOR_SCOPE && token !== OWNERSHIP_SCOPE) return null;
```
gtanczyk, line 25: `ownership` is now accepted here on every surface that calls `parseOAuthScopes`, including the CLI device flow (`oauth-device.ts:128` only requires `creator`). But `devicePage()` renders no scope copy at all, so a creator can approve a device grant that carries `ownership` without ever being shown what it permits — which undercuts the PR's "not granted unless the creator consents" property.

Either render the requested scopes (with the ownership warning) on the device page, or reject `ownership` in the device-flow path and keep it to the surfaces that do show consent.

---

apps/api/src/store/slices/game-transfer-proposal.ts
```diff
@@ -0,0 +1,281 @@
...
+    });
+  }
+
+  async eraseTransferProposalsForUid(uid: string, at: string): Promise<void> {
+    const snap = await this.db.collection('gameTransferProposals').where('ownerUid', '==', uid).get();
+    for (const doc of snap.docs) {
+      const row = doc.data() as GameTransferProposal;
+      if (!proposalIsOpen(row, at)) continue;
+      await doc.ref.set({ ...row, invalidatedAt: at });
+      await this.slugDoc(row.slug).delete();
```
gtanczyk: The Firestore implementation deletes `gameTransferProposalBySlug/{slug}` unconditionally, while the in-memory twin guards the equivalent delete with `openBySlug.get(row.slug) === id`.

Failure case: owner A has an open proposal for `slug`, it gets superseded and the slug pointer now points at owner B's open proposal. Erasing A's data (account deletion / GDPR erase) deletes the pointer to B's proposal too, so a later `invalidateOpenTransferProposalsForSlug(slug)` silently no-ops and B's proposal stays confirmable until its 24h TTL.

Suggest mirroring the in-memory guard: read the slug doc and only delete it when it still points at this proposal id. Worth a Firestore-path test too — the erase behaviour is currently only covered in-memory.

---

apps/api/src/store/slices/game-transfer-proposal.ts
```diff
@@ -0,0 +1,281 @@
...
+}
+
+function openBlocks(
+  open: GameTransferProposal | null,
+  ownerUid: string,
+  accessRevision: number,
+  receiptKey: string,
+  at: string,
+): 'busy' | 'stale' | null {
+  if (!open || open.ownerUid !== ownerUid || !proposalIsOpen(open, at) || open.receiptKey === receiptKey) return null;
```
gtanczyk: `open.ownerUid !== ownerUid` returns `null`, i.e. "nothing blocks" — so when ownership changed and the *previous* owner still has an open proposal, propose neither blocks nor invalidates it. The new proposal then overwrites the slug pointer and the old one is orphaned: unreachable by slug, still `open`, still confirmable by whoever holds its link for the full 24h TTL.

The `stale` branch (same-owner, different `accessRevision`) already invalidates the old proposal; the different-owner case is strictly more stale and should do at least the same.

---

apps/api/src/store/in-memory.ts
```diff
@@ -226,9 +228,7 @@ export class InMemoryStore extends SubmissionFacade implements Store {
     this.gameAdmissionStore.gameAgentKeys.set(slug, copy);
   }
 
-  async getUser(uid: string): Promise<User | null> {
-    return this.identityStore.getUser(uid);
-  }
+  getUser = (uid: string) => this.identityStore.getUser(uid);
```
gtanczyk: Drive-by unrelated to ownership transfer: `getUser` changed from a prototype method to an instance arrow field. Instance fields are assigned in the constructor and shadow subclass prototype methods, so any subclass overriding `getUser` silently stops taking effect (`InMemoryStore` is subclassed in `submissions.test.ts`).

If the arrow form isn't needed for `this` binding at a call site, keeping it a normal method avoids the footgun.

---

gtanczyk (commented):
Reviewed the ownership-transfer slice (39 files, +2219/−161): new `ownership` OAuth scope, three MCP tools, the proposal store slice (in-memory + Firestore), the Studio confirm route and SPA surface.

Most of it holds up. The confirm route re-checks `canActOnGame` and `accessRevision` before minting the invitation and cancels the invitation if `confirmTransferProposal` loses the race; the Firestore propose transaction does all reads before writes; scope matching is token-based with no substring bleed; route paths don't collide and the i18n keys, CSS classes and `TransferStep` value all exist.

Four things worth a look, inline:

1. `ownership` reaches the CLI device flow, which renders no scope copy — consent gap.
2. Firestore `eraseTransferProposalsForUid` deletes the slug pointer unconditionally, unlike the in-memory twin — can orphan another owner's open proposal.
3. `openBlocks` returns `null` for a different-owner open proposal, so it is neither blocked nor invalidated.
4. Unrelated: `getUser` moved to an instance arrow field, shadowing subclass overrides.

Non-blocking notes: `TRANSFER_PROPOSAL_RETRY_WINDOW_MS`, the `'pending'` receipt status and the `'stale_owner'` propose reason are never produced or consumed; and `transferProposalRequestHash` hashes the raw `expectedAccessVersion` string, so `"v01"` vs `"v1"` on a retry with the same idempotency key yields `idempotency_conflict`.

---

gtanczyk:
@codex

gtanczyk:
@codex

gtanczyk:
@codex

PR #1406, opened 2026-09-18
fix(api): cache derived GameAccess lookups for 30s

gtanczyk:
## Why

Query Insights (17–18 Sep) ranked `COLLECTION /submissions WHERE slug = ?` first by reads: **41,288 executions, 96,482 reads**, 2.337 docs scanned on average. Cheap per call, enormous in volume — an uncached lookup on a polled path, not an expensive query.

`resolveGameAccess` does the `gameAccess/{slug}` doc-get first and falls back to `listSubmissionsBySlug` whenever there is no canonical record. That fallback sat under `canActOnSlug` / `canActOnSubmissionOrSlug` (Studio status poll, slug routes, transfers, editor invites, moderation) with no window at all.

The population that takes the fallback is structural: the GameAccess backfill left 193 slugs out because several uids hold rounds on them, plus anything that predates the model. This change does **not** promote those slugs to canonical records.

## What

A process-local cache in front of the **derived** branch only (`derived-access-cache.ts`):

- 30s TTL, sized against authorization risk (private prior-round chat), matching the session-user bound — not a content window
- in-flight dedup so concurrent callers share one read
- keyed by store and slug
- explicit invalidation on authority-changing writes: `ensureGameAccess`, `recordSettledOwner`, `backfillGameAccess`, transfer accept (`transferredAccess`), editor accept (`withEditorAdded`), editor remove / leave (`withEditorRemoved`), membership erasure (`withMemberErased`), slug claim, and account erasure

The canonical branch is **not** cached. It is already one doc get; folding it in would change the staleness contract for every permission check in the product.

## Staleness

Caches are process-local and the API runs `--max-instances 4`. Invalidation on one instance does not reach the others. The bound is **one window for any action, including your own**. On the instance that served the write, the next resolve is live.

A derived-only abandon (newest live round closed, no GameAccess row) is not hooked and can sit until the window ends, on every instance.

Recorded in `docs/firestore-read-cost.md`.

## Tests

Covered on both `InMemoryStore` and `FirestoreStore(fakeFirestore().db)`.

Fixtures are discriminating: `legacy-shared` has two uids and a **newer** round lives on `other-game`. If `listSubmissionsBySlug` dropped its slug filter, derived owner would be hopper, not grace.

Saving is asserted, not only correctness: a spy on `listSubmissionsBySlug` shows a second resolve inside the window does not call it, and that `deleteAccountIdentity` makes the next resolve call it again.

Revert check, as this repo does it:

1. Remove the `readDerivedAccessCached` wrap in `resolveGameAccess` — the “lists once inside the window” and “lists again after an invalidating write” tests fail (`called 2 times` instead of 1 on the first pair of resolves). Restore — they pass.
2. Remove `dropDerivedAccessMany` from `deleteAccountIdentity` — the invalidation test fails (`called 1 times` instead of 2; the former owner stays cached). Restore — it passes.

<div><a href="https://cursor.com/agents/bc-2fb9e391-ddad-4171-9ed0-8e66a7e2ab1b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2fb9e391-ddad-4171-9ed0-8e66a7e2ab1b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/api/src/platform/game-access-resolve.ts
```diff
@@ -62,15 +63,24 @@ function fromRecord(record: GameAccessRecord): ResolvedGameAccess {
   };
 }
 
-export async function resolveGameAccess(store: GameOwnerLookup, slug: string): Promise<ResolvedGameAccess> {
+export async function resolveGameAccess(
+  store: GameOwnerLookup,
+  slug: string,
+  now: () => number = Date.now,
+): Promise<ResolvedGameAccess> {
   const record = await store.getGameAccess(slug);
```
gtanczyk, line 71: **The guaranteed miss is still billed on every call.** For the population this cache exists for — the 193 slugs the backfill quarantined — this get misses *every single time*, and Firestore bills a read for a get on a missing document. It shows up as its own `NOT_FOUND` type in `document/read_count`: roughly 10k of them in a three-day sample.

So after this change a derived resolve still costs one billed read per call. The collection scan is gone, which is the big win, but "cached" reads as cheaper than it is.

Caching the *absence* of the canonical record would remove it, and would be nearly free to wire: every path that creates a canonical record (`ensureGameAccess`, `recordSettledOwner`, `backfillGameAccess`, `claimSubmissionSlug`, `setSubmissionSlug`) already calls `dropDerivedAccess`.

I do not think you should do it. Today a canonical record written on another instance takes effect immediately, because this get is live; caching the absence would ignore it for up to 30s, and being the authority is the whole point of that record. But the remaining cost is worth a sentence in `docs/firestore-read-cost.md` so the next person measuring this path knows it was a decision and not an oversight.

apps/api/src/platform/derived-access-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+  cache.inFlight.delete(slug);
+  cache.generation += 1;
+}
+
+export function invalidateDerivedAccessMany(store: object, slugs: readonly string[]): void {
+  for (const slug of slugs) invalidateDerivedAccess(store, slug);
+}
+
+// Tests share a process; a carried window is a false pass.
+export function clearDerivedAccessCache(store: object): void {
```
gtanczyk, line 80: This is the one path that does not bump `generation`, while `invalidateDerivedAccess` does. A read already in flight when `clear()` runs will still resolve and write itself into the map that was just cleared, because its captured generation still matches.

The comment directly above says "Tests share a process; a carried window is a false pass" — which is exactly the failure this leaves open, in the one helper written to prevent it. Bumping `generation` here too is a line.

apps/api/src/platform/derived-access-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+interface Entry {
+  expiresAt: number;
+  value: unknown;
+}
+
+interface StoreCache {
+  entries: Map<string, Entry>;
+  inFlight: Map<string, Promise<unknown>>;
+  // Bumped by each drop, so a read cannot seal in staleness.
+  generation: number;
```
gtanczyk, line 24: Worth a comment that this is deliberately per-store rather than per-slug: invalidating one slug stops *every* in-flight read, for any slug, from caching its result. That is the safe direction, and on `dropDerivedAccessMany` (account erasure touching many slugs) it bumps once per slug and will drop unrelated concurrent reads.

It self-corrects on the next call, so it is not a bug. But a per-slug generation looks like an obvious optimisation to someone reading this cold, and the reason it is not one should be written down rather than rediscovered.

gtanczyk (commented):
Reviewed by checking the claims rather than reading them: removed the slug filter (4 tests fail across both stores), removed the `readDerivedAccessCached` wrap (4 tests fail), ran the full API suite (no regressions beyond the known local `improvement-admission` timeout), lint and API type-check both clean.

The design holds where it matters. The generation guard against a read in flight during invalidation is present and correct — that exact race was a real bug in #1384, and it is right here first time. `dropDerivedAccess` passes `this` from `SubmissionFacade`, which is the same object callers hand to `resolveGameAccess`, so the `WeakMap` key matches and invalidation is not silently dead. And what is cached is the ownership record rather than a per-viewer permission answer, which is what makes a per-slug key safe.

Three notes below. None blocks the merge.

apps/api/src/platform/game-access-resolve.ts
gtanczyk: Addressed in `f0e3d5a`-era follow-up: the doc now states that a miss is billed and that the canonical get stays live on purpose. That was the ask — the cost is a decision on the record rather than a surprise for whoever measures this path next.

apps/api/src/platform/derived-access-cache.ts
gtanczyk: Fixed — `clearDerivedAccessCache` now bumps `generation` alongside clearing both maps, so a read that straddles the clear can no longer seal itself back in.

apps/api/src/platform/derived-access-cache.ts
gtanczyk: Documented: "Store-wide on purpose: one slug drop voids every in-flight seal" plus the erasure case. That is exactly what I wanted — the next reader now knows a per-slug generation is a deliberate non-optimisation rather than an oversight.

PR #1407, opened 2026-09-18
Stop forgotten localStorage tokens from polling forever

gtanczyk:
## What happened

Query Insights for 2026-09-18 put three `/submissions` collection scans at the top of the day's reads. Cloud Run logs for 16:50–17:30Z showed **1102 of 1120** status polls from one Chrome user agent, across **32 job tokens**, in lockstep (29 tokens in the same one-second bucket). The load died at 17:25Z when that browser closed. `/submissions` was 221,334 of ~280k reads.

## Root cause

`loadCreatorGames` takes every token in this browser's localStorage (`getSavedSpecs()`), subtracts what `/api/submissions/mine` returned, and `Promise.all`s `getSubmissionStatus` for the rest. The resolved statuses were never written back. `removeSpec` existed and had no callers. A browser used since the anonymous era therefore re-asks about every token it ever created — including jobs 19, 22, 24, 29, 32, 35, 37, 92, 130 — forever, once a minute.

Each of those polls does a permission check. For a game with no canonical `GameAccess` record, `resolveGameAccess` falls through to `listSubmissionsBySlug` — the `WHERE slug = ?` scan. `/mine` and the open-round badge from the same tab explain the other two Insights rows.

This is occupancy with a longer memory: it scales with how long a browser has been used, not with how many creators are watching something now.

## Fix (web only)

An unlisted token that resolves terminal or missing (`abandoned`, HTTP 404, or a rejected token / 400) is pruned via `removeSpec`. A settled, not-in-flight status (`published`, `needs_changes`) is stored as `lastStatus` so the next load still renders the game and does not re-ask.

In-flight is the only remaining re-ask. A live round the server's shelf has not listed yet (anonymous-era, signed-out, or a round `/mine` has not collapsed) can still move, and the Studio chip should see that. Lengthening the home poll would not have helped; the cost is the size of the fan-out.

The derived `listSubmissionsBySlug` cache that was in the first revision of this PR is dropped. That backstop overlaps #1406, which hooks slug-claim writes this change missed and is stricter about in-flight invalidation. Do not promote the quarantined slugs to canonical records; the backfill excluded them because their ownership is genuinely ambiguous.

Recorded in `docs/firestore-read-cost.md`.

## Residual

Unlisted tokens that stay in-flight (`queued` / `building` / `in_review` / `publishing`, or `null`) are still asked on every home load. That is the legitimate re-ask, and nothing here ages it out. Ancient jobs the notify sweep has already auto-abandoned prune on the first answer. Rounds that stay non-terminal keep fanning out from that browser until they settle or the spec is cleared.

Do not cap the list by age: an age cut would hide a live anonymous round the shelf has not listed yet, which is the case this list exists for. Measure after deploy. If the remaining fan-out is still the day's hottest query, that is a new decision.

## Tests

**Revert-to-fail, then restore:**

- Web: reverted the prune/remember path in `loadCreatorGames`. `re-asks only unlisted tokens that last resolved live` failed — the second call asked `tok-done`, `tok-gone`, and `tok-dead` again. Restored; passed.
- A 502 keeps the token on the ask list; a 400 prunes it.

The API cache tests from the first revision are gone with that cache.

## Gates

On this revision (`069e0023c`): `npm run type-check`, `npm run lint`, `npm run test`, and `npm run build` all exited 0.

<div><a href="https://cursor.com/agents/bc-06f7990e-4644-4008-8499-a7fab6b8fedb?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-06f7990e-4644-4008-8499-a7fab6b8fedb&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


gtanczyk (commented):
Verified by executing rather than reading: neutered the prune/remember path (2 web tests fail, including the one separating a transient failure from a rejected token), removed the cache wrap (API tests fail), lint clean, 19/19 green. `throwResponseError` does set `error.status` and `fetchCached` rethrows the original error, so the 404/400 prune actually fires. `isSubmissionInFlight(null)` is `true`, so a token that has never resolved still gets its one ask — no regression for a genuinely new round.

The web half is the real fix and it is sound.

**The blocker is not in the code: this PR and #1406 both add the same API cache.** #1406 adds `platform/derived-access-cache.ts`; this one adds `platform/game-access-derived-cache.ts`. Both rewrite the same branch of `game-access-resolve.ts`, both touch `submission-facade.ts`, `firestore.ts`, `in-memory.ts`, `module-boundary-map.mjs`, `module-size-baseline.json` and the same doc section. Whichever merges second conflicts, and a careless resolution lands two caches doing one job.

They are not equivalent, and the difference matters — see the `submission-facade.ts` note. My suggestion: keep #1406's API half (it hooks three invalidation paths this one misses and is stricter on the in-flight/invalidate interaction), drop the API half from here, and let this PR be the web fix it is described as. The alternative — porting the three hooks here and closing #1406 — works too, but it is more moving parts for the same result.

apps/api/src/store/submission-facade.ts
```diff
@@ -115,7 +116,9 @@ export abstract class SubmissionFacade {
   }
 
   async ensureGameAccess(slug: string, ownerUid: string, workAt: string, at: string): Promise<GameAccessRecord | null> {
-    return this.gameAccessStore.ensureGameAccess(slug, ownerUid, workAt, at);
+    const record = await this.gameAccessStore.ensureGameAccess(slug, ownerUid, workAt, at);
+    invalidateDerivedGameAccess(this, slug);
```
gtanczyk: **The hooks stop short of the slug-claim paths, and that gap lands on exactly the population this cache exists for.**

This one is right. But `claimManualRoundSlug`, `claimSubmissionSlug` and `setSubmissionSlug` are not hooked, and assigning a slug to a round changes the derived answer: `deriveOwnerFromSubmissions` returns the newest **live** round's owner.

The usual reasoning — "a slug claim settles a canonical record, which shadows the stale entry" — does not hold in the case that matters. In `setSubmissionSlug`, when a slug has more than one live owner, `owners.size !== 1` returns **before** `ensureGameAccess`, so no canonical record is written. That is the quarantined multi-uid case: the 193 slugs the backfill deliberately left derived, which is why this cache was written at all.

So on those slugs the cache is the only authority and nothing drops it:

1. Legacy slug, live rounds from A (newest) and B. No canonical record. Derived owner resolves to A, cached 30s.
2. B's round takes the same slug via `setSubmissionSlug`. Newest live round is now B's, so the true derived owner is B.
3. The multi-owner early return fires, no canonical record, stale entry not shadowed.
4. For up to 30s `canActOnSlug` still answers **A**.

B is refused on the slug they just claimed, and A keeps owner-level access — including the private prior-round chat this window was explicitly sized around — after ownership has already moved.

`claimSubmissionSlug` has a narrower version: `tryRecordOwner` calls `recordSettledOwner`, which returns the existing record without writing whenever it is not pristine or settlement does not win.

#1406 hooks all three. That is the concrete reason to prefer its API half over this one.

apps/api/src/platform/game-access-derived-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+      return access;
+    })
+    .finally(() => {
+      cache.inFlight.delete(slug);
+    });
+  cache.inFlight.set(slug, reading);
+  return reading;
+}
+
+export function invalidateDerivedGameAccess(store: object, slug: string): void {
```
gtanczyk: This drops the cached entry but leaves the in-flight promise in place, so a caller arriving immediately after an invalidating write joins a read that started *before* it and gets the pre-write answer.

The window is milliseconds and the generation guard still stops that read from caching itself, so this is not the same severity as the facade note. But for an authorization cache the stricter behaviour is the right default: dropping the in-flight entry too makes the next caller start a fresh read. #1406 does exactly that, and pairs it with an identity check in `finally` (`if (cache.inFlight.get(slug) === read)`) so the settling read cannot delete a newer one that replaced it.

apps/api/src/platform/game-access-derived-cache.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function invalidateAllDerivedGameAccess(store: object): void {
+  const cache = caches.get(store);
+  if (!cache) return;
+  cache.entries.clear();
+  cache.generation += 1;
+}
+
+export function clearDerivedGameAccessCache(store: object): void {
```
gtanczyk: `invalidateDerivedGameAccess` and `invalidateAllDerivedGameAccess` both bump `generation`; this one does not. A read already in flight when `clear()` runs will still resolve and write itself into the map that was just cleared, because its captured generation still matches.

This is the test-isolation helper, so the failure it leaves open is a window carried between tests — a false pass. One line.

apps/web/src/unlistedSpecs.ts
```diff
@@ -0,0 +1,27 @@
+import { isSubmissionInFlight } from '@gamedevpl/contract';
+import type { SubmissionState } from './submissionApi.js';
+import { getSavedSpecs, removeSpec, saveSpec } from './mySpecs.js';
+
+export function shouldAskUnlisted(status: SubmissionState | null): boolean {
```
gtanczyk, line 5: Worth stating the residual explicitly, because it decides whether this fix actually removes the observed load: a token stuck in a non-flight state is pruned, but a token stuck **in flight** is re-asked forever, by design.

The ancient jobs in the 2026-09-18 sample (19, 22, 24, 29, 32, 35, 37, 92, 130) are old enough that quiet-round auto-abandon should have settled them, so they should prune — but this repo has a documented population of long-lived non-terminal rounds (the sweep reports ~34 active submissions that have all long since been delivered). Those keep their fan-out.

Not something to fix speculatively. Worth measuring the poll count for that browser after this deploys, and if it has not collapsed, capping by age: a spec that has been in flight for months is not going to move.

apps/api/src/store/submission-facade.ts
gtanczyk: Resolved by removal: the API half is gone from this branch entirely (`fix(web): drop the duplicate derived GameAccess cache`). The diff is now six files, all web plus the doc and a baseline.

That takes the invalidation bug with it — there is no derived cache here to leave stale — and settles the duplication with #1406. Confirmed against the branch: the only files still shared with #1406 are `docs/firestore-read-cost.md` and `eslint-rules/module-size-baseline.json`, which will conflict textually but resolve trivially.

apps/api/src/platform/game-access-derived-cache.ts
gtanczyk: Obsolete — the cache module is no longer part of this PR. The equivalent code lives in #1406, where `invalidateDerivedAccess` does drop the in-flight entry and the `finally` carries the identity check.

apps/api/src/platform/game-access-derived-cache.ts
gtanczyk: Obsolete for the same reason — this module left the branch. #1406's `clearDerivedAccessCache` now bumps the generation.

apps/web/src/unlistedSpecs.ts
gtanczyk: No change needed, and not acting was the right call — this was a note to measure, not a defect. `shouldAskUnlisted` still re-asks anything in flight, which is correct for a round that can still move.

Worth keeping on the list for after this deploys: if that browser's poll count does not collapse, the residue is rounds stuck in a non-terminal state, and an age cap is the answer. Measuring first beats capping speculatively.

PR #1408, opened 2026-09-18
fix(api): keep the shelf mirror in agreement with source

gtanczyk:
Readers stay on source. This only hooks the writers that were drifting the `shelves/{ownerUid}` document, so the shadow can run clean for a week before anyone considers graduating.

## What production showed (week to 2026-09-18)

`shelf-shadow.ts` compared the document against live rounds on every shelf load:

- 656 `collapse` (same round count, different content)
- 102 `count` (`sourceCount` 4, `shelfCount` 3)
- 3 distinct owners
- 0 `absent`

Not a race and not warm-up: it repeated every few minutes. `MAX_SHELF_ROUNDS` is 2,000 and the owners had 4 and 155 rounds, so truncation was not it. `WHERE ownerUid = ?` remains the second-heaviest query (~123 documents per execution) until the mirror agrees.

Recorded in `docs/firestore-read-cost.md`. Graduate readers only after a later week is **zero mismatches** split by verdict. Do not flip in this PR.

## The two faults

`shelf-write-through.test.ts` claimed every shelf-relevant writer and passed because `agrees()` judged `listSubmissionsByOwner` rather than the reconciled records production actually serves.

1. **`count` (4 vs 3)** — a round in source never reached the document.
   - `acceptGameTransferInvitation` / `acceptEditorInvitation` changed membership without rebuilding the shelf, so a transferred or shared round existed in `reconcileTransferredOwnership` and not in the document.
   - Firestore `countSubmissionsByOwner` took a raw `ownerUid` `count()` whenever the owner had no remaining `gameAccess` rows. A sender who kept slugless rounds and transferred the slugged one compared 4 against a correctly rebuilt 3 forever. The hourly pass cannot heal a count definition that disagrees with the rebuilt document.

2. **`collapse` (same count, different content)** — `leaveGame` wrote `state: canceled` on the editor's live tip without refreshing the shelf, so the document still served that round as the tip while source collapse dropped it. The diverging field is one the fingerprint already covers (`jobId` / `createdAt` of the tip, via `state`).

`setSubmissionSlug` / `claimSubmissionSlug` also wrote `gameAccess` *after* the slug rebuild. Reconcile reads that record, so the document was one rebuild behind; both now rebuild again after access lands.

## What this does

- Rebuild affected shelves from the store facade after transfer accept, editor accept, editor remove, and leave.
- Rebuild after a slug's access record is written.
- Make Firestore `countSubmissionsByOwner` always use `reconcileTransferredOwnership`, matching InMemory and the rebuilt document.
- Judge write-through tests through `judgeShelfShadow(...).verdict === 'match'` against reconciled records, on both stores.
- Pin an owner with several slugged rounds and **no** `gameAccess` rows (`agrees for an owner whose slugs never kept a gameAccess row`) — the shape where dropping the `count()` early-exit costs more reads.

Does **not** point readers at the document, quiet the shadow, or rebuild the shelf on read.

## Read cost

Dropping the `count()` early-exit looks like a regression (aggregation replaced by a full reconcile). Measured against #1409's gate on this branch, `GET /api/submissions/mine` goes **44 → 39**: the old pre-check queried `listGameAccessByMember` and then `reconcileTransferredOwnership` queried it again.

Whoever merges this and #1409 second should run:

```
npm run firestore-read-cost -- "GET /api/submissions/mine" --write --force
```

so those five reads stay locked. That fixture's creator has `gameAccess` rows, so it cannot see the no-access owner. Add that shape when resealing if the cost of this path should be gated too.

## Revert-fix-restore

Each new test was verified against its writer, on both stores, through the shadow's own verdict:

| Reverted writer | Test | Without the fix | Restored |
| --- | --- | --- | --- |
| `acceptGameTransferInvitation` shelf rebuild | `mirrors a transferred round that list-by-owner would miss` | `count` (grace 4 vs 3) | `match` |
| `leaveGame` shelf rebuild | `refreshes the tip when an editor leave cancels the newer round` | `collapse` | `match` |
| Firestore `countSubmissionsByOwner` `!access.length` shortcut | sender assertion in the transfer test (and the editor after leave) | `count` on FirestoreStore only | `match` |

Write-through tests stamp `createdAt` with a fake clock rather than sleeping.

## Gates

`npm run type-check`, `npm run lint`, `npm run test`, and `npm run build` all exited 0 on the writer/count fix. The derived-only owner test is 30/30 on both stores.

<div><a href="https://cursor.com/agents/bc-3a36285d-b72a-4b32-bfc5-d3aab0f6658a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3a36285d-b72a-4b32-bfc5-d3aab0f6658a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/api/src/store/firestore.ts
```diff
@@ -438,8 +438,7 @@ export class FirestoreStore extends SubmissionFacade implements Store {
   }
 
   async countSubmissionsByOwner(ownerUid: string): Promise<number> {
-    const access = await this.listGameAccessByMember(ownerUid);
-    if (!access.length) return this.shelfStore.countSubmissionsByOwner(ownerUid);
+    // Must match the rebuilt shelf, not the ownerUid query.
```
gtanczyk, line 442: Measured: this is a **net win** on the fixture (`/api/submissions/mine` 44 → 39 with #1409's gate applied), because the old pre-check queried `listGameAccessByMember` and then `reconcileTransferredOwnership` queried it again.

But the early exit was doing one real job that is now gone. An owner with **no** `gameAccess` rows used to cost 2 reads here: the access query plus a `count()` aggregation. Now they pay the full reconcile — one read per round from `listSubmissionsByOwner`, plus a `resolveGameAccess` for every slugged round, since with no canonical slugs every record falls into `nonCanonical`.

That is precisely the quarantined population: an owner whose games are all among the 193 the backfill left derived. The fixture cannot see it because its creator has access rows.

Not a blocker — the shape is bounded, and the derived-access cache in #1406/#1407 blunts exactly this once one of them lands. But it deserves a fixture so the gate can watch it, rather than being a case nobody measures.

apps/api/src/store/shelf-write-through.test.ts
```diff
@@ -2,23 +2,46 @@ import { describe, expect, it } from 'vitest';
...
 
 // Both stores: the write-through spans facade and class.
 const IMPLEMENTATIONS: Array<[string, () => Store]> = [
   ['InMemoryStore', () => new InMemoryStore()],
   ['FirestoreStore(fake)', () => new FirestoreStore(fakeFirestore().db)],
 ];
 
-// The write-through claim, asked as the shadow asks it.
+// Collapse sorts ties by input order; a shared millisecond is not drift.
+function afterATick(): Promise<void> {
```
gtanczyk: A wall-clock sleep to separate timestamps is a flake this repo has already paid for: `creator-messages-parity.test.ts` used a 1ms gap, two writes landed in the same millisecond under full-suite CPU load, and the ordering assertion failed. It had to be widened.

5ms is more generous, and it passed here — but it is the same dependency, and the full suite is ~400 files. If the tie really is about ordering, injecting the timestamps is deterministic where sleeping is only probably long enough.

docs/firestore-read-cost.md
```diff
@@ -298,6 +301,31 @@ The 09-20 checkpoint should read the weekly mismatch count split by verdict, not
 only `version` / `truncated` / `count` / `collapse` mean the document and source actually
 disagreed.
 
+**What the week to 2026-09-18 actually showed**, before the writers below were hooked:
+656 `collapse`, 102 `count`, 0 `absent`, 3 distinct owners. Not a race and not warm-up —
+it repeated every few minutes, and `absent` (the benign "not built yet") never appeared.
+`MAX_SHELF_ROUNDS` is 2,000 and the owners had 4 and 155 rounds, so truncation was not it.
+
+The two verdicts were two different unhooked writers, both real:
+
```
gtanczyk, line 395: Follow-up for after both land, so the gain does not evaporate quietly: #1409's gate records `GET /api/submissions/mine` at **44**, and this PR takes it to **39**. The gate lets a route shrink silently, so the ratchet stays parked at 44 until somebody reseals it.

Worth a line in the merge order — whoever merges second should run `npm run firestore-read-cost -- "GET /api/submissions/mine" --write --force` so the 5 reads this PR saves are actually locked in.

gtanczyk (commented):
The diagnosis holds up. I confirmed the vacuity claim directly: the old `agrees()` judged `listSubmissionsByOwner`, while production's `loadShelfRecords` hands `recordShelfShadow` the **reconciled** list — so the helper was asking a different question than the shadow, which is why it passed while prod logged 656 collapses. The new helper reconciles first and matches.

The tests are load-bearing: reverting the `leaveGame` hook fails `refreshes the tip when an editor leave cancels the newer round` on **both** stores. Full API suite clean, lint clean.

**I measured the read impact rather than trusting my own reasoning, and I was wrong.** Removing the `count()` early exit looked like it had to make the polled path more expensive, so I merged #1409's gate onto this branch and ran it: `GET /api/submissions/mine` goes **44 → 39**. The old code called `listGameAccessByMember` twice — once in the pre-check, once again inside `reconcileTransferredOwnership` — so dropping the pre-check nets a saving. Worth recording in the PR body, because "replaced an aggregation with a full reconcile" reads like a regression to the next person too.

It also does none of the three things it should not: readers stay on source, the shadow is not quieted, and nothing rebuilds on read.

Two notes and one follow-up below.

apps/api/src/store/shelf-write-through.test.ts
gtanczyk: Fixed, and better than what I suggested: `afterATick` is gone in favour of `vi.useFakeTimers` with an injected clock and `afterEach(() => vi.useRealTimers())`. That is deterministic rather than probably-long-enough. Re-ran the file on the branch: 28/28, no hang from mixing fake timers with the async store paths.

docs/firestore-read-cost.md
gtanczyk: Recorded — the doc now carries the 44 → 39 figure and the reason (the old pre-check queried `listGameAccessByMember` before `reconcileTransferredOwnership` queried it again).

The reseal itself still has to happen after both land, since #1409's baseline is measured on master and reads 44. Whoever merges second: `npm run firestore-read-cost -- "GET /api/submissions/mine" --write --force`.

apps/api/src/store/firestore.ts
gtanczyk: Still open, and still not a blocker — flagging so it is a decision rather than an omission.

The two new tests cover the transferred round and the editor-leave tip. Neither exercises an owner with **no** `gameAccess` rows at all, which is the shape the removed `count()` early exit was carrying: 2 reads before, one read per round plus a `resolveGameAccess` per slugged round after, because with no canonical slugs every record lands in `nonCanonical`.

That is an owner whose games are all among the 193 quarantined slugs. #1406's derived cache blunts it once merged, so the practical urgency is low — but without a fixture the gate in #1409 cannot see this shape at all, and it is the one case where this change is a regression rather than the measured 44 → 39 win.

gtanczyk:
Follow-up on the one thread still open, with enough detail to act on.

## The fixture that is missing

Every new test here gives the owner `gameAccess` rows. That is the shape where removing the pre-check is a **win** — measured 44 → 39, because the old path queried `listGameAccessByMember` and then `reconcileTransferredOwnership` queried it again.

The shape where it is a **loss** has no coverage: an owner with **no `gameAccess` rows at all**. Before, that cost 2 reads (the access query, then a `count()` aggregation). Now it is the full reconcile — one read per round from `listSubmissionsByOwner`, plus a `resolveGameAccess` for every slugged round, because with no canonical slugs every record falls into `nonCanonical`.

That is not a hypothetical owner. It is one whose games are all among the 193 slugs the GameAccess backfill left derived on purpose.

Roughly what it needs:

- an owner with several slugged rounds and **no** `ensureGameAccess` call for any of them
- assert the shelf still agrees (`judgeShelfShadow(...).verdict === 'match'`), so the correctness this PR is about is pinned for that owner too
- and, if you want the cost pinned rather than just the correctness, a route case in #1409's fixture that drives `/api/submissions/mine` for such an owner — today the gate cannot see this shape at all

I do not think it blocks the merge. #1406's derived cache absorbs most of it once that lands, and the owners affected are few. But it is the single case where this change makes reads worse, and right now nothing measures it, so it would regress silently.

## Merge order, so the saving is not lost

#1409 records `GET /api/submissions/mine` at **44**, measured on master. This PR takes it to **39**. The gate lets a route shrink silently, so the ratchet stays parked at 44 until someone reseals it.

Whoever merges second:

```bash
npm run firestore-read-cost -- "GET /api/submissions/mine" --write --force
```

Otherwise the 5 reads this PR saves stay available for the next regression to spend.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

PR #1409, opened 2026-09-18
feat(api): gate polled Firestore reads per route

gtanczyk:
Query Insights ranks routes by total reads, which is traffic times cost. The number we own is **documents billed per execution**, and until now nothing caught a query getting more expensive until it showed up in the bill a day later.

This is the same ratchet as module-size and comment-prose, applied to billed Firestore reads on the four polled surfaces.

## Why not `read-meter.ts`

`installReadMeter()` monkey-patches the prototypes of the real `@google-cloud/firestore` classes. Tests never construct those: they run on `InMemoryStore`, or on `FirestoreStore(fakeFirestore().db)` whose fake is a hand-rolled object. Wired as-is, every route would measure zero and the gate would be vacuous.

The counter lives on `fakeFirestore()`, which already materialises gets, queries, `count()` and `getAll`.

## What Firestore bills (not what the code called)

- a document get is 1, whether or not it exists
- a query is one read per document returned, **minimum 1 even when it returns nothing**
- `count()` is 1, not the number of rows it counted
- `getAll` is one per reference requested

## Recorded today (exact, no headroom)

| Route | Billed reads |
| --- | ---: |
| `GET /api/submissions/mine` | 44 |
| `GET /api/submissions/:token` | 22 |
| `GET /api/review/status` | 12 |
| `GET /api/notifications` | 8 |

A baseline is not a target. These are what the fixture costs today.

## Fixture

Fixed-size and explicit so the baseline cannot flap with helper growth: 8 creator rounds, 12 decoy rounds, 5 events and 3 messages on the polled job, 6 creator notifications, 10 decoy notifications, 7 reviewer assessments, 15 decoy assessments, 4 decoy re-reviews, one active sweep. Decoys are why removing a `where` fails instead of still passing.

Each route is measured on a fresh fake + store + app so identity / load-shed caches cannot leak across routes.

The HTTP harness is `apps/api/src/store/firestore-read-cost.fixture.ts`, excluded from the API compile (`src/**/*.fixture.ts`). It is not shipped.

## Checker

```
npm run firestore-read-cost
npm run firestore-read-cost -- "GET /api/submissions/mine"
npm run firestore-read-cost -- "GET /api/submissions/mine" --write --force   # raise ONE
npm run firestore-read-cost -- --write --reseal
```

Wired into `npm run lint`. Unscoped `--write` is refused (same as module-size). Shrinks freely. Raise one path at a time with `--write --force`. CI runs lint before type-check, so the checker builds workspace packages when their dist is missing.

## Regression proof (real edit, not reasoning)

1. Removed `where('ownerUid', '==', ownerUid)` from `FirestoreSubmissionQueryStore.listSubmissionsByOwner`.
2. `firestore-read-cost.test.ts` failed: `GET /api/submissions/mine: expected 92 to be less than or equal to 44`. The unbounded-scan assertion also failed (`expected 20 to be greater than 20`) because the owner query had become a full collection get.
3. `npm run firestore-read-cost` failed: `GET /api/submissions/mine: 92 billed reads (baseline 44)`.
4. Restored the `where`. Both the test file (8/8) and the checker passed, mine back at 44.

Counter tests (kept, not just reasoned): empty query = 1, `count()` = 1, missing and present document gets = 1, `getAll` = 1 per ref.

## Docs

`docs/firestore-read-cost.md` now has **The gate** as what keeps "read a collection once per window, never once per request" true after the person who wrote it has moved on.

<div><a href="https://cursor.com/agents/bc-cad773cb-714a-4ea2-83d4-7e3c75ecead5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-cad773cb-714a-4ea2-83d4-7e3c75ecead5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


eslint-rules/firestore-read-cost-check.mjs
```diff
@@ -0,0 +1,171 @@
...
+    for (const failure of failures) {
+      if (failure.isNew) {
+        console.error(
+          `${failure.route}: ${failure.reads} billed reads — new route. ` +
+            `Record it with \`npm run firestore-read-cost -- ${JSON.stringify(failure.route)} --write --force\`.`,
+        );
+      } else {
+        console.error(
+          `${failure.route}: ${failure.reads} billed reads (baseline ${failure.allowed}). ` +
+            `Routes may shrink, never grow — fix the query, then \`npm run firestore-read-cost -- --write\`.`,
```
gtanczyk: The failure hint points at a command the tool refuses. Running `npm run firestore-read-cost -- --write` hits the guard at line 63, which then tells you to use `<route> --write --force`.

So the first thing someone does after a red lint is run the suggested command and get a refusal. Worth pointing this line straight at the per-route form, since that is the one you actually want them to use — raising a single ceiling deliberately rather than resealing the file.

apps/api/src/store/firestore-read-cost.fixture.ts
```diff
@@ -0,0 +1,349 @@
+import type { FastifyInstance } from 'fastify';
```
gtanczyk, line 1: `apps/api/tsconfig.json` excludes only `src/**/*.test.ts`, so this file compiles into the production build. Unlike `fake-firestore.ts` — which already lives in `src/` and is a passive in-memory double — this one imports `buildApp`, constructs a full app, and mints session cookies against the dev session secret.

Nothing executes it in production and the secret is the documented dev default, so this is hygiene rather than a vulnerability. But an app builder and a token minter sitting in the shipped bundle is the kind of thing that is much easier to keep out now than to explain later. Either name it `*.test.ts` (it is only ever run by the checker and the vitest file) or give the harness a path the build excludes.

gtanczyk (commented):
Verified by driving the gate rather than reading it.

**It is a real ratchet, not a decorative one.** Reverted the count proof in `studio-shelf-records.ts` so the shelf lists per game again: `GET /api/submissions/mine` went 44 → 50 and the checker exited **1**, `npm run lint` exited **1**, and both returned to 0 on restore. Dropping the `ownerUid` filter entirely took it 44 → 92. The decoys do their job: a filter that stops filtering is caught rather than absorbed.

**The billing model matches what Firestore charges**, which is the part that decides whether any of this means anything: a doc get bills 1 whether or not the document exists, a query bills one per document returned with a floor of 1 on an empty result, `count()` bills 1, and `getAll` bills per reference because it delegates to `ref.get()` — so it is covered without a special case.

Deterministic across three consecutive runs. Unscoped `--write` is refused with the same wording as module-size. Loading the harness through `tsx` rather than build output means `lint` does not depend on a prior `build`.

Two notes, neither blocking. Also worth saying out loud: the gate's first output is already interesting — `/api/submissions/mine` costs **44 reads against a fixture of 8 creator rounds**, which is more than the shelf scan alone explains. That is the next thing to pull on, and it is exactly what this PR was supposed to make visible.

eslint-rules/firestore-read-cost-check.mjs
gtanczyk: Fixed — the hint now names the per-route form with the route quoted, so the first thing someone runs after a red lint is the command that actually works.

apps/api/src/store/firestore-read-cost.fixture.ts
gtanczyk: Fixed — the harness is now `firestore-read-cost.fixture.ts` and `apps/api/tsconfig.json` excludes `src/**/*.fixture.ts` alongside the test glob, so the app builder and the token minter stay out of the shipped build. Verified the checker still resolves it through `tsx` after the move: 4 routes measured, same numbers.

PR #1412, opened 2026-09-18
Measure derived-only owner reads on GET /api/submissions/mine

gtanczyk:
The access-row `/api/submissions/mine` fixture cannot see the `countSubmissionsByOwner` shape #1408 made more expensive. This adds the missing owner: slugged rounds, **no** `gameAccess` rows — the 193 slugs the GameAccess backfill left derived on purpose.

## Two owner shapes

| Route | Owner | Billed reads |
| --- | --- | ---: |
| `GET /api/submissions/mine` | eight rounds, canonical `gameAccess` rows | 39 (unchanged) |
| `GET /api/submissions/mine (derived-only owner)` | three slugged rounds, no `gameAccess` rows | **18** |

A baseline is not a target. 18 is what the fixture costs today on a cold 30s derived-access window (fresh fake, store, and app per route). No headroom.

With no canonical slugs every record lands in `nonCanonical`, so this poll pays `listSubmissionsByOwner` plus a `resolveGameAccess` for every slugged round. The access-row owner never takes that branch.

## Exact vs at-or-under

The lint gate stays shrink-freely for every route. The derived-only vitest is **exact**, because that route exists to see movement in either direction. The other four stay at-or-under on purpose; tightening them is a separate decision. Written down in the gate section of `docs/firestore-read-cost.md` (the #1410 reseal is why the looser rule is not an oversight).

## Discrimination proof (real edit, not reasoning)

1. Restored the old pre-check in `FirestoreStore.countSubmissionsByOwner`:

```ts
const access = await this.listGameAccessByMember(ownerUid);
if (!access.length) return this.shelfStore.countSubmissionsByOwner(ownerUid);
```

2. `GET /api/submissions/mine (derived-only owner)` moved **18 → 13**. The exact-read expectation fails (`expected 13 to be 18`). Access-row `mine` moved 39 → 44 (the extra `listGameAccessByMember` that #1408 removed), which is a different curve.
3. Restored the current count path. Derived-only is 18 again; both the exact expectation and `npm run firestore-read-cost` pass.

The `agrees for an owner whose slugs never kept a gameAccess row` test still returns 3 either way — correctness cannot see this. Cost can.

## Not in this PR

The 18 is ugly on purpose. Do not "fix" the derived-only path here; the point is to see it.

## Docs

`docs/firestore-read-cost.md` gate section now names both owner shapes, why the derived-only curve is different, and why one route is exact while the gate is at-or-under.

<div><a href="https://cursor.com/agents/bc-f5a3222a-034a-4b5a-a682-2ce7ef774953?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f5a3222a-034a-4b5a-a682-2ce7ef774953&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


apps/api/src/store/firestore-read-cost.test.ts
```diff
@@ -102,4 +113,14 @@ describe('polled route read baseline', () => {
     expect(row.statusCode).toBe(200);
     expect(row.reads).toBeGreaterThan(0);
   });
+
+  it('GET /api/submissions/mine (derived-only owner) records that cost exactly', async () => {
```
gtanczyk, line 118: This route gets an **exact** expectation while the gate itself is at-or-under (line 107). Worth deciding deliberately whether that asymmetry stays, because the looser rule has already cost something.

#1408 took `GET /api/submissions/mine` from 44 to 39. The gate let it shrink silently, the ceiling stayed at 44, and the five reads sat spendable until I noticed and opened #1410 to reseal by hand. Nothing failed; nothing told anyone. That is the same gap, on a route where it actually happened, hours ago.

For a measurement route like this one, exact is clearly right — the point is to see movement in either direction. The question is whether the other four should follow. Exact everywhere means an improvement turns CI red until the author reseals, which is friction — but it is friction at the moment someone is already editing that path, rather than a silent decay nobody is looking at.

Not something this PR has to answer. Worth writing the choice down in the gate section either way, since the next person will wonder why one route is stricter than the rest.

gtanczyk (commented):
Every number in the description reproduces. I ran the gate on the branch (5 routes, derived-only at **18**), then restored the old pre-check in `FirestoreStore.countSubmissionsByOwner` and watched it move to **13** while the access-row owner went 39 → 44. The exact-read test fires with `expected 13 to be 18`, and the gate fires separately with `44 billed reads (baseline 39)`. Two shapes, two curves, both now visible.

That settles the thread left open on #1408: the derived-only path costs **5 reads more** per poll than the pre-check it replaced, and correctness testing genuinely could not see it — `agrees for an owner whose slugs never kept a gameAccess row` returns 3 either way.

The fixture establishes the shape rather than assuming it: slugs are assigned (which writes the row) and then the row is deleted, and `the derived-only mine owner has slugged rounds and no gameAccess rows` asserts `listGameAccessByMember` is empty with three slugged rounds. The cost test also asserts the two routes differ, so they cannot silently collapse into measuring the same thing. That is the failure mode I was going to ask about, already closed.

lint, API type-check, `apps/api/src/store` (468) and `eslint-rules` (86) all clean.

One question below, and it is about the gate rather than this PR.

apps/api/src/store/firestore-read-cost.fixture.ts
```diff
@@ -169,6 +178,25 @@ async function seedRound(
   await store.ensureGameAccess(round.slug, ownerUid, AT, AT);
 }
 
+// setSubmissionSlug writes a row; derived-only owners never kept one.
+async function dropGameAccess(store: Store, slug: string): Promise<void> {
```
gtanczyk, line 182: Reaching through `store as unknown as { db }` to delete the row is brittle, and I do not think there is a better option today — `setSubmissionSlug` always writes a `gameAccess` row, so there is no public way to construct an owner who never kept one. Creating and then deleting is the honest reproduction of a population that predates the model.

Worth one line saying that, so the next person who sees a private-field cast does not "clean it up" into a helper that quietly stops producing the shape. The shape assertion in the test file is what would catch them, which is a good reason it exists.

apps/api/src/store/firestore-read-cost.test.ts
gtanczyk: Resolving so this stops blocking the merge — it was a question about the gate's design, not a change request against this PR.

Carrying it forward as its own decision: whether the other four routes should get exact expectations too, given that #1408's silent 44 → 39 shrink needed #1410 to reseal by hand. This PR is the right place to *demonstrate* the stricter pattern and the wrong place to *impose* it everywhere.

apps/api/src/store/firestore-read-cost.fixture.ts
gtanczyk: Resolving — this was a suggestion for one explanatory line, not a defect. The shape assertion in the test file already protects against the failure I was worried about, which is the substantive half.

PR #1415, opened 2026-09-19
fix(web): enable agent play mode for draft and raw HTML games

gtanczyk:
## Problem
When reviewer agents (e.g. Grok bot) or users with `reviewer` permissions opened a draft or unpublished game with `?agent=1`, Agent Play Mode never engaged and the More menu did not show the agent mode option on desktop.

This happened because:
1. `GameTheater.tsx` evaluated `useAgentBridge('slug' in source)`. For draft and raw HTML games, `source` is `{ html: string }`, so `'slug' in source` was `false` and the bridge script was never fetched.
2. `GameTheater.tsx` rendered `<GameFrame ... embed />` without forwarding `agentBridge`.
3. On desktop, `showMoreMenu` checked `Boolean(reportSlug)`, which is not set for unpublished drafts.

## Solution
- In `GameTheater.tsx`, enabled `useAgentBridge(true)` so the bridge is requested whenever reviewer credentials are present.
- Forwarded `agentBridge` to `GameFrame`, and held mount with `GameLoadScreen` while `agentBridge === undefined` to prevent iframe reloading.
- Added `agentAvailable` to `showMoreMenu` so the More menu and Agent Mode button are accessible on drafts.
- Made `progress` optional with a fallback in `GameLoadScreen.tsx`.
- Initialized `useAgentBridge` state synchronously to `null` for non-reviewers.
- Extracted `useOrientationMismatch` to stay below the `GameTheater.tsx` file size baseline.
- Added comprehensive unit tests in `GameTheater.agent.test.tsx` covering bridge delivery, menu presence, and `?agent=1` auto-open on draft games.

## Verification
- `npm run type-check` (passed)
- `npm run lint` (passed)
- `npm run test` (passed: 303 suites, 2408 tests)
- `npm run build` (passed)


gtanczyk:
Addressed in 0144a9e:
- `useAgentBridge` now accounts for `loading` from `useAuth()`. While auth is in flight (`loading === true`), `useAgentBridge` returns `undefined` (pending), holding mount with `GameLoadScreen`.
- Once auth settles, it either sets `null` (non-reviewers) or fetches the bridge and mounts with `agentBridge` (reviewers), ensuring the iframe is mounted exactly once without remounts or flicker.
- Added a dedicated unit test in `GameTheater.agent.test.tsx` verifying this exact lifecycle.

gtanczyk:
Reviewed at `8073f5c`. The core of this is right and the security invariant holds — one design concern worth raising before merge, with the evidence for it.

## The gate is intact

`useAgentBridge(true)` does not widen access. `/api/agent-play/bridge` checks `isReviewerSession` and 404s otherwise, this PR touches nothing under `apps/api`, and the panel only opens on `agentOpen = agentAvailable && agentWanted` — i.e. only once the script actually arrived. A non-reviewer setting `?agent=1` or reaching the menu gets nothing. `trackPlay={trackPlay && !agentAvailable}` still keeps agent sessions out of the play funnel, and the raw-HTML branch does not track plays at all. Holding the mount is also the right instinct: without it `srcDoc` changes once the bridge lands and the iframe reloads the game mid-session.

## The concern: "no auth answer ⇒ no game, ever"

`agentBridge === undefined → <GameLoadScreen />` makes the raw-HTML path — drafts and previews, which need no auth at all — depend on auth resolving, for everyone rather than just reviewers.

The `AuthContext` default flip in `8073f5c` is load-bearing, and what it is holding up is exactly this. Restoring `loading: true` and running `--dir apps/web` fails **26 tests** that pass otherwise. I checked why rather than assuming: `SubmissionStatusView.test.ts` and `CodeSurface.*.test.tsx` render the draft theater with **no `AuthProvider` and no `AuthContext` mock**, so they get the context default. With `loading: true` the hook stays `undefined` forever, the load screen never lifts, and the frame never mounts.

In production this is safe today — `refreshUser` has `finally { setLoading(false) }`, so auth always settles. But the hold has no timeout and no fallback; that `finally` is the only thing standing between it and a permanent load screen. Those 26 tests are evidence the coupling is broad, not hypothetical.

My worry is where the fix was applied. Flipping the shared default changes what `loading` means for *every* consumer rendered outside a provider, turning "not known yet" into "known: nobody" — and `BootGate` reads `loading` before deciding on the closed-beta splash. Dead in production today, since everything sits under `AuthProvider`, but it is a primed footgun in a file unrelated to this PR's subject.

Two options, either of which removes the need to touch `AuthContext` at all:

- **Bound the hold.** After a short timeout, mount with `agentBridge={null}`. A reviewer whose auth was pathologically slow pays one reload; nobody else waits.
- **Hold only when a reviewer is plausible** (a remembered client-side hint — a hint, not a grant, same standing as `hinted`) instead of for every visitor.

## Smaller note

The PR body describes the hold as preventing iframe churn, which reads as a reviewer-side concern. It is worth saying plainly that an ordinary creator opening their own draft now waits for `/api/auth/me` + `/api/health` before the game mounts — that is a new cost on a path that previously mounted immediately.

## Corrections to my own checking

Two things I got wrong and fixed before writing this, in case the reasoning gets re-traced: the new tests appeared to fail for me until I ran `build:packages` (`embedGameHtml` resolves from `@gamedevpl/contract`), and my first attribution of the failure count was wrong — 27 tests fail in my container on the unmodified branch too. The 26 above come from diffing that baseline against the reverted-default run, not from a raw count. CI on this head is fully green.

---

gtanczyk:
Addressed in 824cb06:

• **Reverted `AuthContext.tsx` entirely:** The context default is back to `loading: true` — no shared defaults or unrelated files are modified.
• **Hold only when a reviewer is plausible:** `useAgentBridge` now evaluates `plausible = hinted || agentModeRequested() || readReviewerHint()`. For regular creators, players, and isolated unit tests, it resolves to `null` synchronously on frame 1 without delay, mounting draft games immediately without waiting on `/api/auth/me` or `/api/health`.
• **Bounded hold:** If a reviewer is plausible but auth is slow or hangs, a fallback timeout (`AUTH_HOLD_TIMEOUT_MS = 1500`) expires to `null` so the game frame is never trapped in a permanent load screen.
• **Clean hint lifecycle:** When auth settles, `writeReviewerHint(hinted)` records or clears the client hint in `localStorage`.
• **Tests:** Updated `GameTheater.agent.test.tsx` to cover immediate mount for unhinted visitors, reviewer hold during auth, and the timeout fallback. All 303 suites (2,413 tests) and green gate passed.

PR #1419, opened 2026-09-19
fix(api): check membership on the routes a submission token reaches

gtanczyk:
Found while reviewing #1416. That PR kept producing access findings whose stated harm was "and mints its submission token", so I finally read what the token is worth. It is worth less than we assumed in one direction and more in another.

## What a submission token is

`mintToken` (`apps/api/src/platform/submission-token.ts:14`):

```
base64url(`${jobId}.${HMAC_SHA256(String(jobId), secret)}`)
```

Deterministic, no expiry, no subject binding, no revocation. One secret for the whole project (`deploy.yml` pins `PROJECT_ID: gamedevpl`). The same jobId yields the same token forever.

It also travels widely: it is the path segment of `/studio/:token`, so it is in the address bar, browser history, and every Cloud Run access log; and `notify.ts:609` puts `/status/${statusToken}` into notification records, **email bodies** and Web Push payloads. A forwarded email is a forwarded capability.

## Severity, checked rather than assumed

I surveyed every `:token` route. **No mutating route is reachable with the token alone** — `improve`, `feedback`, `turn`, `share`, `abandon`, `delete-game`, `handoff`, `seal`, `local-activity` and `connect` all require a session *and* an authorization predicate. That is the reassuring half, and it means the shelf-staleness findings on #1416 were genuinely about stale display rather than privilege escalation.

Two read paths were weaker.

### 1. `GET /api/submissions/:token/preview` served any token holder

It required a session but did no per-job authorization. `replyWithDraft` never calls `canActOnSubmissionOrSlug`; the helper is only used by the sibling `/api/drafts/:slug`. The comment above the route read:

> Only reachable by the token holder for that specific submission.

True, and precisely the assumption that breaks once membership changes. A removed editor — or anyone a status link reached — could fetch **unreviewed draft HTML**, including a red or unshared draft, for a game they no longer belong to.

Now checks `'read'` via `canActOnSubmissionOrSlug`, matching `creator-media.ts`, which already guards `/shot/:id` and `/preview/:id` this way. Test: the owner gets 200, a different logged-in account holding the same token gets 404. Fails without the fix.

### 2. `POST /api/submissions/:token/feedback` failed open on a missing record

```ts
if (store && record && !(await canActOnSubmissionOrSlug(store, record, request.user!.uid, 'build'))) {
```

`record` null skipped the ownership check entirely and fell through to the write path. A token outlives the round it names, so this is reachable. Without the fix the added test gets **200** and `createIssueComment` is called; now it is 404.

One existing test minted a token for a jobId it never created, which is not a state production can produce — it now seeds the record, which is also a more faithful fixture.

## Deliberately not changed

`GET /api/submissions/:token` requires no session at all. That is by design: status links are shared and emailed, and requiring auth would break every deep link already in someone's mailbox. It is a product decision, not a defect, so it is not in this PR.

Two things about it are worth a decision rather than silence:

- its refresh path calls `notifyOnTransition` (`submissions.ts:1557`), so an **unauthenticated** token holder can cause a notification and an email to be emitted on the owner's account, rate-limited only;
- combined with never expiring, a `/status/<token>` link from an old email is valid forever.

If we want that closed, the options are binding the token to a uid, adding an expiry, or keeping the unauthenticated route but making it a narrower projection. All three invalidate or change existing links, so it needs a call from you rather than from me.

## Gates

`npm run lint` (eslint, `comment-prose`, `module-size`, `env-manifest`, `firestore-read-cost`) clean; `apps/api` `tsc --noEmit` clean; 1533 tests across `submissions`, `delivery` and `creation` passing. Both fixes verified by reverting them and watching the specific new test fail.


gtanczyk:
@codex

gtanczyk:
@codex review

PR #1427, opened 2026-09-19
fix(api): count a document get once in the read meter

gtanczyk:
Found while verifying #1416 in production. The shelf document path — one `count()` plus one document get, **2 billed reads** — logged `fsReads=3` with `fsPaths={count:submissions:1, shelves/*:2}`. The code has exactly one `getShelf` on that path.

## Cause

`@google-cloud/firestore` 8.7.0, `build/src/reference/document-reference.js:183`:

```js
get() { return this._firestore.getAll(this).then(([result]) => result); }
```

`read-meter.ts` patches `DocumentReference.prototype.get` **and** `Firestore.prototype.getAll`. One `doc.get()` therefore records twice in production. `fake-firestore.ts` implements `get` directly, so tests and the `firestore-read-cost` gate count it once — which is why the gate says 2 and prod said 3.

## Consequence

Every historical `fsReads` figure that involved document gets is inflated by one per get. That includes the 553-reads-per-request baseline behind #1416, which was dominated by per-record `getGameAccess` gets. Ratios between metered numbers still hold roughly; absolute "billed" claims quoted from prod logs did not. The prod numbers in `docs/firestore-read-cost.md` should be read with that in mind; the gate baselines need no change.

## Fix

A metered `get` runs the underlying call inside an `AsyncLocalStorage` mark; the metered `getAll` skips recording while the mark is set. Async-context, not a module flag, because requests overlap. A direct `getAll(a, b, c)` still records once per reference.

## Test

Builds a real `DocumentReference` over a stubbed `Firestore.prototype.getAll` — no RPC, but the `get → getAll` delegation is the library's own code — and asserts one read for `get()` and three for a three-reference `getAll`. The first assertion fails without the guard (checked by removing it and running the file).

## Gates

`comment-prose`, `lint`, `apps/api` `tsc`, `apps/api/src/store` suite — each gated on its exit code with an explicit `|| exit 1`.


gtanczyk:
@codex

gtanczyk:
Re Copilot review 5258023517 (no inline thread): added in 38931224b. A document get and a two-reference direct `getAll` run concurrently in one tally and total three reads. I also checked it guards what it claims: replacing the `AsyncLocalStorage` mark with a module-level flag makes it fail (`expected 1 to be 3`) — a flag set by one request's `get` hides the other request's `getAll`, which is exactly why the mark is per async context.

@codex review 38931224b

PR #1430, opened 2026-09-20
perf(api): resolve a round's game access once per status poll

gtanczyk:
## What

The status poll is the hottest route on the site — three seconds per open Studio tab — and it resolved the same `GameAccess` document **twice per request**:

- `canActOnSubmissionOrSlug(...)`, to decide whether the viewer may see events, media and prior rounds;
- `currentOwnerUidSoft(...)`, to find whose quota the managed-builder headroom belongs to.

Both go through `resolveGameAccess`. One resolution now answers both, using the pure helpers that already existed: `canActOnGame` and `canonicalCreatorOwnerUid`.

## Cost

The existing baseline measures a **cold** request, but a 3s poll is warm, and warm is the number that multiplies by open tabs. Both are now gated:

| route | before | after |
| --- | --- | --- |
| `GET /api/submissions/:token` | 20 | **19** |
| `GET /api/submissions/:token (steady state)` | 3 | **2** |
| `GET /api/submissions/:token (share link, steady state)` | — | **1** |

A third off the steady-state poll. The remaining two are one `getSubmission` and one `getGameAccess`, both unavoidable per poll.

## Semantics

Unchanged on every path, deliberately:

- a resolution that **throws** leaves `access` null, which denies the viewer and falls back to the author's uid — exactly what `.catch(() => false)` and `currentOwnerUidSoft` did separately;
- an **unslugged** round is still its author's draft and reads no access at all;
- the resolution is gated on a viewer being present. It has to be: a **share link** polls the token with no session, and the old code never reached an access read there because the viewer test short-circuited on `viewerUid` first. The new `(share link, steady state)` row is the regression gate for exactly this — without the gating it costs 2.

## Tests

Full API suite: 5,628 passed.

`apps/api/src/improvement-admission.test.ts` times out at the default 5000 ms here, identically on `origin/master` with this change reverted — pre-existing, not touched by this PR.


gtanczyk:
@codex review

PR #1431, opened 2026-09-20
feat(api): play a shared draft by token, as the slug route already allows

gtanczyk:
Closes the one product question left open by #1419 and #1420.

## Why

`POST /api/submissions/:token/share` exists to let a creator show an unpublished game to
people who are not members, and `GET /api/drafts/:slug` serves exactly those people. The
token-addressed preview did not: after #1419 it required membership, full stop.

So a creator could share a game and send the link they actually had open — the Studio one,
`/studio/:token` — and the visitor got a 404. The draft was shared; only the address was
wrong. That is the behaviour change #1419 introduced, flagged at the time as a product
call rather than part of the fix.

## What

Both routes now answer one rule, `grantForRecord(record, uid)`; `canPlayDraft` becomes its
slug-shaped wrapper. How a draft was addressed no longer changes whether it plays.

Everything the rule already refused, it still refuses — this widens membership to *shared*,
not to *anyone*:

| Case | Answer |
| --- | --- |
| member (owner or editor) | plays, including a red build |
| shared + green gate | plays, pinned to the shared version |
| never shared | 404 |
| shared, but moderation-pulled | 404 for the visitor; the creator still reaches it |
| abandoned round | 404 |

A visitor is pinned: `?version=` cannot steer them off the shared build onto another one.
A member still passes it, which the Studio build rail needs.

## Tests

`draft-share-parity.test.ts`, four cases: both routes serve the visitor the same bytes;
the visitor is pinned while the creator is not; an unshared draft refuses; a pulled game
refuses the visitor but not its creator.

Negative control: with the route change reverted, the two widening tests fail and the two
refusal tests still pass — they are guardrails against over-widening, not new behaviour.

API suite green (5625). The eight red tests locally are `bake-media` and `image-webp` on
absent `@jsquash/webp` in this worktree — identical on master.

## Reviewer note

This is an access *widening* on a route that two PRs just narrowed. The narrowing was
correct and stays: a non-member with a token still gets nothing unless the creator pressed
Share and the gate is green. If you would rather keep token previews member-only and have
creators send `/play/<slug>`, this PR is the thing to close — the alternative is recorded
in the internal review doc as an open question either way.


gtanczyk (commented):
### Code Review: LGTM :white_check_mark:

The unification of draft playback authorization under `grantForRecord` cleanly resolves the parity gap between `/api/drafts/:slug` and `/api/submissions/:token/preview` without reopening the access holes closed in #1419.

#### Key Invariants & Access Control
- **Parity without over-widening:** Token previews are now accessible to non-members if and only if `draftSharedAt` is set and `shareGate.isGreen` passes. Unshared or pending builds remain strictly hidden (404).
- **Version pinning for visitors:** Returning `version: sharedDraftVersion(record)` in `DraftGrant` for non-members ensures `grant.version ?? requestedVersion` stays locked to the shared version, preventing visitors from traversing unshared or newer builds via `?version=`.
- **Member privileges preserved:** Owners and editors retain the ability to inspect unshared or in-progress builds and pass `?version=` for historical builds needed by the Studio rail.
- **Fail-closed boundaries:** Abandoned rounds, moderation-pulled games (`moderationBlockedAt`), and unslugged records fail closed to `null` (404) for non-members.
- **Merge-base:** Replays cleanly on top of latest `master` (`30515e2c5`) with no merge conflicts.

#### Verification Run
- **Type-check:** `npm run type-check` passed across all workspaces (contract, zone-core, api, cli, e2e, web, world).
- **Lint:** `npm run lint` passed with 0 errors.
- **Tests:** `npm test` passed across all workspaces (including all 312 web suites and the new parity suite in `draft-share-parity.test.ts`).
- **Build:** `npm run build` passed cleanly.

PR #1436, opened 2026-09-20
perf(web): stop the operator console polling a hidden tab

gtanczyk:
## What

`/api/admin/summary` is the **second most expensive route on the site**, and the operator console polled it every 30 seconds with no visibility check — the same flaw #1429 just removed from the home shelf.

One admin tab left open is 2,880 requests a day at ~93 reads each: **~262,000 reads a day** for a page nobody is looking at. In a production sample of the last day it was 32% of all metered reads, second only to `notify-sweep`.

The comment above that interval claimed the poll "costs one store read". It does not, and has not for a long time. Measured in production:

```
/api/admin/summary reads=93 calls=45
   33 submissions
   29 submissions/*/messages
   20 globalUsageShards/*
    7 globalUsage/*
    1 count:waitlist
```

It scans the active submissions and reads each one's messages, so the cost grows with the queue.

## How

All three timers in the console (`AdminConsole` 30s, `AdminJobsPanel` 30s, `AdminJobsQueue` 60s) now skip a hidden tab and catch up on `visibilitychange` — the rule `useActiveBuildCount` already follows. A visible console behaves exactly as before.

## Tests

New test in `AdminConsole.test.tsx`: a console mounted hidden asks for nothing, stays silent across five minutes of timers, and fetches exactly once on the way back. It fails if the gate is removed. Admin suite: 99 passed.

Two `module-size` ceilings raised for genuine additions — `AdminConsole.tsx` 278 → 285 (the gating) and its test 256 → 282 (the new case). No prose baselines touched.

## Not in this PR

The per-request cost itself. 93 reads for one dashboard is the next thing worth cutting, and `notify-sweep` at ~79 reads every two minutes is now the single largest source of reads on the site.


gtanczyk:
@codex review

gtanczyk:
Sharpening the numbers in the description, measured on the live revision `gamedev-app-02317-vaq` over three hours:

| route | requests | reads | avg |
| --- | --- | --- | --- |
| `/api/admin/summary` | 29 | 2,653 | **91.5** |
| `/api/internal/notify-sweep` | 30 | 2,538 | 84.6 |
| `/api/review/status` | 60 | 739 | 12.3 |
| `/api/catalog` | 17 | 470 | 27.6 |
| `/api/submissions/:token` | 24 | 329 | 13.7 |

Those top two are 74% of all metered reads in the window.

One correction to the description, for the record: 29 requests in three hours is **not** a 30-second poll, so the console is currently being opened intermittently rather than left open. The ~262,000 reads a day is the exposure when a tab *is* left open, not today's measured rate — which is closer to 21,000 a day, still 38% of the total. The gate removes the exposure; it does not claim to remove 262k reads from today's bill.

gtanczyk:
### Code Review Summary: LGTM ✅

The change is clean, well-scoped, and directly fixes a major Firestore read sink caused by background polling in the operator console.

#### Green Gate Verification (tested on `e40f87405` in isolated worktree):
- **`npm run type-check`**: Passed with 0 errors across all workspaces.
- **`npm run lint`**: Passed with 0 errors (module boundaries, env manifest, and size baselines verified).
- **`npm run test`**: Passed (313/313 web test files, 99 admin tests, all world and eslint-rules tests green).
- **`npm run build`**: Built cleanly.

---

#### Minor Notes & Observations:

1. **Mock cleanup in `AdminConsole.test.tsx`:**
   In `AdminConsole.test.tsx`:
   ```ts
   Object.defineProperty(document, 'hidden', { configurable: true, value: false });
   ```
   If an assertion throws earlier in the test, this line won't execute, and it leaves an own property on `document`. In `activeBuilds.test.ts`, `vi.spyOn(document, 'hidden', 'get')` paired with `vi.restoreAllMocks()` in `afterEach` is used. Not a blocker by any means, just a minor test-isolation pattern to keep in mind for future visibility tests.

2. **Next candidate for background poll gating (`NavHeader.tsx`):**
   In the production read breakdown from the PR description:
   - `/api/admin/summary`: 29 requests, 2,653 reads (addressed here)
   - `/api/internal/notify-sweep`: 30 requests, 2,538 reads
   - `/api/review/status`: **60 requests, 739 reads (avg 12.3 / call)**

   Checking `NavHeader.tsx` (line 112):
   ```ts
   const timer = setInterval(read, 120_000);
   ```
   `fetchReviewStatus()` runs an ungated interval for reviewer sessions even when the tab is hidden in the background. Applying the same `!document.hidden` + `visibilitychange` pattern there in a follow-up PR would be an easy win to cut down on those reads as well.

Ready to merge.

PR #1437, opened 2026-09-20
docs: close the read-cost cadence work and date the alert re-derivation

gtanczyk:
Documentation only.

`docs/firestore-read-cost.md` still read as though open browser tabs were the bill. They stopped being it this morning, with #1430, #1429 and #1436.

## What this records

**Where the cadence work ended.** Every polled surface is now gated — the status poll stops while hidden and widens as it idles, the header badge and the operator console skip a hidden tab, and the shelf has no timer at all. `/api/submissions/mine` and `:token` both bill 2 reads in steady state.

**The cold-versus-warm trap**, which cost real time here. The gate's baseline injects a route once against empty caches, and that cold number was quoted for a long time as if it were the per-poll cost. `:token` measures 19 cold and 2 warm; a 3s poll is warm. The fixture now carries `(steady state)` rows for exactly this reason.

**What is left, measured** on the serving revision over three hours:

| route | requests | reads | avg |
| --- | --- | --- | --- |
| `/api/admin/summary` | 29 | 2,653 | 91.5 |
| `/api/internal/notify-sweep` | 30 | 2,538 | 84.6 |

74% of every metered read, and both grow with the queue rather than with open tabs — a different problem from everything else in the file.

## The part worth reading

The alert re-derivation note moves from the week ending 2026-09-19 to **2026-09-27**, because the floor moved a second time on 2026-09-20. It now also states the risk the old paragraph left implicit:

> A31 fires above 600K/day and A30's drift condition above 8/s; the pace measured on 2026-09-20 is near 128K/day and about 1.5/s. A regression would have to be roughly **five times** the current floor before either speaks, which is larger than the 2026-09 incident this file exists because of.

No thresholds are moved here. Doing that from one weekend's numbers is what the file already forbids.


gtanczyk:
Superseded: the doc has moved to the private ops repo, and the closing section this PR added went across with it (ops `docs/firestore-read-cost.md`, commit `5e489bd` on `main`). The removal and the reference updates are in the PR that replaces this one.

PR #1438, opened 2026-09-20
docs: move firestore-read-cost.md to the private ops repo

gtanczyk:
Moves `docs/firestore-read-cost.md` to the private ops repo and deletes it here. Supersedes #1437, whose content went across with it.

## Why

The doc carries production read volumes, per-route costs, alert thresholds and the 2026-09 incident history. That is operational detail rather than contributor guidance, so it belongs next to the alert catalog and the runbooks that moved on 2026-09-15.

It is now at `docs/firestore-read-cost.md` in the ops repo, committed to `main` — including the closing section that was open here as #1437: the cadence axis is shut, both polled routes bill two reads in steady state, and what is left is cost per call.

## What stays

Everything an outside contributor or a review bot can act on:

- the gate — `npm run firestore-read-cost`, `eslint-rules/firestore-read-cost-check.mjs`, the baseline and the fixture;
- `docs/comment-prose-debt.md` and `docs/module-size-debt.md`. Those are **rules**, not measurements, and Copilot cited the prose one in a review earlier today — moving them would take that away from a reviewer that cannot read the private repo.

## References

Sixteen files pointed at the old path. Each now names the doc the way the moved runbooks are already named (`ops repo docs/firestore-read-cost.md`) rather than linking a path that no longer resolves. `docs/README.md` keeps a row with the move date and a link, matching the runbooks row directly below it.

Seven comments crossed the prose seal's twelve-word line once `ops repo` was added, so they lost a couple of words each. **No baseline moves in this PR** — `npm run lint` is green as is.

Type-check passes for `apps/api` and `apps/web`; `package.json` parses and both touched shell scripts pass `bash -n`.


gtanczyk:
@codex review

PR #1443, opened 2026-09-21
test(api): seal becoming shared apart from the marginal editor

gtanczyk:
Closes the one finding #1441 merged with still open: [`4058137975`](https://github.com/gamedevpl/www.gamedev.pl/pull/1441#discussion_r4058137975).

## What was wrong

The editor axis ran from **0 co-editors to 12**, so its slope averaged two costs that behave differently.

The first editor is not like the next one. It makes `editorUids.length === 0` stop holding, so `ownerQueryCoversAccess` turns false and the single-read count is replaced by a `listSubmissionsBySlug` scan. That is a fixed cost of a game *becoming shared*, paid once — no later member repeats it.

Averaged over 0→12, it read as `1.58 reads per editor`, which is a number no shape actually pays.

## Split

| step | reads | writes |
| --- | ---: | ---: |
| becoming shared (0 → 1 editor) | **+8** | +1…+2 |
| each member after that | **+1** | +1…+2 |

`claimSeal` is free of reads on both and pays one write either way.

The gate now reports them apart:

```
setSubmissionTitle    1r+0w per round, 2r+0w per game, 8r+1w per sharing, 1r+1w per editor
claimSeal             flat/round, flat/game, 0r+1w per sharing, 0r+1w per editor
```

## Why it matters for the ratchet

A refactor could trade the fixed cost for per-member work, stay under both the 12-editor total *and* the averaged slope, and still be more expensive for a game with twenty editors. Two nonzero counts close that hole; the new `sharing` axis seals the one-time step on its own.

## Tests

11 pass. `seals becoming shared apart from the marginal editor` asserts the fixed step exceeds the marginal one. Verified non-vacuous: merging the axes back to 0→12 fails three cases, including the slope seal.

`npm run lint` green, no baselines raised — the write-cost baseline is re-recorded because the shape set changed, which is what that file is for.


gtanczyk:
@codex review

PR #1446, opened 2026-09-22
Improve error handling for unknown CLI tools in intake agent

gtanczyk:
## Summary
Improved error handling in the intake agent to provide better feedback when the model calls non-existent CLI tools, and added fallback behavior to use reply text when available instead of failing closed.

## Key Changes
- **Better error messages**: When an unknown CLI tool is called, the error now includes the tool name (e.g., `unknown CLI tool: delete_everything`) instead of a generic message
- **Graceful fallback**: If a bad tool call is accompanied by reply text, the agent now returns the reply instead of throwing an error, allowing the model to provide helpful context to the user
- **Improved error detection**: Changed the validation logic to only throw when there's an unknown tool AND no fallback reply text is available

## Implementation Details
- The error check now specifically names the problematic tool in the error message
- Added a condition to check if reply text exists via `resultText(result).trim()` before throwing an error for unknown tools
- This allows the model to gracefully handle cases where it cannot execute a CLI action but can still provide a helpful response to the user

https://claude.ai/code/session_01USfxskcqcGcSoSUfW8WjcD


gtanczyk:
@codex

PR #1454, opened 2026-09-22
fix(api): stop treating a week-old dispatch as a session booting

gtanczyk:
## What happened

On 2026-09-22 `GET /api/submissions/:token` was **67% of all Firestore reads**, at a pace near 440K/day. Attributed from the request logs:

- 484 of 500 sampled polls came from a `node` client, 488 of 500 on **one token**, **489 of 500 on one instance**, every ~4s;
- the token is job **1000167** (`tabletop-turbos`, builder `self`), in `dispatched` since **2026-09-15**, with `agentEndedAt` on the 16th and no signal since;
- each poll billed ~20 reads — `submissions/*/messages` alone was 9.76 a poll.

## Why

`dispatched` bought two things meant for session boot, with **no limit on age**:

```ts
const ttlMs = status.phase === 'dispatched' ? 2_000 : 60_000;   // status cache
if (input.dispatched) return DISPATCHED_MS;                       // poll floor, 2s
```

A 4s poll lands past a 2s cache every time, so each one recomputed the whole status — `nativeJobStatus`, `listCreatorMessages` and all. A round stuck in `dispatched` for a week was being treated as a session starting up for a week.

This also retires a hypothesis from #1448, which blamed four instances with four caches. The polls were on one instance; the cache was being told to expire.

## Fix

`stillBooting(stateSince, now)` decides both sites: a dispatch younger than **10 minutes** keeps the 2s cache and floor, an older one gets the ordinary 60s cache and the age-based floor like any quiet round.

## Gate

The read-cost fixture gains the production shape — a round dispatched long ago, with a real round's messages, polled twice with the clock moved **4s** between them (the observed cadence):

| | reads |
| --- | ---: |
| `(stale dispatch, steady state)` without the fix | 6 |
| with the fix | **2** |

The fixture round carries three messages where the real one carried many, which is why production paid ~20 rather than 6; the mechanism is the same. The fixture's creator owns one more round as a result, so `GET /api/submissions/mine` moves **20 → 23** — verified to come only from that seed round.

Took the clock to make the gate honest: polled milliseconds apart, the fixture measured 2 with *and* without the fix.

## Not fixed here

- The client. It does not read `pollAfterMs`, so it will keep asking every 4s — just cheaply.
- The stuck round itself. The dispatch reaper did not collect a `self` round a week past its agent's end.

## Tests

`npm run lint` green. `status-poll-floor.test.ts` and `submissions.test.ts`: 252 pass, including three new `stillBooting` cases.


gtanczyk:
@codex review

Issue #1444, opened 2026-09-21
Self-build platform gaps from the Transport Tycoon Remake build (publish, OAuth, typecheck migration, byte budget, MCP errors)


gtanczyk:
Backlog captured from a long self-build round on `transport-tycoon-remake` (~Aug–Sep 2026, ~60 `game/*.ts` files, last good preview `v20260920T205926819Z-0441b78b95df`, round 89). Filed so it is not lost while the Agent-mode work lands.

**Agent-mode contract (`ui` / `observation` / helpers / policy) is deliberately NOT in this issue** — it is in flight as gamedevpl/www.gamedev.pl#1428 and gamedevpl/www.gamedev.pl-games#1620, with the panel layout already merged in #1426. One gotcha worth recording here because it will bite that build specifically: the bridge reads helpers from `harness.api`, `harness.helpers` and functions on the harness — **not** from `globalThis.__TTD__`. Since Check 17 forbids game sources writing `__GAME_HARNESS__`, `defineGame().agentApi()` is the supported route.

## 1. Publish is blocked for the builder

`submit_sources({ mode: "publish" })` fails with *only the owner can publish this game*; `mode: "preview"` works. The round is opened by the creator, the building agent holds a bot key / OAuth. The agent can stage and preview but cannot seal, so a delivery is spent discovering it.

Options, roughly in order of cheapness:
- surface it early — `canPublish: false` with a reason on `start` / `get_brief`, not at `submit_sources`
- let the builder publish while the creator has the round open
- Studio affordance: hand publish to the agent / co-owner / transfer

## 2. OAuth connector does not equal creator permissions

With the MCP server connected, `continue_draft` / `start` through the connector failed; opening round 89 needed a creator key over HTTP `tools/call`. A direct probe with the wrong bearer returns `{"error":"pass Authorization Bearer (creator key or OAuth) + slug"}`.

Either the connector account should carry opener rights for `continue_draft` / `start` / `stage_*` / `submit_sources` (preview at least), or the docs should say plainly that a creator key is required, with a safe handoff in Studio rather than pasting it into a chat.

## 3. MCP error responses break the tool output schema

An error body of `{error: …}` does not satisfy the success schema (no `jobId` / `sessionKey`), so clients raise `-32602 Structured content does not match the tool's output schema`. The operator sees "schema broken" when the truth is "not the owner" or "quota". This makes every auth or permission problem look like a platform defect.

[...]

PR #1457, opened 2026-09-23
A batch stage_upload_url mints what it can and names what it refused

gtanczyk:
One of the two items #1447 left open on #1444.

## What was wrong

`stage_upload_url({ paths })` validated every path and refused the **whole call** on the first bad one. Forty-nine good URLs thrown away, and the agent re-minting all fifty to find out which path was wrong — `tsconfig.json`, in the Transport Tycoon build that reported it.

## What changes

- A batch mints every deliverable path and lists the others in `rejected: [{ path, reason }]`, declared in the tool's output schema. The reason is the same message a refusal carried before, so it still names the allowlist.
- A lone `path` still refuses outright — there is nothing else to hand back.
- A batch where **nothing** is mintable refuses too, with the full `rejected` list, so it cannot read as a success that minted zero URLs.

Documented in `.claude/skills/byoca-mcp/SKILL.md` next to the upload notes from #1447.

## Verification

- New assertions in the existing `stage_upload_url + raw PUT` test: `tsconfig.json` between two good paths comes back as two URLs plus one named refusal; an all-bad batch refuses with both reasons. Against the old abort the test fails.
- `npm run lint` exit 0, `tsc` clean, agent-surface suite 758 green.

The other open item — the kit's `SKILL.md` naming the local text harness — is in the games repo as a separate PR, since that half needs its own `check:pr` gate. Opener rights for OAuth connector accounts remain a product decision and are not in either.

https://claude.ai/code/session_01LptvkD1n5mxDptHvhJfA53

---
_Generated by [Claude Code](https://claude.ai/code/session_01LptvkD1n5mxDptHvhJfA53)_


gtanczyk:
@codex review

Small change: a batch `stage_upload_url` used to refuse the whole call over one bad path; it now mints the rest and lists refusals in `rejected`. Worth adversarial attention: whether a partially-refused batch should be `isError` (I chose no — the minted URLs are real and usable — and made an all-refused batch the only error), and whether `rejected` should be omitted or empty when nothing was refused (omitted).

---

PR #1464, opened 2026-09-23
Split runtime signing keys across services

gtanczyk:
## Changes

- Use separate configurable signing keys for zone tickets, player tags, party rooms, and email unsubscribe links, with API fallbacks for the rollout.
- Let the world service accept the previous zone ticket key during rotation and require `ZONE_TICKET_SECRET` at boot.
- Limit relay-only registration to its room, health, and CSP-report endpoints; disable session authentication and use an ephemeral session key in that mode.
- Add key-selection, rotation, world-boot, and relay-route tests.

## Validation

- `npm run type-check`
- `npm run lint`
- `VITEST_MAX_WORKERS=4 npm run test`
- `npm run build`

## Deployment gate

`apps/world` must not be deployed until the owner has created `zone-ticket-secret` and wired it (plan step I-2). Infrastructure changes remain owner-run and are outside this PR.


gtanczyk (commented):
Verified in an isolated worktree (`npm run type-check && npm run lint && VITEST_MAX_WORKERS=4 npm run test && npm run build` all pass cleanly).

A few observations and suggestions for follow-up or refinement:

1. **`relayOnly` detection in `buildApp` (`apps/api/src/platform/app.ts:227`)**:
   `const relayOnly = isRelayOnly();` only checks `process.env.MP_RELAY_ONLY`. If a caller or test sets `{ multiplayerRoutes: { relayOnly: true } }` programmatically without setting `MP_RELAY_ONLY` in the environment, `relayOnly` evaluates to `false` and the early-return relay setup doesn't activate. To follow the convention that options override env:
   ```ts
   const relayOnly = options.multiplayerRoutes?.relayOnly ?? isRelayOnly();
   ```

2. **Programmatic `prevSecret` in `WorldAppOptions` (`apps/world/src/app.ts:54`)**:
   `WorldAppOptions` currently extends `Omit<ZoneHostOptions, 'secret' | 'prevSecret'>` and defines `secret?: string`, but omits `prevSecret`. In `buildWorldApp`, `prevSecret: process.env.ZONE_TICKET_SECRET_PREV` is hardwired from env. Allowing `prevSecret?: string` in `WorldAppOptions` and using:
   ```ts
   prevSecret: options.prevSecret ?? process.env.ZONE_TICKET_SECRET_PREV,
   ```
   would allow testing and constructing `buildWorldApp` with rotation keys directly without needing `vi.stubEnv`.

3. **Check order in `registerAuthPlugin` (`apps/api/src/platform/auth.ts:338`)**:
   The check `if (!sessionSecret && isProd) throw ...` runs before `if (options.enabled === false) return;`. Moving `if (options.enabled === false) return;` to the top of `registerAuthPlugin` would allow omitting `sessionSecret` entirely when auth is disabled, avoiding the need for `sessionSecret: relayOnly ? randomBytes(32).toString('hex') : ...` in `buildApp`.

4. **Rollout prerequisite**:
   As noted in the deployment gate, `apps/world` must not be promoted until `zone-ticket-secret` is created and mapped in Cloud Run (with `ZONE_TICKET_SECRET_PREV` mapped to the old `session-secret` to validate in-flight tickets), because `world` intentionally no longer falls back to `SESSION_SECRET`.

gtanczyk:
@codex code review

